PR #1, opened 2018-04-06
Timeline for Removal of Deprecated Code

buskamuza:
Concern raised by the community that it's not clear how long deprecated code will be preserved based on the rules documented at [Removal of deprecated code](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#removal-of-deprecated-code).

This proposal is to update existing documentation/processes in order to address existing community concerns.

### DIFF
--- a/proposals/backward-compatibility/deprecated-code-removal-timeline.md
+++ b/proposals/backward-compatibility/deprecated-code-removal-timeline.md
@@ -1,0 +1,37 @@
+# Timeline for Removal of Deprecated Code
+
+## Overview
+
+Concern raised by the community that it's not clear how long deprecated code will be preserved based on the rules documented at [Removal of deprecated code](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#removal-of-deprecated-code).
+
+## Requirements and Acceptance Criteria
+
+1. Clear message about the timeline of deprecated code removal, which allows the community to plan their activities on refactoring the extensions
+1. Strict enough rules, so that the extension developers have enough time for upgrading their code
+1. Loose enough rules, so that Magento core team could avoid unreasonable maintenance costs caused by deprecated code
+
+## Theory of Operation
+
+### Option 1: Component-based rules (current)
+
+Preserve current rules. Deprecated code is preserved:
+* `@api` code: Until the next major version of the **component**
+* non-`@api` code: The next 2 minor releases or until a major release of the **component**
+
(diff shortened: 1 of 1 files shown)


MagentoSeba:
I think a good start is when documenting changes like this: 
What Changed                                                           -- | --                           How Changed
Magento\Elasticsearch\Elasticsearch5\SearchAdapter\Mapper | Class was added.
Magento\Elasticsearch\Elasticsearch5\SearchAdapter\Query\Builder | Class was added.
Magento\Elasticsearch\Elasticsearch5\Model\Adapter\FieldType | Class was added.

To actually include How it changed, which class was added, or which class was removed, etc. being more specific in the documentation.

dmanners:
The suggestion raised by the community member to me was to update the deprecated tag to include the version that the code will be removed in. Though again I think this could come down to an issue or lack of visibility between marketing (or meta package) and module versioning. It would be fine to add that this deprecation will be removed in version 100.0.x but knowing that this would translate to 2.2.x would also be important for people.

It might also be a nice idea to have a documentation "What was deprecated in 2.2.X" similar to http://php.net/manual/en/migration53.deprecated.php

heldchen:
not just saying that something was deprecated, but also giving examples what to use instead would be most useful... especially when lots of the deprecated stuff is actually still heavily used internally.

buskamuza:
@MagentoSeba , could you please elaborate on this?
```
Magento\Elasticsearch\Elasticsearch5\Model\Adapter\FieldType | Class was added.
```
says that `Magento\Elasticsearch\Elasticsearch5\Model\Adapter\FieldType` was added.
I assume you're saying about this page http://devdocs.magento.com/guides/v2.2/release-notes/changes/ce_changes.html ?

buskamuza:
@heldchen , this is in our guidelines and this is the practice we follow when possible

antonkril:
@dmanners I agree that such annotation would be useful for people. But it is not possible to guarantee the version of api removal: backlog prioritization is very dynamic and does not depend on single group opinion.

piotrekkaminski:
The general rule of thumb is we deprecate in version 2.X and remove in 2.(X+1), however what @heldchen said about showing what is the right replacement (or where to look for the right example, if core code does not show the best way in some places) is critical and a feedback we saw a lot.

buskamuza:
Closing the PR, as it looks like most of the real concerns are unrelated to the described issue and addressed either by the existing docs or by the existing process.

PR #10, opened 2018-07-25
Allowed HTTP Methods for Actions

AlexMaxHorkun:
Limit actions to processing only requests with certain HTTP methods in a declarative way

### DIFF
--- a/design-documents/allowed-http-methods-for-actions.md
+++ b/design-documents/allowed-http-methods-for-actions.md
@@ -29,27 +29,26 @@ Since we use by-convention routing it would be inconsistent to defined actions�
 
 etc.
 
-One interface per each HTTP method (using list of HTTP methods from \Zend\Http\Request::METHOD_* constants). We will have an exentable map *interface* => *HTTP method* passed to the validator to allow custom HTTP methods to be processed.
+One interface per each HTTP method (using list of HTTP methods from \Zend\Http\Request::METHOD_* constants). We will have an exentable map *HTTP method* => *Interface* passed to the validator to allow custom HTTP methods to be processed.
 
 Request method will be validated in the FrontController before executing a matched action, somewhat like this:
 
->foreach ($interfaceMap as $interface => $httpMethod) {
+>if (!(array_key_exists($request->getMethod(), $interfaceMap) && $actionInstance instanceof $interfaceMap&#91;$request->getMethod()&#93;)) {
 >
->&nbsp;&nbsp;&nbsp;&nbsp;if ($actionInstance instanceof $interface && $request->getMethod() !== $httpMethod) {
->
->&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;throw new NotFoundException(__(‘Page not found.’));
->
->&nbsp;&nbsp;&nbsp;&nbsp;}
+>&nbsp;&nbsp;&nbsp;&nbsp;throw new NotFoundException(__(‘Page not found.’));
 >
 >}
 
  
 To limit already existing Actions and get developers to know these new interfaces a temporary event listener will be added for controller_action_postdispatch event to log HTTP methods used with actions and then the full suite of functional tests (MTF and MFTF) would be ran to get information on as much actions as possible.
 
-After corresponding Http*Method*ActionInterface will be added to every Action class we have information on. If an action is logged to be accessed only by GET methods it will allow only GET, if only by POST – allowing only POST etc. If by more than one methods – the Action class will not be updated.
+After corresponding Http*Method*ActionInterface will be added to every Action class we have information on. If an action is logged to be accessed only by GET methods it will allow only GET, if only by POST – allowing only POST etc. If by more than one methods – the Action will be investigated and either updated to implement multiple interfaces or left alone if proven too difficult to determine HTTP methods.
+
+To sway developers to specify HTTP methods for actions a new rule for code mess detector will be added that will require implementations of ActionInterface to implement also at least one of Http*Method*ActionInterface interfaces.
 
 ### Work Breakdown
  * Add Http*Method*ActionInterface and implement request validation in the FrontController
+ * Add a rule to code mess detector
  * Create logging mechanism for Actions
  * Log HTTP methods used when running full functional tests’ suite
  * Create script to add Http*Method*ActionInterface to Action classes we’ve collection information about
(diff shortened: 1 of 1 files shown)


design-documents/allowed-http-methods-for-actions.md
```diff
@@ -0,0 +1,56 @@
...
+>&nbsp;&nbsp;&nbsp;&nbsp;if ($actionInstance instanceof $interface && $request->getMethod() !== $httpMethod) {
+>
+>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;throw new NotFoundException(__(‘Page not found.’));
+>
+>&nbsp;&nbsp;&nbsp;&nbsp;}
+>
+>}
+
+ 
+To limit already existing Actions and get developers to know these new interfaces a temporary event listener will be added for controller_action_postdispatch event to log HTTP methods used with actions and then the full suite of functional tests (MTF and MFTF) would be ran to get information on as much actions as possible.
```
melnikovi: Can we leave logging in place to let extension developers know that they need to update their extensions?

design-documents/allowed-http-methods-for-actions.md
```diff
@@ -0,0 +1,56 @@
...
+
+One interface per each HTTP method (using list of HTTP methods from \Zend\Http\Request::METHOD_* constants). We will have an exentable map *interface* => *HTTP method* passed to the validator to allow custom HTTP methods to be processed.
+
+Request method will be validated in the FrontController before executing a matched action, somewhat like this:
+
+>foreach ($interfaceMap as $interface => $httpMethod) {
+>
+>&nbsp;&nbsp;&nbsp;&nbsp;if ($actionInstance instanceof $interface && $request->getMethod() !== $httpMethod) {
+>
+>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;throw new NotFoundException(__(‘Page not found.’));
```
melnikovi: I remember that we agreed to return 405 response code or no?

design-documents/allowed-http-methods-for-actions.md
AlexMaxHorkun: I remember we discussed this, I just forgot to mention it - we can add a debug message each time an action without HTTP method declared s processed

design-documents/allowed-http-methods-for-actions.md
AlexMaxHorkun: No, we chose 404 because it's more consistent with what we do now and server must never return a 405 to GET and HEAD requests (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405)

design-documents/allowed-http-methods-for-actions.md
```diff
@@ -0,0 +1,56 @@
...
+>
+>}
+
+etc.
+
+One interface per each HTTP method (using list of HTTP methods from \Zend\Http\Request::METHOD_* constants). We will have an exentable map *interface* => *HTTP method* passed to the validator to allow custom HTTP methods to be processed.
+
+Request method will be validated in the FrontController before executing a matched action, somewhat like this:
+
+>foreach ($interfaceMap as $interface => $httpMethod) {
```
buskamuza: It shouldn't be necessary to iterate through all methods. You can do ``if (isset($interfaceMap[$request->getMethod()])``. If doesn't exist, the method is not supported by the application. If `$actionInstance` does not implement the interface, also - an error.
The map will be opposite:
```
POST => HttpPostInterface
GET => HttpGetInterface
etc.
```

design-documents/allowed-http-methods-for-actions.md
```diff
@@ -0,0 +1,56 @@
...
+>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;throw new NotFoundException(__(‘Page not found.’));
+>
+>&nbsp;&nbsp;&nbsp;&nbsp;}
+>
+>}
+
+ 
+To limit already existing Actions and get developers to know these new interfaces a temporary event listener will be added for controller_action_postdispatch event to log HTTP methods used with actions and then the full suite of functional tests (MTF and MFTF) would be ran to get information on as much actions as possible.
+
+After corresponding Http*Method*ActionInterface will be added to every Action class we have information on. If an action is logged to be accessed only by GET methods it will allow only GET, if only by POST – allowing only POST etc. If by more than one methods – the Action class will not be updated.
```
buskamuza: > If by more than one methods – the Action class will not be updated

Why not make it `implement` multiple interfaces? Or if it's not clear why it really supports multiple methods, create a task to review it (as it can be a security issue).

AlexMaxHorkun: because we cannot be sure that an action actually expects multiple request methods or a test is invalid without investigation and if too many cases like that would be found it can take too match time to investigate them. I think I can write something like this: "If by more than one methods – the Action class will be updated to implement multiple interfaces or ignored." Would that be ok?

design-documents/allowed-http-methods-for-actions.md
AlexMaxHorkun: ok, I'll update that part

AlexMaxHorkun:
I think adding a rule to dev/tests/static/framework/Magento/CodeMessDetector/* that will detect classes implementing ActionInterface but not any HTTP method aware interfaces should be sufficient enough to alert developers about using these new interfaces

AlexMaxHorkun:
I have updated the document to address the concerns

design-documents/allowed-http-methods-for-actions.md
melnikovi: I agree with Olga and would propose for now agree that we will need to review each case and discuss one more time if turns out that there are too many cases that need to be reviewed. Or you already know?

design-documents/allowed-http-methods-for-actions.md
melnikovi: I'm ok with not having this if we going to have a mess detector rule.

buskamuza:
> Everybody's ok with adding a debug message each time an action without HTTP method restriction is found?

It may be an overkill, if used in production mode.

schmengler:
In the example, a `PageNotFound` exception is thrown. The correct response in HTTP semantics would be a `405 Method Not Allowed` status code.

PR #11, opened 2018-07-26
MAGECLOUD-2503: Cron-module batch DELETEs to prevent SQL lock wait timeouts

JacobBrownAustin:


JacobBrownAustin:
I did some benchmarks on this today.  Before the changes, 7-28 seconds for the DELETE queries.  In the prototype, with batching the deletes, the longest query one was 9.5 seconds.
That was with 775064 records in cron_schedule.

akaplya:
reason for wait locks is another (2nd) transaction (which obviously locks rows request by 1st) not the number of rows in a table.

kandy:
We decided to introduce with "Read Commited" transaction mode  for cron operations

PR #12, opened 2018-07-26
MAGECLOUD-2503: cron-module add index to speed up DELETEs

JacobBrownAustin:


JacobBrownAustin:
Just did some newer benchmarks on this.  On my test cloud environment with 264,055 rows in cron_schedule, before this index, DELETEs that don't delete anything take 204-232 ms.  After the index, they only take 0-2 ms.  :grin:

kandy:
Lets to add key on this fields (`job_code`, `status`, `scheduled_at`)

PR #13, opened 2018-07-26
MAGECLOUD-2503: cron-module add new table for trySetJobUniqueStatusAtomic()

JacobBrownAustin:


JacobBrownAustin:
I did some benchmarking on this today.  Before changes, trySetJobUniqueStatusAtomic() method takes between 25-5000 ms.  After all changes in my prototype, it takes between 0-2 ms.

PR #14, opened 2018-07-26
MAGECLOUD-2503: Mysql Adapter Should Retry After Deadlock

JacobBrownAustin:


akaplya:
Database adapter is not an appropriate layer for rerunning failed operations. Because an adapter does not have information about other data changes processed by a transaction before failure.

JacobBrownAustin:
@akaplya There's two things about that:
A) It already does that for disconnects (server has gone away and lost connections).
Source: https://github.com/magento/magento2/compare/2.2-develop...JacobBrownAustin:MAGECLOUD-2503-2.2-develop?expand=1#diff-2b5067e3a6c0d21915ae0431519dcc76L556

B) In my proposal here, I check that there is nothing else in this transaction.  Therefore, there are no other data changes in this transaction before failure.
Source: https://github.com/magento/magento2/compare/2.2-develop...JacobBrownAustin:MAGECLOUD-2503-2.2-develop?expand=1#diff-2b5067e3a6c0d21915ae0431519dcc76R559

akaplya:
This is non-transparent which operations will be rerun. It always will depend on the context of execution. So for the same operation, we may face different behavior depends on the execution context. As a result, the complexity of troubleshooting significantly increase.

Definitely, we need re-try mechanism. But it has to be implemented at a higher level. Ideally above the contracts.

JacobBrownAustin:
With this proposal, the only thing I'm changes is that it will retry if it gets a Deadlock error.  Therefore, the only operation that will be rerun is the query that the SQL deadlock detector decided to kill.  The context should not matter at all.  Just like the context of why the network connection failing doesn't matter.  

If you are saying that we should add more logging when the automatic retry happens, I totally agree with that.

Also, if you are saying that the network connection retries should also look at getTransactionLevel() to make sure they aren't retrying in the middle of a transaction, I agree with that too.

I've seen something wrong with my prototype, and I'll try to clear it up tonight.

JacobBrownAustin:
Okay, I've updated the example so it's easier to read, and fixed two small things.
A) When retrying query that got Deadlock error, we don't need to reconnect.
B) When there were disconnect errors, we were retrying even when in the middle of a transaction.  That doesn't make sense, so I changed that.  (This looks like an existing bug in Magento.)

https://github.com/magento/magento2/compare/2.2-develop...JacobBrownAustin:MAGECLOUD-2503-2.2-develop?expand=1

akaplya:
I would say that B) case is a mistake we have to fix

JacobBrownAustin:
I'm adding the example SQL Deadlock I attached to MAGECLOUD-2503 to this issue so you can see it for reference.  This is an example of the DELETE query from one process deadlocking against the same DELETE query from a different process.  Both processes are `magento cron:run`, but one of them is a different cron group than the other.

------------------------
LATEST DETECTED DEADLOCK
------------------------
2018-07-18 20:42:54 7fdce6ca5700
*** (1) TRANSACTION:
TRANSACTION 8936031966, ACTIVE 0 sec starting index read
mysql tables in use 1, locked 1
LOCK WAIT 10 lock struct(s), heap size 2936, 79 row lock(s)
MySQL thread id 31644, OS thread handle 0x7fdce6d80700, query id 20912368 246.4.160.9 user updating

DELETE FROM `cron_schedule` WHERE (status = 'success') AND (job_code in ('indexer_reindex_all_invalid', 'indexer_update_all_views', 'indexer_clean_all_changelogs', 'magento_targetrule_index_reindex')) AND (created_at < '2018-07-11 20:42:38') LIMIT 1024

*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 2762 page no 8 n bits 208 index `PRIMARY` of table `main`.`cron_schedule` trx table locks 1 total table locks 2  trx id 8936031966 lock_mode X locks rec but not gap waiting lock hold time 0 wait time before grant 0
*** (2) TRANSACTION:
TRANSACTION 8936031821, ACTIVE 0 sec fetching rows
mysql tables in use 1, locked 1
14 lock struct(s), heap size 2936, 1432 row lock(s), undo log entries 70
MySQL thread id 31639, OS thread handle 0x7fdce6ca5700, query id 20912288 246.4.160.9 user updating

DELETE FROM `cron_schedule` WHERE (status = 'success') AND (job_code in ('currency_rates_update', 'backend_clean_cache', 'visitor_clean', 'catalog_index_refresh_price', 'catalog_product_flat_indexer_store_cleanup', 'catalog_product_outdated_price_values_cleanup', 'catalog_product_frontend_actions_flush', 'catalog_product_attribute_value_synchronize', 'catalogrule_apply_all', 'system_backup', 'sales_clean_quotes', 'sales_clean_orders', 'aggregate_sales_report_order_data', 'aggregate_sales_report_invoiced_data', 'aggregate_sales_report_refunded_data', 'aggregate_sales_report_bestsellers_data', 'sales_grid_order_async_insert', 'sales_grid_order_invoice_async_insert', 'sales_grid_order_shipment_async_insert', 'sales_grid_order_creditmemo_async_insert', 'sales_send_order_emails', 'sales_send_order_invoice_emails', 'sales_send_order_shipment_emails', 'sales_send_order_creditmemo_emails', 'captcha_delete_old_attempts

*** (2) HOLDS THE LOCK(S):
RECORD LOCKS space id 2762 page no 8 n bits 208 index `PRIMARY` of table `main`.`cron_schedule` trx table locks 1 total table locks 2  trx id 8936031821 lock_mode X lock hold time 0 wait time before grant 0
*** (2) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 2762 page no 29 n bits 208 index `PRIMARY` of table `main`.`cron_schedule` trx table locks 1 total table locks 2  trx id 8936031821 lock_mode X waiting lock hold time 0 wait time before grant 0
*** WE ROLL BACK TRANSACTION (1)

JacobBrownAustin:
Here's a different example, also from `magento cron:run` command that I captured earlier when I was first investigating MAGECLOUD-2445 (the source of MAGECLOUD-2503).

This example is different because the two queries are different.  This time, one of them is one of the DELETE queries, but the other query is the UPDATE query.

------------------------
LATEST DETECTED DEADLOCK
------------------------
2018-07-05 05:07:05 7f372a3ac700
*** (1) TRANSACTION:
TRANSACTION 280666259, ACTIVE 1 sec starting index read
mysql tables in use 2, locked 2
LOCK WAIT 4 lock struct(s), heap size 1184, 3 row lock(s)
MySQL thread id 13811, OS thread handle 0x7f3744109700, query id 44662775 246.4.144.114 user Sending data
UPDATE `cron_schedule` AS `current`
 LEFT JOIN `cron_schedule` AS `existing` ON existing.job_code = current.job_code AND (existing.executed_at > UTC_TIMESTAMP() - INTERVAL 1 DAY OR existing.executed_at IS NULL) AND existing.status = 'running'
SET `current`.`status` = 'running'
WHERE (current.schedule_id = '3846510') AND (current.status = 'pending') AND (existing.schedule_id IS NULL)
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 1587 page no 6624 n bits 264 index `PRIMARY` of table `main`.`cron_schedule` trx table locks 1 total table locks 2  trx id 280666259 lock mode S locks rec but not gap waiting lock hold time 1 wait time before grant 0
*** (2) TRANSACTION:
TRANSACTION 280666258, ACTIVE 1 sec fetching rows
mysql tables in use 1, locked 1
911 lock struct(s), heap size 112168, 114280 row lock(s)
MySQL thread id 13816, OS thread handle 0x7f372a3ac700, query id 44662773 246.4.144.114 user updating
DELETE FROM `cron_schedule` WHERE (status = 'pending') AND (job_code in ('currency_rates_update', 'backend_clean_cache', 'visitor_clean', 'catalog_index_refresh_price', 'catalog_product_flat_indexer_store_cleanup', 'catalog_product_outdated_price_values_cleanup', 'catalog_product_frontend_actions_flush', 'catalog_product_attribute_value_synchronize', 'sales_clean_quotes', 'sales_clean_orders', 'aggregate_sales_report_order_data', 'aggregate_sales_report_invoiced_data', 'aggregate_sales_report_refunded_data', 'aggregate_sales_report_bestsellers_data', 'sales_grid_order_async_insert', 'sales_grid_order_invoice_async_insert', 'sales_grid_order_shipment_async_insert', 'sales_grid_order_creditmemo_async_insert', 'sales_send_order_emails', 'sales_send_order_invoice_emails', 'sales_send_order_shipment_emails', 'sales_send_order_creditmemo_emails', 'system_backup', 'catalogrule_apply_all', 'aggregate_sales_report_co
*** (2) HOLDS THE LOCK(S):
RECORD LOCKS space id 1587 page no 6624 n bits 264 index `PRIMARY` of table `main`.`cron_schedule` trx table locks 1 total table locks 2  trx id 280666258 lock_mode X lock hold time 1 wait time before grant 0
*** (2) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 1587 page no 8005 n bits 176 index `PRIMARY` of table `main`.`cron_schedule` trx table locks 1 total table locks 2  trx id 280666258 lock_mode X waiting lock hold time 0 wait time before grant 0
*** WE ROLL BACK TRANSACTION (1)

JacobBrownAustin:
I've been able to get the Deadlocks to happen today, but has always happened outside of the process I have attached with XDEBUG, so I've been unable to get a stack trace for you.

JacobBrownAustin:
> InnoDB uses automatic row-level locking. You can get deadlocks even in the case of transactions that just insert or delete a single row. That is because these operations are not really “atomic”; they automatically set locks on the (possibly several) index records of the row inserted or deleted.

Source: https://dev.mysql.com/doc/refman/8.0/en/innodb-deadlocks-handling.html

JacobBrownAustin:
> Before and above all diagnosis, it is always an important practice to have the applications catch deadlock error (MySQL error no. 1213) and handle it by retrying the transaction.

Source: https://www.percona.com/blog/2014/10/28/how-to-deal-with-mysql-deadlocks/

Note that in this proposal, I'm only suggesting retrying the query if it is not part of a bigger transaction, because it is super easy that way.  For transactions that are multiple queries, it will have to be handled specific to each of those transactions on how or when to retry.

kandy:
Retry After Deadlock should be done on application level

PR #15, opened 2018-07-31
[ongoing discussion] Draft for cart operations and other GraphQL improvements

paliarush:
For the context see [recording](https://youtu.be/zGHqYnm8ByI?t=30m53s) of previous GraphQL community meeting.

GraphQL project wiki: https://github.com/magento/graphql-ce/wiki


design-documents/graph-ql/coverage/add-items-to-cart/AddConfigurableProductToCart.graphqls
```diff
@@ -0,0 +1,37 @@
...
+    cartId: String!
+    cartItems: [ConfigurableProductCartItemInput!]!
+}
+
+input ConfigurableProductCartItemInput {
+    sku: String!
+    quantity: Float!
+    configurable_options:[ConfigurableOptionInput!]!
+    customizable_options:[CustomizableOptionInput!]
+}
```
paales: Can't configurable_options be replaced by a single `simple_sku: Sku!` field?

paliarush: If we go with `simple_sku`, then the client on its end will have to map selected options combination to the SKU:
- Get a map during product request (option1, option2, optionN => simple SKU, ...)
- Calculate current simple SKU based on selected options
- Send simple SKU along with `add product to cart` request

Looks like more work for client app developers. I am probably missing something, please clarify why `simple_sku` is preferable.

Thanks for the review!

design-documents/graph-ql/coverage/add-items-to-cart.md
```diff
@@ -0,0 +1,49 @@
+**Overview**
+
+As a Magento developer, I need to manipulate the shopping cart via GraphQL so that I can build basic ecommerce experiences for shoppers on the front-end using only GraphQL.
```
DrewML: > for shoppers on the front-end using only GraphQL.

This is a big nitpick, so...sorry!

I think it's super important that we avoid language that couples the concept of GraphQL in Magento to the front-end specifically.

I know it's a small thing, but we want to make sure everyone has the mindset that this is an API for _any_ external consumer.

design-documents/graph-ql/coverage/add-items-to-cart.md
```diff
@@ -0,0 +1,49 @@
...
+    - Add items to cart
+    - Set shipment method
+    - Set payment method
+    - Set addresses
+    - Same granularity for updates and removals
+- Possibility to combine mutations for checkout steps
+- Can create "order in one call" mutation in the future if needed
+- Hashed IDs for cart items
+- Single input object
+- Async nature of the flow must be supported on the client side (via AJAX calls)
```
DrewML: Might want to expand on this point more, since all requests from the browser are async (barring sync `XMLHTTPRequest`, which is deprecated).

Maybe add some additional language around this clarifying that is creates a "job" of sorts, which helps imply that the results of the mutation will be some object representing a task that will be finished some time in the future.

design-documents/graph-ql/coverage/add-items-to-cart.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+**Open questions:**
+
+- Do we want to implement server-side asynchronous mutations by default?
```
DrewML: By default for cart mutations, or for _all_ mutations?

design-documents/graph-ql/coverage/add-items-to-cart.md
DrewML: I think it's important to include _why_ there is a desire for these mutations to be async. When we chatted briefly yesterday, I got the impression that it's because this could be a long-running operation, and we don't want to keep the request open. Is that accurate?

It's just worth making the justification clear, because a concept of a "job" that needs to be polled (or a subscription) pushes some complexity to the consumer of the API.

design-documents/graph-ql/coverage/add-items-to-cart/AddConfigurableProductToCart.graphqls
paales: I think there already is some form of simple_sku mapping in the frontend, especially if a frontend developer wishes to fetch images from the simple product etc. Or are there other strategies in that regard?

design-documents/graph-ql/coverage/add-items-to-cart.md
paliarush: Please suggest how to stress, that GraphQL in Magento is intended for store front scenarios only, not for admin ones.

design-documents/graph-ql/coverage/add-items-to-cart.md
paliarush: Asynchronous server side implementation is not approved yet. The schema may change, if it will be approved.

design-documents/graph-ql/coverage/add-items-to-cart.md
paliarush: @akaplya suggested for all mutations. Still under discussion.

design-documents/graph-ql/coverage/add-items-to-cart.md
paliarush: Not necessarily long-running, may be just a lot of concurrent requests during peak hours. Decision on the async server side is not made yet.

design-documents/graph-ql/coverage/add-items-to-cart/AddConfigurableProductToCart.graphqls
DrewML: @paales Can you help by providing the motivation for using `simple_sku` in lieu of sending along options?

AFAICT, it seems like, in any situation you'd be adding an item to a cart, you'd always have a list of the options (since the shopper needs to select them). So it feels like needing to track a `simple_sku` on the client-side might be work that is unnecessary. Granted, it's very little extra work, but I'm curious what the motivation would be for that > options list.

design-documents/graph-ql/coverage/add-items-to-cart/AddSimpleProductToCart.graphqls
```diff
@@ -0,0 +1,79 @@
...
+    id: Int!
+    qty: Float!
+    product: CartItemProduct!
+    image: CartItemImage!
+    prices: CartItemPrices!
+}
+
+type CartItemImage {
+    alt_text: String!
+    thumbnail: ProductMediaGalleryEntriesContent!
```
masodu: Probably out of scope the current PR scope but would be really nice to have the absolute url here.

design-documents/graph-ql/coverage/add-items-to-cart/AddSimpleProductToCart.graphqls
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type AddSimpleProductsToCartOutput {
+    cart: Cart
+}
+
+type Cart {
+    id: String
+    items: [CartItemInterface]
+}
```
masodu: Missing the cart totals, currency, updated dates and all the other information. I know that this is not in the scope but my suggestion is to add all these data to cart definition ASAP to make things clear. Btw how can a client be sure that the cart he is currently updating is the last one? The use case here is when the same cart is updated by the `end user` and also by a `back office`.

design-documents/graph-ql/coverage/add-items-to-cart.md
masodu: For a server less architecture as we have in `CIF` this might not work OOTB.

design-documents/graph-ql/coverage/CouponOperations.graphqls
```diff
@@ -0,0 +1,30 @@
...
+type Cart {
+    applied_coupon: AppliedCoupon
+}
+
+type AppliedCoupon {
+    # Wrapper allows for future extension of coupon info
+    code: String!
+}
+
+input CancelCouponForCartInput {
```
misha-kotov: suggest "cancel coupon on cart", as in "CancelCouponOnCartInput", "CancelCouponOnCartOutput"

design-documents/graph-ql/coverage/add-items-to-cart.md
DrewML: How about:

```
As a Magento developer, I need to manipulate the shopping cart via GraphQL so that I can programmatically create orders on behalf of a shopper.
```

Thoughts?

design-documents/graph-ql/coverage/add-items-to-cart.md
paliarush: The decision is to have Synchronous implementation by default. Asynchronous implementation may be added later on the framework level as it was for REST.

design-documents/graph-ql/coverage/CartPrices.graphqls
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type Cart {
+    prices: CartTotals
+    addresses: [CartAddress]
+    # additional fields will be added later
+}
+
+type CartAddress {
+    pices: CartAddressTotals
```
masodu: Is this the CartAddressPrices? Also, note the small typo, the field name should be `prices`

masodu: Why plural?

design-documents/graph-ql/coverage/CartPrices.graphqls
```diff
@@ -0,0 +1,57 @@
+interface CartItemInterface {
+    prices: CartItemTotals
```
masodu: Where is the CartItemTotals defined?

design-documents/graph-ql/coverage/CartPrices.graphqls
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type Cart {
+    prices: CartTotals
```
masodu: Where is the CartTotals defined?

design-documents/graph-ql/coverage/CartPrices.graphqls
paliarush: Good catch, I called the type `CartItemTotals` first and then renamed it to `CartItemPrices`, forgot to change here.

design-documents/graph-ql/coverage/CartPrices.graphqls
paliarush: I think plural makes sense because there are multiple price types stored in this field. 
Do you strongly believe it should be singular in this case?

masodu: Not really, but there's this small inconsistency where you have a plural to designate lists but here is a single object. Why don't name it totals?

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -0,0 +1,103 @@
...
+    is_guest: Bool!
+    email: String!
+    prefix: String
+    first_name: String!
+    last_name: String!
+    middle_name: String
+    suffix: String
+    gender: GenderEnum
+    date_of_birth: String
+    vat_number: String # Do we need it at all on storefront? Do we need more details
```
Vinai: The presence of a vat_number can influence the rendering in many ways, so I'd say yes.

design-documents/graph-ql/coverage/add-items-to-cart/AddConfigurableProductToCart.graphqls
Vinai: In the past I've always found it much easier to supply a sku rather than a list of options that map to a simple product sku. I agree with @paales that usually the sku is already known on the frontend.

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+    firstname: String!
+    lastname: String!
+    company: String
+    street: [String!]!
+    city: String!
+    region: String
+    postcode: String
+    country_code: String!
+    telephone: String!
+    fax: String
```
misha-kotov: do we really need fax anymore? I would get rid of it if we can do it without breaking things

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+    addShippingAddressToCart(input: AddShippingAddressToCartInput): AddShippingAddressToCartOutput
+}
+
+input AddBillingAddressToCartInput {
+    address_id: Int
+    address: CartAddressInput!
+}
+
+input AddShippingAddressToCartInput {
+    address_id: Int # Can be provided in one-page checkout and is required for multi-shipping checkout
```
misha-kotov: Have we accounted for multishipping scenarios here? I don't yet see how we tie cart items with shipping addresses, in case there are multiple

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
+type Query {
+    addBillingAddressToCart(input: AddBillingAddressToCartInput): AddBillingAddressToCartOutput
+    addShippingAddressToCart(input: AddShippingAddressToCartInput): AddShippingAddressToCartOutput
+}
+
+input AddBillingAddressToCartInput {
```
misha-kotov: Suggestion: rename to setBillingAddressOnCart (to show that there can only be one)

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+    address_id: Int
+    address: CartAddressInput!
+}
+
+input AddShippingAddressToCartInput {
+    address_id: Int # Can be provided in one-page checkout and is required for multi-shipping checkout
+    address: CartAddressInput
+    shipping_method_code: String # Should be on address level for multi-shipping checkout
```
misha-kotov: Needs to be a separate mutation

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+}
+
+input CartAddressInput {
+    firstname: String!
```
irenelagno: How are we going to add customer address attributes?

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+type AddBillingAddressToCartOutput {
+    cart: Cart!
+}
+
+type Cart {
+    addresses: [CartAddress]!
+}
+
+type CartAddress {
+    firstname: String!
```
irenelagno: how are we going to set information about customer address attributes

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+    code: String
+    label: String
+}
+
+enum AdressTypeEnum {
+    SHIPPING
+    BILLING
+}
+
+type CheckoutShippingMethod {
```
irenelagno: I think we also need to add information about the error message, it can be displayed if in the system configuration we choose carriers/dhl/showmethod = 1

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
```diff
@@ -0,0 +1,84 @@
...
+    code: String!
+    label: String!
```
irenelagno: Also, online shipping methods can have different rates, how are we going to display this information?

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -0,0 +1,58 @@
...
+
+enum GenderEnum {
+    MALE
+    FEMALE
+}
+
+type CheckoutPaymentMethod {
+    code: String!
+    label: String!
+    # Do we need more info from quote_payment DB table?
```
misha-kotov: need to add "amount" or "balance" for things like store credit, gift card, reward pts

misha-kotov: Add sort order

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
misha-kotov: case code/label become optional if error message is displayed (setting "show method if not applicable")

design-documents/graph-ql/coverage/Cart.graphqls
misha-kotov: handle payment method-specific field such as "instructions", "payable to", "send check to", etc

misha-kotov: consider "terms and conditions" and "billing agreements"

misha-kotov: consider gift messaging, gift wrapping, gift receipt

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -0,0 +1,58 @@
+type Query {
```
irenelagno: please add information about shipping address if we add an item from gift registry;
how can we set gift messages, gift wrapping, printed cart information;
how can we set checkout agreements
`<extension_attributes for="Magento\Quote\Api\Data\PaymentInterface">
        
        <attribute code="agreement_ids" type="string[]" />
    </extension_attributes>`

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: Done

paliarush: Done

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
paliarush: Done

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
paliarush: Done

design-documents/graph-ql/coverage/CartAddressOperations.graphqls
paliarush: Done

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -0,0 +1,59 @@
...
+    customer_notes: String
+
+    gift_cards_amount_used: Money
+    applied_gift_cards: [CartGiftCard]
+
+    is_multishipping: Boolean!
+    is_virtual: Boolean!
+}
+
+type CheckoutCustomer {
```
real34: It would be great to use polymorphism here.

```
interface CheckoutCustomer {
  # the mandatory fields
}

type GuestCheckoutCustomer implements CheckoutCustomer {
  # …
  middle_name: String
}
type LoggedInCheckoutCustomer implements CheckoutCustomer {
  # …
  middle_name: String!
}
```

React components could then require different fragments depending on the type

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -0,0 +1,59 @@
...
+    is_guest: Boolean!
+    email: String!
+    prefix: String
+    first_name: String!
+    last_name: String!
+    middle_name: String
+    suffix: String
+    gender: GenderEnum
+    date_of_birth: String
```
real34: Are there any plans for a scalar `Date` type in the core?

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -0,0 +1,59 @@
...
+enum GenderEnum {
+    MALE
+    FEMALE
+}
+
+type CheckoutPaymentMethod {
+    code: String!
+    label: String!
+    balance: Money
+    sort_order: Int
```
real34: Is there a use case for exposing this value (vs having sort params in queries returning this type)?

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: We do not have plans for that, but I know that this topic was discussed in scope of Webonyx project: https://github.com/webonyx/graphql-php/issues/228

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: We thought of a use case when the client app needs to honor sort order of available payment methods during rendering.

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: Good idea in general. I don't understand though, why `middle_name` should be required for the registered customer. The only difference I see right now is that we can eliminate `is_guest` field in this way. 

Please clarify which exact fields should be different between guest and customer implementations.

real34: I agree that the `middle_name` is not a good example. It is for mandatory user information upon registration, such as `email` for instance.

The idea of having such a structure, is to open rooms for extension in the shop. For instance in a B2B context, a `LoggedInCheckoutCustomer` may have a `sales_representative: SalesPerson!` attached which would not be available for guests.
Developers would then only extend the `LoggedInCheckoutCustomer` type, making implementation more explicit.

design-documents/graph-ql/coverage/Cart.graphqls
real34: I think this pattern is better to be delegated to the server logic. By default `available_payment_methods` would return sorted values.

If a use case where methods should be sorted differently, it could lead to a different signature with an optional param:

```
available_payment_methods(sortDesc: Boolean)
```

design-documents/graph-ql/coverage/add-items-to-cart/AddBundleProductToCart.graphqls
```diff
@@ -0,0 +1,59 @@
...
+    updateBundleProductsInCart(input: UpdateBundleProductsInCartInput): UpdateBundleProductsInCartOutput
+}
+
+input UpdateBundleProductsInCartInput {
+    cart_id: String!
+    cartItems: [UpdateBundleProductCartItemInput!]!
+}
+
+input UpdateBundleProductCartItemInput {
+    details: UpdateCartItemDetailsInput!
```
naydav: Maybe it will be more better to use `data` instead of `details`
```
data: CartItemInput!
```

design-documents/graph-ql/coverage/add-items-to-cart/AddSimpleProductToCart.graphqls
```diff
@@ -0,0 +1,78 @@
...
+    id: Int!
+    label: String!
+    type: String!
+    values: [SelectedCustomizableOptionValue!]!
+    sort_order: Int!
+}
+
+type SelectedCustomizableOptionValue {
+    id: Int
+    label: String!
```
naydav: What is `label` for `SelectedCustomizableOptionValue `?
Is the value entered / selected by the user?

design-documents/graph-ql/coverage/add-items-to-cart/AddSimpleProductToCart.graphqls
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type SimpleCartItem implements CartItemInterface {
+    customizable_options: [SelectedCustomizableOption]
+}
+
+type SelectedCustomizableOption {
+    id: Int!
+    label: String!
+    type: String!
```
naydav: Also, need to add data about `Required` flag

design-documents/graph-ql/coverage/add-items-to-cart/AddSimpleProductToCart.graphqls
```diff
@@ -0,0 +1,78 @@
...
+    values: [SelectedCustomizableOptionValue!]!
+    sort_order: Int!
+}
+
+type SelectedCustomizableOptionValue {
+    id: Int
+    label: String!
+    price: CartItemSelectedOptionValuePrice!
```
naydav: Do we need price info?

Maybe id and value will be enough for the response about selected options
```
{
   "id": 14,
   "value": "test",                    
 }
```

design-documents/graph-ql/coverage/add-items-to-cart/AddSimpleProductToCart.graphqls
```diff
@@ -0,0 +1,78 @@
...
+    updateSimpleProductsInCart(input: UpdateSimpleProductsInCartInput): UpdateSimpleProductsInCartOutput
+}
+
+input UpdateSimpleProductsInCartInput {
+    cart_id: String!
+    cartItems: [UpdateSimpleProductCartItemInput!]!
+}
+
+input UpdateSimpleProductCartItemInput {
+    details: UpdateCartItemDetailsInput!
```
naydav: Why do we need one more level like `UpdateCartItemDetailsInput`

Now request is
```
cartItems: [
        {
          data: {
            sku: "product_dynamic_24494", 
            qty: 1
          }, 
          customizable_options: [
            {id: 5, value: "test"},
           ]
        }
      ]
```

Can we use next?
```
cartItems: [
        {
          sku: "product_dynamic_24494", 
          qty: 1,
          customizable_options: [
            {id: 5, value: "test"},
           ]
        }
      ]
``

buskamuza:
Merged the initial document.
Please process any remaining questions in smaller documents that modify the original one.

PR #18, opened 2018-08-15
Proposal for adding a composer update lookahead to resolve root conflicts

pdohogne-magento:
With Magento 2.3, we needed to [add additional steps to the upgrade process](https://devdocs.magento.com/guides/v2.3/comp-mgr/cli/cli-upgrade.html) due to incompatibilities with dependencies in the root `composer.json` file.  Without the fixed requirements, the `composer update` command fails to install the new version of Magento because it has conflicts with the root requirements, and 2.3 code can't fix those requirements before it is installed because it would need to be installed in order to run the code.

In the future, we are likely to encounter this issue again, especially during upgrades where we drop support for old PHP versions (PHP 7.1 will soon stop receiving security patches, for example).  To avoid needing to add more manual steps for each version, we should add hooks to the upgrade process so any installations that start at 2.3 or higher check the target version for changes that need to be made before it is installed.

See the changed file for the proposed solution.

### DIFF
--- a/proposals/composer-update-lookahead.md
+++ b/proposals/composer-update-lookahead.md
@@ -1,0 +1,48 @@
+# Add Lookahead for Pre-`composer update` Changes
+> _Related JIRA ticket: [MAGETWO-94153](https://jira.corp.magento.com/browse/MAGETWO-94153)_
+## Context
+With Magento 2.3, we needed to [add additional steps to the upgrade process](https://devdocs.magento.com/guides/v2.3/comp-mgr/cli/cli-upgrade.html) due to incompatibilities with dependencies in the root `composer.json` file.  Without the fixed requirements, the `composer update` command fails to install the new version of Magento because it has conflicts with the root requirements, and 2.3 code can't fix those requirements before it is installed because it would need to be installed in order to run the code.
+
+In the future, we are likely to encounter this issue again, especially during upgrades where we drop support for old PHP versions (PHP 7.1 will soon stop receiving security patches, for example).  To avoid needing to add more manual steps for each version, we should add hooks to the upgrade process so any installations that start at 2.3 or higher check the target version for changes that need to be made before it is installed.
+
+### Composer Functionality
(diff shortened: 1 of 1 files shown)


pdohogne-magento:
Review requested: @buskamuza

pdohogne-magento:
This is all possible using a Composer plugin; I have a proof-of-concept attaching custom parameters and composer modification functionality to the `update` command running locally.

Question: Is it necessary to address non-`composer.json` changes?  The decision to include `magento/updater` in the 2.3 upgrade script was not a requirement, and I'm not sure trying to include that functionality in this plugin would make sense (and it would make the implementation much more complicated than handling just `composer.json` itself).  Non-composer changes shouldn't break any upgrades unless they are in autoloaded files, and in our current project packages the only non-`composer.json` content we have is `magento/updater`, which isn't autoloaded.

pdohogne-magento:
I think this should live in its own git repo and be versioned independently, as it should have no problems being compatible with earlier Magento versions, plus it's not an actual Magento module.  Do you see any issue with that?  By necessity it doesn't have any actual composer requirements on any other Magento packages.

proposals/composer-update-lookahead.md
```diff
@@ -0,0 +1,48 @@
...
+ - Any changes to be made come from a trusted/controlled location
+ - Upgrades should be as minimal as possible, ideally only making changes to the `composer.json` file (adding/modifying/removing specific entries, never whole sections)
+ - Dry-run mode to preview changes that will be made (and also output changes that were made in non-dry-run mode)
+ - Avoid clobbering any non-Magento changes that may have been made to the same values
+ - Ability to skip the lookahead script without disabling any other scripts that might be installed
+
+## Implementation Strategy
+There are a few hiccups in the actual implementation.
+- The upgrade needs to be able to handle starting at any previous Magento version, each of which could have different values that need to be added, changed, or removed.
+- There are potentially non-Composer updates that need to be made, such as `magento/updater`, which exists in the top-level Magento directory but is not its own composer package.
```
buskamuza: Please, take into account a case when Updater application is absent.

PR #19, opened 2018-08-16
[Proposal] Declarative Theme Media Breakpoints

davemacaulay:
Currently we declare theme breakpoints through LESS variables, notably `@screen__m` declares the breakpoint at which the design should change to a mobile design.

Page Builder requires more dynamic generation of styles due to the dynamic nature of how the content is created. We have background image functionality which requires the generation of CSS to provide a high performant solution, due to this we need to be able to determine the current themes mobile breakpoint.

### DIFF
--- a/proposals/declarative-theme-media-breakpoints.md
+++ b/proposals/declarative-theme-media-breakpoints.md
@@ -33,78 +33,34 @@ For instance the file for both `luma` and `blank` contains the following nodes:
 
 We would look to unify the breakpoint configuration for all themes under a single node and not specific for a particular module. 
 
-We have 3 different options for this proposal
+### Long Term Solution
 
-### Define full breakpoint
+In the long term we'd like to be able to support population of LESS variables from the themes XML configuration. This allows for the greater Magento system to have an understanding of when various different content should be displayed, this could be represented in the themes XML as follows:
 
-```xml
-<breakpoints>
-    <breakpoint name="large">@media(-webkit-min-device-pixel-ratio:1.5),(min--moz-device-pixel-ratio:1.5),(-o-min-device-pixel-ratio:3/2),(min-resolution:1.5dppx)</breakpoint>
-    <breakpoint name="desktop">@media all and (min-width: 960px) and (max-width: 1199px)</breakpoint>
-    <breakpoint name="tablet">@media all and (min-width: 767px)</breakpoint>
-    <breakpoint name="mobile">@media all and (min-width: 480px)</breakpoint>
-</breakpoints>
+```
+<theme_variables>
+    <var name="screen__m">767px</var>
+    <var name="default_font_size">14px</var>
+</theme_variables>
 ```
 
-**Pros:**
+These variables names would be converted over to `@screen__m` and `@default_font_size` respectively.
 
-- Greater control over the point at which we display mobile content
+### Short Term Page Builder Specific Solution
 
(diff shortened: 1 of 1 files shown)


proposals/declarative-theme-media-breakpoints.md
```diff
@@ -0,0 +1,67 @@
...
+These variables names would be converted over to `@screen__m` and `@default_font_size` respectively.
+
+### Short Term Page Builder Specific Solution
+
+The long term solution has a large requirement of resource and careful design to ensure it's implementation fits it's purpose correctly.
+
+Page Builder requires an understanding of break points now ,_similar to how Magento_Catalog does currently_, due to this we can introduce new configuration specific for our module, within the `blank` and `luma` themes` as follows:
+
+```
+<vars module="Magento_PageBuilder">
```
melnikovi: We can have this configuration in the PageBuilder module and we allow to override it from the theme.

For the future we should consider generic solution that would allow us to pass different variables from configuration to LESS(option 1). We need to give good thought to option 1 configuration and implementation as well as receive feedback from community. Option 2 seems like good solution for now, that we can deprecate when we implement option 1.

nathanjosiah: It's not currently implemented for the specific issue @davemacaulay is trying to solve.

nathanjosiah: After talking in Slack I think we are on the same page. I misunderstood your comment.

irenelagno: I think if we can solve our problem using Option 2 it is better to proceed with it.

buskamuza:
With Option 2, as I understand you want to put breakpoints configuration in PageBuilder module. How this will work for a theme developer scenario?

1. A new theme defines custom breakpoints for Catalog because theme developer can find an example in `lima/blank` and it's Magento Open Source
2. How the same theme developer will know that he/she should also specify same breakpoints in PageBuilder section? PageBuilder is Magento Commerce, so not every theme developer will be even thinking about its support.

Also, this approach will require duplication of breakpoints: in Catalog section and in PageBuilder section. This will lead to mistakes. A workaround can be a static test probably.

proposals/declarative-theme-media-breakpoints.md
```diff
@@ -0,0 +1,69 @@
...
+</vars>
+```
+
+This allows us to understand the mobile breakpoint and correctly render our dynamic CSS. Utilising `view.xml` file we can also add defaults to our module within `Magento_PageBuilder/etc/view.xml` to ensure that mobile background images are always displayed.
+
+**Note:** would look to implement the short term solution to resolve this issue with Page Builder but with a  mindset to consider the long term solution in a later release.
+
+## Overall Concerns
+
+This does mean, in the current situation, we have 2 sources of truth for breakpoints within a theme. Those defined within the themes LESS and those defined in the `etc/view.xml`. We do already have this issue with the `Magento_Catalog` module declaring these breakpoints for the product gallery.
```
buskamuza: As I can see, it will be duplicated in more places:
1. https://github.com/magento/magento2/blob/2.3-develop/lib/web/css/source/lib/variables/_responsive.less
2. In the theme's `etc/view.xml` for Catalog
3. In the theme's `etc/view.xml` for PageBuilder (for custom theme)

PR #20, opened 2018-08-23
Proposal for extensible objects and models

paliarush:
The idea was originally proposed by @maghamed


buskamuza:
Needs explanation of why specific one should be preferred and another one deprecated (pros/cons).

PR #23, opened 2018-08-28
Add proposal to deprecate JS Component mixins

DrewML:
Originally from https://github.com/magento/magento2/issues/17842

### DIFF
--- a/proposals/deprecate-js-component-mixins.md
+++ b/proposals/deprecate-js-component-mixins.md
@@ -1,0 +1,49 @@
+# Deprecate JS Component Mixins
+
+## Context
+
+Magento 2 uses the term `mixins` to define 2 different concepts:
+
+1. [RequireJS Mixins](https://devdocs.magento.com/guides/v2.2/javascript-dev-guide/javascript/js_mixins.html)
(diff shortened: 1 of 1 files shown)


Igloczek:
This is the first time I see the second meaning of `mixins`, so from my perspective would be better to deprecate, rather than documenting it.

DrewML:
Will be discussed at the September 5th arch meeting: https://github.com/magento/architecture/wiki/September-5,-2018

Vinai:
Moar unscientific research: https://twitter.com/VinaiKopp/status/1039540483691827200

DrewML:
Update: Still pending running some static analysis to verify lack of usage. Will be following up on that in the next week

DrewML:
We're very likely going forward with this, but I'm just waiting on the scan over popular extensions to verify there aren't use-cases we didn't consider

DrewML:
Confirmed none of the large extensions in the marketplace are using this functionality, and have not found anyone in the community relying on this functionality.

We'll move forward with the deprecation. The work will likely need to be done in concert with #33 as the Fotorama implementation is currently the only consumer of this undocumented API.

PR #25, opened 2018-09-13
Proposal for MDE (Magento Developed Extension) project structure

paliarush:


design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+    "repositories": [
+        {
+            "type": "path",
+            "url": "./<mde-name>/app/code/*/*"
+        }
+    ]
+    ```
+1. Add dependencies on all relevant MDE modules to the `require` section of `<project_root>/magento2ce/composer.json`. Use `*` as a package version
+1. Run `composer update`, which will create symlinks to the requested MDE modules inside of `<project_root>/magento2ce/vendor`. 
+   On Windows environment packages will be copied by composer (instead of being symlinked), that's why on Windows hosts it is better to link MDE using EE linking tool, or perform linking inside of linux VM
```
buskamuza: "EE linking tool" - please include the reference or add explanation to what's this

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+
+Currently, MSI is developed in a fork of main Magento repository, while Page Builder is developed in a separate repository (like B2B).
+
+## MDE project structure for Git-based installations
+
+MDE project must be structured as Magento B2B:
+ 1. Modules must be placed under `app/code`
+ 1. Each module must have valid `composer.json` with no version specified. Package version must be generated and added to the `composer.json` during release publication
+ 1. No core modules must be committed to the MDE repository
+ 1. Unit and MFTF tests must be distributed between modules (since they are modular). All other tests must be put inside `dev/tests` like it is done in Magento Open Source repository (because these testing frameworks do not support modularity)
```
buskamuza: How other tests are run on the current code base of MSI?
It's preferable to not go back to the old-style approach if possible, and keep tests in modules.

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
+# Overview
+
+There is a need to have unified approach to structuring the Git-based project while developing MDEs (Magento Developed Extensions), like Page Builder and MSI.
+
+Currently, MSI is developed in a fork of main Magento repository, while Page Builder is developed in a separate repository (like B2B).
+
+## MDE project structure for Git-based installations
+
+MDE project must be structured as Magento B2B:
+ 1. Modules must be placed under `app/code`
```
buskamuza: Is it only because of non-modular tests?
Based on the previous agreement, the directory structure should be:
```
<repo_root>/
           <Module1>
                Test/
                    Unit/
                    Integration/
                    Mftf/
                               TestSuite/
                               composer.json
           <Module2>
           <Module1SampleData>
           <Module2SampleData>
```

So there should be no `app/code`.

See https://wiki.corp.magento.com/display/MAGE2/Bundled+Extensions+Repository+Structure+--+HLD (internal link).

design-documents/mde-project-structure.md
paliarush: @maghamed could you please provide some insight into this?

design-documents/mde-project-structure.md
paliarush: I would say this is because of EE linking tool

vkublytskyi: I agree with @buskamuza and we already made a lot of effort to adopt the described structure (without app/code folder) or prepare source code for simple migration in future. As an example, all MSI Inventory modules contains integration tests inside Test/Integration modules. This is not resolve issue with integration tests modularity but make it more simple in future.

design-documents/mde-project-structure.md
vkublytskyi: In MSI all modules are placed inside corresponding module directories. All phpunit.xml files in mainline already contain configurations to run such tests. As far as I know PageBuilder team also utilize this approach (e.g. https://github.com/magento/magento2/blob/2.3-develop/dev/tests/integration/phpunit.xml.dist#L24)

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+ 1. No core modules must be committed to the MDE repository
+ 1. Unit and MFTF tests must be distributed between modules (since they are modular). All other tests must be put inside `dev/tests` like it is done in Magento Open Source repository (because these testing frameworks do not support modularity)
+
+## Git-based project linking for local development
+
+Preferred way of linking the project is using [Composer path repository](https://getcomposer.org/doc/05-repositories.md#path).
+
+1. Check out magento 2 open source edition from Git into `<project_root>/magento2ce`
+1. Check out MDE into `<project_root>/magento2ce/<mde-name>`
+1. Add path composer repository declaration to `<project_root>/magento2ce/composer.json`:
```
vkublytskyi: It will create bad developer experience. If contributor clones both core and MDE and would like to contribute in both she will have modified uncommited changes in composer.json file. Composer.lock file also should be managed separately. I believe we will have a tons of cases when composer.kson and composer.lock files will be modified in PRs unintentional.

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+    ```json
+    "minimum-stability": "dev",
+    "repositories": [
+        {
+            "type": "path",
+            "url": "./<mde-name>/app/code/*/*"
+        }
+    ]
+    ```
+1. Add dependencies on all relevant MDE modules to the `require` section of `<project_root>/magento2ce/composer.json`. Use `*` as a package version
```
vkublytskyi: MSI has more than 30 modules. Because of high MSI modularity, there is no single module in a repo that will install all other. For the release process, we resolved this with an introduction of additional composer metapackage that includes all modules (API, implementation, UI and admin modules).

We need another approach or improve this one with an introduction of standardized composer metapackage structure or providing a tool for automatic modification of composer.json (but as mentioned in other comment, I think that we need to avoid this for git-based installations)

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+1. Run `composer update`, which will create symlinks to the requested MDE modules inside of `<project_root>/magento2ce/vendor`. 
+   On Windows environment packages will be copied by composer (instead of being symlinked), that's why on Windows hosts it is better to link MDE using [EE linking tool](https://github.com/magento/magento2ee/blob/2.3-develop/dev/tools/build-ee.php), or perform linking inside of linux VM
+1. In case Magento has already been installed, the new modules must be enabled and setup upgrade executed
+
+
+## Alternative options
+
+1. Link MDE using the same [script](https://github.com/magento/magento2ee/blob/2.3-develop/dev/tools/build-ee.php) which is used for linking EE (the tool should be open sourced)
+2. Manually create symlinks for each MDE modules to `<project_root>/magento2ce/app/code`
```
vkublytskyi: This will not work if MDE declares composer dependency that not listed in core composer file.

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+        }
+    ]
+    ```
+1. Add dependencies on all relevant MDE modules to the `require` section of `<project_root>/magento2ce/composer.json`. Use `*` as a package version
+1. Run `composer update`, which will create symlinks to the requested MDE modules inside of `<project_root>/magento2ce/vendor`. 
+   On Windows environment packages will be copied by composer (instead of being symlinked), that's why on Windows hosts it is better to link MDE using [EE linking tool](https://github.com/magento/magento2ee/blob/2.3-develop/dev/tools/build-ee.php), or perform linking inside of linux VM
+1. In case Magento has already been installed, the new modules must be enabled and setup upgrade executed
+
+
+## Alternative options
```
vkublytskyi: We may consider also an implementation of composer plugin which will be listed in the core as dev dependency, have own config file which will be listed in Git ignore list and will be able to install MDE dependencies, add MDEs classes and register.php files to PHP autoload.

design-documents/mde-project-structure.md
```diff
@@ -0,0 +1,41 @@
...
+        {
+            "type": "path",
+            "url": "./<mde-name>/app/code/*/*"
+        }
+    ]
+    ```
+1. Add dependencies on all relevant MDE modules to the `require` section of `<project_root>/magento2ce/composer.json`. Use `*` as a package version
+1. Run `composer update`, which will create symlinks to the requested MDE modules inside of `<project_root>/magento2ce/vendor`. 
+   On Windows environment packages will be copied by composer (instead of being symlinked), that's why on Windows hosts it is better to link MDE using [EE linking tool](https://github.com/magento/magento2ee/blob/2.3-develop/dev/tools/build-ee.php), or perform linking inside of linux VM
+1. In case Magento has already been installed, the new modules must be enabled and setup upgrade executed
```
vkublytskyi: Seems to match manual steps, especially for community contributors who might switch between different  MDE development

design-documents/mde-project-structure.md
paliarush: If the fixture is defined in a module, will it be accessible by tests from the other module?

design-documents/mde-project-structure.md
paliarush: Any suggestions on alternative approach to prevent changes in composer.json?

vkublytskyi: As described in another comment. We might keep config in separate file ignored by Git and apply it by composer plugin or independent CLI tool (extract linker from Magento commerce and allow to read config files instead of passing everything from command arguments). To resolve an issue with dependencies of MDE we may utilize https://github.com/wikimedia/composer-merge-plugin

design-documents/mde-project-structure.md
vkublytskyi: I like idea to have modules in the root of repository. With this and meaningful module names visitor will get understanding what this project is about and what features provide. It should not be complex or lead to bad performance to adapt tool for such structure

design-documents/mde-project-structure.md
paliarush: This seems to be overcomplicated and not transparent. Alternative approach is to commit dependencies on all MDEs in the default composer.json in magento2ce/magento2ee. 
Then the developer will have to checkout all MDEs along with the main repository to start development.

PR #26, opened 2018-09-19
Proposal for GraphQL API for cross-platform rich content

zetlen:
:warning: **Addendum 2019/1/9: Added description of MVP workaround for upcoming PWA Studio releases.**

This is a proposal for supporting CMS in our GraphQL layer through a common interface type designed to eventually interoperate with PWA Studio, Adobe Experience Manager, and/or other systems where business users create rich content with a WYSIWYG. It proposes a GraphQL format for traversing a tree and learning metadata about node implementations. A proposed schema:

```graphql
interface VisualContentNode {
  role: String!
  depth: Int
  preload: [String]
  descendentRoles: [String]
  children: [VisualContentNode]
  html: String
}
```

An example request-response:

Request:

```graphql
fragment NodeMetadata on VisualContentNode {
  nodeId
  role
  depth
  preload
  descendentRoles
}

content(nodeId: 1) {
  ...NodeMetadata
  html
  children {
    ...NodeMetadata
  }
}
```

Response:

```json
{
  "content": {
    "nodeId": 1,
    "role": "Column",
    "depth": 4,
    "preload": [
      "banner1.jpg",
      "diagram.svg",
      "thumbnail001.jpg",
      "thumbnail002.jpg",
      "video.mp4"
    ],
    "descendentRoles": ["Banner", "Accordion", "Video", "Image", "Row", "Text"],
    "children": [
      {
        "role": "Banner",
        "depth": 0,
        "preload": ["banner1.jpg"],
        "descendentRoles": []
      },
      {
        "role": "Accordion",
        "depth": 3,
        "preload": [
          "diagram.svg",
          "thumbnail001.jpg",
          "thumbnail002.jpg",
          "video.mp4"
        ],
        "descendentRoles": ["Video", "Image", "Row"]
      },
      {
        "role": "Text",
        "depth": 0,
        "preload": [],
        "descendentRoles": []
      }
    ],
    "html": "Hello world! Text has already loaded."
  }
}
```


zetlen:
@kandy suggests that instead of recursing through a tree, the API represent the nodes as an array with reference pointers to child nodes, and the client reconstruct the tree as necessary. Then depth could be controlled with a filter on the depth property:

```graphql
{
  content(filter: { depth: { lt: 4 }}) {
     depth
     role
     childNodeIds
  }
}
```

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Challenges and FAQs
+
+- **Why do more network traffic on the first pageload?**
+
+  - Block implementations can provide placeholder content of arbitrary size and complexity; the only constraint is that this content must be "static", with no JS-driven behaviors. Blocks optimized for use on landing pages should have acceptable fallbacks implemented in the \`html\` field resolver.
+
+* **Why not use GraphQL introspection for this?**
+
+  - Per the GraphQL specification, the introspection query must provide only metadata, and not any data from resolvers. Implementing "partial introspection" via type tag names is a more efficient way of resolving recursive content nodes.
```
paliarush: Introspection will also be disabled in production.

zetlen: I should bring this up in a different issue, but we should at least permit the "[standard introspection query"](https://github.com/graphql/graphql-js/blob/master/src/utilities/introspectionQuery.js) in production, which GraphQL clients use to generate documentation. It should be enough to cache the results of this query, since it will be invariant in production.

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Proposal
+
+Add an interface to the GraphQL schema representing a recursive tree of PageBuilder content nodes. Each PageBuilder block type should add a concrete type to the GraphQL schema which implements this interface and adds additional fields for any required configuration. PWAs initially query only against the interface, and can therefore only receive the metadata. In the first few content loads, the PWA must make a "followup" request with more concrete query depth. As its cache grows, it does fewer and fewer of those requests.
```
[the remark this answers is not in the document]
zetlen: Because the list of potential implementations is unknown by the client. Extensions that add new PageBuilder blocks could add new implementations. Those implementations would have to include a GraphQL fragment that a PWA Studio PWA could use to add to its CMS query. That can happen at build time, but it should also be possible to do so at runtime.

melnikovi:
@zetlen please update based on our offline discussion.

zetlen:
@melnikovi Done!

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,331 @@
...
+                  ...PageBuilder_Text_Config
+                }
+            	children {
+                  ...NodeMetadata
+                }
+              }
+            }
+            ```
+
+        3.  Re-render the fully hydrated second level of component when that query returns results.
```
jedmao: Any way to avoid a re-render? I'm concerned about the page jumping as new content loads. I know you're not a fan of SSR, but it sounds like it would solve this problem.

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,331 @@
...
+
+        content {
+          ... on PageBuilder_Slider {
+            ...PageBuilder_Slider_Config
+          }
+        }
+        ```
+
+
+        3.  Re-render the fully hydrated Slider component when that query returns results.
```
jedmao: Ditto here. Always concerned about re-renders.

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,331 @@
...
+
+1.  In parallel:
+
+    1.  Render the `html` content by injecting it into a `RawHTML` component.
+
+    2.  Initiate preload requests for all the URLs in the `preload` field.
+
+    3.  In series:
+
+    4.  Dynamically load and inject the `PageBuilder_Slider` React Component. (The bundle system and service worker will load from their caches if possible.)
```
jedmao: How much content are we talking about here? I'm concerned about performance issues of putting too much responsibility on React to render things on the FE instead of on the first page load. I only speak from experience, because I've been 🔥  before on this point.

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Workaround Process
+
+1. Browser renders a `<RichContent id={id} contents={contents} />` element inside a component like `CmsBlock` or `ProductFullDetail`.
+2. After mount, `RichContent` component parses `props.contents` as a [DocumentFragment](https://www.google.com/search?q=documentfragment&oq=documentfragment&aqs=chrome.0.0l6.1759j0j7&sourceid=chrome&ie=UTF-8).
+3. Before initial render, `RichContent` component strips the fragment of dangerous HTML.
+4. For initial render, `RichContent` injects the sanitized fragment.
+5. After initial render, `RichContent` visits the DOM in the fragment to build a tree of PageBuilder rows, columns, widgets, and configurations.
+6. When the content tree is finished, `RichContent` dynamically imports the React components corresponding to all the PageBuilder widget types present in the content.
+7. When the runtime has loaded the components, `RichContent` re-renders the content tree as instances of live React components instead of HTML.
```
jedmao: Ditto.

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,331 @@
...
+            	children {
+                  ...NodeMetadata
+                }
+              }
+            }
+            ```
+
+        3.  Re-render the fully hydrated second level of component when that query returns results.
+
+Subsequent requests for any content which lists any of those components in its `descendentRoles` will automatically include those fragment and clauses in the `content` query. Therefore, subsequent requests will contain the necessary configurations on the first pass.
```
melnikovi: Would be good to measure performance on the prototype. I feel like network communication expenses would be higher than potential slowness of getting large JSON in one request. I think we wouldn't have a lot of content. This also makes caching less efficient.

proposals/rich-content-graphql-api.md
```diff
@@ -0,0 +1,331 @@
...
+2. After mount, `RichContent` component parses `props.contents` as a [DocumentFragment](https://www.google.com/search?q=documentfragment&oq=documentfragment&aqs=chrome.0.0l6.1759j0j7&sourceid=chrome&ie=UTF-8).
+3. Before initial render, `RichContent` component strips the fragment of dangerous HTML.
+4. For initial render, `RichContent` injects the sanitized fragment.
+5. After initial render, `RichContent` visits the DOM in the fragment to build a tree of PageBuilder rows, columns, widgets, and configurations.
+6. When the content tree is finished, `RichContent` dynamically imports the React components corresponding to all the PageBuilder widget types present in the content.
+7. When the runtime has loaded the components, `RichContent` re-renders the content tree as instances of live React components instead of HTML.
+
+### Considerations
+
+- The workaround involves DOM parsing and large string comparisons for update filters, which are both expensive operations which PWAs try to avoid.
```
melnikovi: Could you add pros and cons for each of the solutions?

melnikovi:
@zetlen could you provide more context on "common interface type designed to eventually interoperate with PWA Studio, Adobe Experience Manager, and/or other systems where business users create rich content"? How Adobe Experience Manager and other systems will use proposed API?

proposals/rich-content-graphql-api.md
belbiy: For Slider, we're talking about 3-4 Slides usually, with something like 10-20 is going to be a very rare edge case. 

But the thing is is that the Slide itself can contain more internal components, like buttons, headings, texts, etc.

proposals/rich-content-graphql-api.md
zetlen: In order to avoid this, the HTML fallbacks would have to have dimension. I think that the RichContent component could load a root PageBuilder stylesheet which would apply default dimensions.

As more PageBuilder components become "known" to the PWA runtime, it could add fragments to VisualContentNode queries like `...on PageBuilderGallery {}` etc. which would get more and more specific configuration, including dimension, on the first render...

proposals/rich-content-graphql-api.md
zetlen: I think we'd have to see how it performed. The HTML fallback would already be in the DOM node that the Slider component takes over, so I would hope the React reconciler would reuse as much of it as possible.

How much content are we talking? The point is, we don't know at runtime.

proposals/rich-content-graphql-api.md
zetlen: We should definitely measure the performance, and change approaches if it's terrible.

I think that caching might be better with GraphQL. If VisualContentNodes have unique IDs, then [Apollo cache implementations](https://github.com/convoyinc/apollo-cache-hermes) can store their representations incrementally. Then, the PWA would reuse references to the same content entity.

PR #29, opened 2018-09-24
Service isolation

antonkril:
The high-level design for "Service Isolation" epic. Continuation of Service Contracts track.


maksek:
@antonkril 
Can't find information about DB split per services.
Will Kubernetis be the only solution for developers?

phoenix-bjoern:
Thanks @antonkril for the great draft!
IMHO the catalog part should be extracted first, not checkout. The reason for that is simply the performance impact a scalable and lightweight catalog service would have. The checkout rarely is a bottleneck for an eCommerce platform while the responsiveness of the catalog is crucial.
Furthermore I believe that fast GraphQL response times are key for a PWA application and hopefully the catalog isolation would bring a dramatic increase of responsiveness (at least this should be one of the main goals of the work). Acceptable catalog response times for a PWA application are around 50-100ms (native Elasticsearch allows ~7-20ms).

kandy:
@ phoenix-bjoern, Service isolation goals do not include performance improvements, it's mostly about scalability. Actually, we will expect some small-scale performance degradation due to the use of inter-service communication.

phoenix-bjoern:
@kandy Maybe I'm on the wrong track here but IMHO service isolation should result in a clear definition of dependencies to ran that specific service (e.g. "catalog", "checkout", "crm", "cms"). That means that each service might only have be a small subset of what is "composed" in the vendor directory today as it's code base and which can be deployed (and scaled) independently. Services will talk to other services via web service APIs. Reducing the code base per service and moving the whole admin/frontend UI layer out of the "service layer" should definitely result in a faster and lightweight application. Or did I misunderstood the idea of @antonkril's approach?

design-documents/service-isolation.md
```diff
@@ -0,0 +1,248 @@
...
+    * ```magento instance:remove [name]``` - unregister a managed instance from the tool instance list
+    * ```magento instance:list``` - list registered remote instances
+    * ```magento instance:update``` - load a list of commands supported by the instance
+    * ```magento context:set [name]``` - select the default instance to be used in commands
+
+##### Instance builder
+
+With distributed deployment and split modules, `composer.json` management for different services will be complicated. A tool to build instance `composer.json` files for specific services based on enabled features or installed extensions should be built.
+
+Input:
```
paliarush: Input is missing

PR #31, opened 2018-09-27
OMCONNECT-91:  Add OPTIONS to Allowed Methods in Web Api

JacobBrownAustin:


design-documents/Add-Options-to-Allowed-Methods-in-Web-Api.md
```diff
@@ -0,0 +1,37 @@
...
+                    <xs:enumeration value="DELETE"/>
+                </xs:restriction>
+            </xs:simpleType>
+        </xs:attribute>
+```
+
+I propose it be changed to this:
+
+```
+        <xs:attribute name="method" use="required">
```
paliarush: Please elaborate on 
1. why exactly OPTIONS is required
2. who is going to implement it for each service we have? Or the idea to have such requests processed by the framework?

We currently have swagger documentation (`http://<myhost.com>/swagger`), which gives full description of each endpoint.

buskamuza:
@JacobBrownAustin , could you please review the question above?

design-documents/Add-Options-to-Allowed-Methods-in-Web-Api.md
melnikovi: I found this article from 2012 http://zacstewart.com/2012/04/14/http-options-method.html and author says that no one really uses OPTIONS method. Did you research on adoption OPTIONS method? Is it commonly used now? My concern is that even though it's part of specification, implementing it may not bring enough value because this part of specification is not well adapted and people wouldn't expect that it's available in our APIs. And we also have swagger documentation that developers know about by now.

melnikovi: Did more research and found this https://www.mnot.net/blog/2012/10/29/NO_OPTIONS. Do you have a use case when system makes CORS request to Magento?

melnikovi:
@JacobBrownAustin please respond to questions and reopen this PR.

design-documents/Add-Options-to-Allowed-Methods-in-Web-Api.md
isxam: We have returned to this proposal, because we got issue with our customization (to support OPTIONS) after upgrading Magento

The main cons in comments as i see - "You can use another method instead of OPTIONS"
But there cases when developer should implement API with predefined signature (by 3rd party services).
Usually in can happened during integration different types of Webhook API.
Different services use different ways to register/subscribe webhook url.
Usually it is request to API endpoint, which should response with special data (for example forward "hub.challenge" query or specific header value in our case).

So, i think, that we shouldn't artificially restrict using of HTTP methods, which specified in RFC of HTTP protocol.
There are many best practices and use cases how to use or don't use HTTP methods, but if we implements framework, we shouldn't limit ability to select solution, suitable for separate developer or system.

Below you can see list of frameworks which supports OPTIONS method handling.
If Magento has two parts "Framework"/"Ecommerce-specific functionality", so we can compare functionality of our framework with other frameworks.

JacobBrownAustin:
I need to reopen this issue because after the release of Magento 2.2.9 and 2.3.2,  some of our customers are unable to use the connector module because these XML schema requirements are causing it to fail to work. 

Previously,  the errors that I saw were only when running the static tests.  Now the errors are seen in `bin/magento setup:upgrade` and also in the incoming messages from OMS.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=31) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=31) before we can accept your contribution.<br/>

design-documents/Add-Options-to-Allowed-Methods-in-Web-Api.md
JacobBrownAustin: Hi @melnikovi and @paliarush .  Sorry I didn't reply to your questions previously.  When I originally wrote this issue,  it was just because of a Magento core static test failure when our modules were enabled.  Now,  the issue has grown because when using our modules in Magento 2.2.9 and 2.3.2,  these tests are causing fatal errors when doing upgrades or receiving incoming messages from OMS.  This is preventing some of our customers from upgrading to Magento 2.2.9 and 2.3.2 .  (@isxam has been working on a workaround for them.)

I will attempt to answer your questions.

**1. Why exactly OPTIONS is required?**
This is a requirement of the OMS API,  which has had this requirement for years.  You can read about it here.  
https://omsdocs.magento.com/en/integration/api/

> 
For every attempt to register a new application, a probe is made against the registered URL using the OPTIONS HTTP verb. The server should reply with an empty body including the following header: X-Magento-Service-Bus: *. This process is called “URL probing” and should prevent the Magento Integration Bus from making calls to URLs which are not meant to be used as Magento Integration Bus endpoints.

In that,  it says "OPTIONS HTTP verb",  but as we know,  OPTIONS isn't a verb,  it is a noun.  I'm sure they meant "HTTP method". 

**2.  who is going to implement it for each service we have? Or the idea to have such requests processed by the framework?**
No one needs to implement it,  other than people who require it.  For example,  in our modules,  we use OMS API,  which requires it.   So,  we implemented it in our module.  For other endpoints that do not need it,  they simply do not need to implement it.


**Did you research on adoption OPTIONS method? ** 
No,  we didn't research it.  It is just a requirement of OMS API.  I do not know if OMS team researched it before requiring it.  If you need them to answer,  I will try to get OMS architect to answer.  

**Is it commonly used now?**
Yes,  it is used in our module,  and as far as I know,  it has always been in use in our module.

**Do you have a use case when system makes CORS request to Magento?**
CORS also uses OPTIONS,  but our requirement for this change is unrelated to CORS.  If making this proposed change to Magento core has any security implications relating to our CORS compliance,  then we should consider them before approving this change.  I have no investigated how our change may affect CORS compliance and implementation.

PR #33, opened 2018-10-15
Proposal: Deprecate Fotorama

DrewML:
**Status** (1/4/19): This proposal is still in an active state, but is currently held up on some other priorities. Before this can be closed out and worked, a new library with a comparable feature set needs to be decided on.

[Rendered Proposal](https://github.com/magento/architecture/blob/replace-fotorama/design-documents/deprecate-fotorama/deprecate-fotorama.md)

I'd like to hear from Partners, SIs, and anyone else working on front-end for Magento stores on a day-to-day basis. I'd imagine there are some common slider/gallery libs in use - would love to hear suggestions.

### DIFF
--- a/design-documents/deprecate-fotorama/deprecate-fotorama.md
+++ b/design-documents/deprecate-fotorama/deprecate-fotorama.md
@@ -1,0 +1,61 @@
+# Overview
+
+On Magento 2 product pages, product images are displayed using a forked version of the [`Fotorama`](https://github.com/artpolikarpov/fotorama) library. Some unfortunate design decisions in this library, coupled with the amount of JS execution on product pages, makes the rendering of product photos a very obvious bottleneck for shoppers.
+
+Speaking with Magento SIs and Partners at various conferences, it sounds like a good number of agencies already rip out `Fotorama` as one of their default tasks when building out a new store for a client. This should be a key signal for us that this area needs improvement, and we should lean on them for their suggestions and overall expertise when going through this implementation.
+
+In this document, we'll propose deprecating `Fotorama`, and replacing it with an alternative that has better performance characteristics.
+
+## Discussion/Complaints about Fotorama Performance
+
+- https://github.com/magento/magento2/issues/14667
+- https://github.com/magento/magento2/issues/14667#issuecomment-415813263 (Comment from issue linked above diving into specific bottlenecks)
+- https://github.com/magento/magento2/issues/6018
+
+## Performance Problems with Fotorama
+
+### Library Size
+
+The [minified version of `Fotorama`](https://github.com/magento/magento2ce/blob/8a7f23e95/lib/web/fotorama/fotorama.min.js) in the 2.3 branch is `71.9 KB`. This is _massive_ for a photo gallery script. There are two problems related to the size of the library:
+
+- With gzipping turned on, the library is still roughly `21 KB` that has to be sent over the network. This isn't huge for those on desktops with reliable connections, but could take 1 second or more on a slower 3g connection.
+- [JavaScript parse time on slower smart phones is a massive bottleneck](https://medium.com/@addyosmani/the-cost-of-javascript-in-2018-7d8950fbb5d4). Because parsing happens against the uncompressed (but minified) JavaScript, the parse time is based on the `71.9 KB` figure. Using [`WebPageTest`](http://webpagetest.com/easy) to test on a [Moto 4G](https://www.gsmarena.com/motorola_moto_g4-8103.php), the parse time in [v8](https://en.wikipedia.org/wiki/Chrome_V8) took `398.6 ms`.
+
+### Execution Time
+
+Fotorama takes a significant amount of time to execute once the script has been loaded. In my most recent trace on [`WebPageTest`](http://webpagetest.com/easy) with a [Moto 4G](https://www.gsmarena.com/motorola_moto_g4-8103.php), it took around `336 ms` to execute. In the wild on real Magento stores, I've seen this take as long as `650 ms`, which is all time spent blocking the main thread.
+
+### Resource Discoverability
+
+Modern web browsers all have their own version of a [preload scanner](https://andydavies.me/blog/2013/10/22/how-the-browser-pre-loader-makes-pages-load-faster/). While the browser is spending time on script execution or other blocking tasks, a parser works in the background to discover all external resources, and then starts fetching them to get a head start.
+
+Unfortunately, Magento 2 uses a feature of `Fotorama` that works by being fed an array of images during instantiation of the widget on the client-side. [Magento renders this data from the server in an `x-magento-init` tag](https://github.com/magento/magento2ce/blob/8a7f23e9/app/code/Magento/Catalog/view/frontend/templates/product/view/gallery.phtml#L47), which the browser sees as nothing more than a string. Because no `<img />` tag is rendered from the server, the browser is unable to start preloading images early. The browser only begins fetching the main product photo after `Fotorama` has appended the first `<img />` tag to the DOM.
+
+Injecting the `<img />` tag(s) with JavaScript also has the unfortunate side-effect that, even if the images are already cached in a user's browser (repeat visit), they will still see the loading spinner for several seconds.
+
+### Late Execution
+
+The `Fotorama` library is initalized using an [`x-magento-init`](https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/javascript/js_init.html) configuration. This poses a problem, because Magento "JS Components" do not have a way of expressing priority, and their execution order is non-deterministic. In practice, their execution order typically comes down to what external scripts finish loading first (the core loops over widgets on pageload and loads their JS assets in parallel). Because of the size of `Fotorama`, the gallery is often one of the last scripts to execute.
+
+## Proposed Solution
+
+The proposed solution here is to deprecate usage of `Fotorama` in Magento 2, and replace its usage with a library that does not have the same pitfalls as `Fotorama`.
+
+### Requirements for a replacement
+
+- Library _must_ allow easily server rendering the first `<img />` tag, to ensure the resource is picked up by the preload scanner
+- Library should _not_ be more than `5 KB` minified and gzipped
+- Library _must_ support both desktop and mobile interactions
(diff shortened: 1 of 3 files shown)


johnhughes1984:
@DrewML There is a proof of concept (POC) gallery replacement using [flickity](https://flickity.metafizzy.co/) created by one of our devs at Fisheye: @robaimes - https://github.com/robaimes/module-product-gallery

Video demo: https://gfycat.com/WelltodoAppropriateCony

It's been on our list to create a full implementation from Rob's POC for some time.  I would also note the original focus of this extension was purely to replace fotorama due to it's inflexibility, extensibility and lack of responsive friendly nature as well as poor UX.  However the POC definitely has major rendering speed improvements over fotorama in addition the above.

Flickity does however not meet your 5KB gzipped / compressed limit so appreciate may not be viable (comes in at ~13KB - https://unpkg.com/flickity@2.1.2/dist/flickity.pkgd.min.js)

dankoz51:
but for some zoom and tablet type of features i think i would be more in favor of just using as much css and html as possible.  

Things like photorama (have not looked at flicket yet) are crated to make the html/css markup easier on an html coder, that way the create simple symantech html, and have it transformed for them, we dont need that wee need it to perform, who cares how hard the css and html are.

jackmcpickle:
Agreed with @brendanfalkowski on photoswipe. But for something more simple to cover swiping and sliding I use http://owlcarousel2.github.io/OwlCarousel2/

elioermini:
Agree we use swiper http://idangero.us/swiper/ and works well either desktop and touch, in addition we first render the static base image then once swiper loads will load the gallery, in this way the the first image loads same time the rest of the product content.

DrewML:
I expect this thread will get noisy, so I'm pulling out each of your lib suggestions to [this document](https://github.com/magento/architecture/blob/replace-fotorama/design-documents/deprecate-fotorama/proposed-alternatives.md).

Keep them coming, and I'll start looking through them soon to see which lib(s) fit the bill

AngelBS:
I agree on this proposal, and i'll like to suggest Slick.js, i've been using it for a while and it's easy to use and has a lot of options

http://kenwheeler.github.io/slick/

DrewML:
@AngelBS I've seen Slick used in the past, but it's _massive_ and won't fit the size requirements in this proposal. The minified version is actually 2kb larger than the version of Fotorama we're using

dankoz51:
> I expect this thread will get noisy, so I'm pulling out each of your lib suggestions to [this document](https://github.com/magento/architecture/blob/replace-fotorama/design-documents/deprecate-fotorama/proposed-alternatives.md).
> 
> Keep them coming, and I'll start looking through them soon to see which lib(s) fit the bill

I think we should add "None" as an option. I dont see why we need a lib to bloat the load and have support for lots of "features" that are never used. 

At the very least my requirement would be that it loads with no DOM modification at all, only register events for zoom, scroll, pinch etc.

Igloczek:
@jackmcpickle Owl is not actively maintained (it's not a first time, AFAIK they even changed maintainer like 2 years ago), weight too much, plus require jQuery (it's not in the requirements, but to make it really fast, will be nice to use something that can handle own shit, without asking colleagues for help :) )

jackmcpickle:
@Igloczek and @vasilii-b agreed. I throw my vote in with Slick which I have also used.

robaimes:
As a note, I wouldn't necessarily recommend Flickity for this, as it's not particularly well adapted to being used as a gallery. For example, it doesn't have a vertical slider option. The main advantage of why I chose to use Flickity was because it is highly customisable via CSS instead of JS. The POC posted [here](https://github.com/magento/architecture/pull/33#issuecomment-430041362) is just a prototype. I did make a build using slick, and it works just as well.

Not sure the 5KB requirement is entirely achievable but would it not make sense to use Slick, assuming it's still used in the new PagerBuilder module?

DrewML:
It seems like we're all in agreement that `Fotorama` needs to go, which is good! Pretty rare that everyone in a Github thread reaches a consensus 😂

The open question is obviously what a replacement is going to be. It's clear that what I'm missing right now is a minimum viable features list.

Because I don't work on client sites, I'm far from the best person to draft up a list of what use-cases we need to cover. Would someone here be willing to draft up a list to be included in this proposal? Maybe @brendanfalkowski or @robaimes?

DrewML:
> Not sure the 5KB requirement is entirely achievable 

That's very possible, and I'm open to changing it. _But_, I'm starting aggressively small, and we'll adjust if it's not possible. Tight constraints are important here to ensure we really think through this decision.

DrewML:
I pushed a change that added a [list of minimal viable features for a replacement lib](https://github.com/magento/architecture/blob/replace-fotorama/design-documents/deprecate-fotorama/minimum-viable-feature-set.md).

Kindly requesting review/suggestions from everyone that frequently deals in this area day to day 🙂

brendanfalkowski:
@DrewML — In my todo tabs, will get some ideas into that doc.

brendanfalkowski:
Ok, here's my take on "a client would actually approve this UX-wise" more than a minimal simulacra of most eComm sites. I couldn't get away with skipping the finer points in a real project and it doesn't help use to have an MVP with rudimentary functionality. If that's an aim, then I'd ship the easiest thing to remove because it'll be replaced regardless.

These are the questions and implementation points that would be addressed in a site's design plan. If the MVP skips these for the sake of minimalism, the behavior would feel incomplete relative to other sites.

I wouldn't aim for a single library that solves "product carousel + media swapper + media zoomer" in one. They're each simpler + smaller to spec individually (needing no library in some cases).

Some of these lists include "this or that" features where both approaches likely won't be implemented for the same client (ex: snapping or frameless carousel). But that decision satisfies a design constaint that differs across clients, and there isn't a "right" approach.

### Product List Carousel — Ex: recently viewed

- Keyboard accessible
- Items per frame "snapping" to container width per breakpoints in instance's config
- Frameless x-scrolling, so an item edge falls across the container hinting there's more
- Pagination as indicator of volume and nav element (dots or numbers)
- Nav left + right per frame by clicking button
- Nav left + right per frame by swiping across items
- Infinite carousel mode (or not) by wrapping when reaching the end
- Lazy-loading images as frame enters container (full HTML in DOM)
- Lazy-loading additional frames via AJAX for performance (first-frame only in DOM)

### Product Page — Primary + Secondary image switcher

- Keyboard accessible
- Render primary image from HTML
- Option to lazy-load secondary images
- Render all secondary images as multi-line list of thumbnails
- Render all secondary images as single-line carousel of thumbnails
- - Swipe to move along line
- - Scroll-x to move along line
- - Nav left + right button to move along line
- Swap secondary image to primary slot
- - By clicking secondary image
- - By swiping across primary image (1 item carousel)
- - By clicking buttons for nav left + right (1 item carousel)
- Option to require/disallow manipulating product options to swap/load alternative swatches. Depending on volume of swatches + media this either hurts or helps UX.
- Source for full-size primary + secondary images is processable by 3rd-party (ex: pin to Pinterest)

### Product Page — Zoom Viewer

- Keyboard accessible
- Click to expand (rarely: hover to expand, always feels clunky on touch devices)
- When expanded, render highest resolution available
- When expanded, supports pinch to zoom
- When expanded, supports (+) and (-) buttons to zoom
- When expanded, supports mouse wheel to zoom
- When expanded, supports double-tap to zoom
- When expanded, supports nav left + right buttons to cycle secondary images
- When expanded, supports swiping to cycle secondary images
- When expanded, supports infinite carousel to wrap last ➔ first image
- On load, preload the primary image at full resolution to improve perceived performance
- When expanded, preload N+1-3 full-size assets to improve perceived performance

DrewML:
@brendanfalkowski Thanks for doing that!

I'm a little bit confused about the section for `Product List Carousel`. I did not think that features like "Recently Viewed" and "Customers also Bought" used the Fotorama carousel - am I incorrect?

brendanfalkowski:
@DrewML — I don't think it is either.

In most real builds we have a generic carousel library, and would make the UI for "recently viewed carousel" and "PDP primary image carousel" using the same carousel library if possible.

I added those ideal specs for the "product list carousel" concept just because it might help to research that a carousel library selected for swiping across PDP images also be usable for another UI piece. Just easier for devs to lookup one library config.

guz-anton:
Hi @DrewML,
> ### Requirements for a replacement
>  - ...
>  - Should _not_ be initialized using the `data-mage-init` or `x-magento-init` functionality in core 

What are the concerns about initializing mechanism?
What will be the suggestion to do it instead of `data-mage-init`?

DrewML:
@guz-anton 	

> What are the concerns about initializing mechanism?
> What will be the suggestion to do it instead of data-mage-init?

For the gallery, I want to avoid using `mage-init` to initialize because it is one of the most critical pieces of above the fold content on a product page, and the order of `mage-init` widget execution is not deterministic.

To help illustrate, the screenshot below shows 2 different tabs from the same store, and the order that widgets were initialized. Note that the order is not the same between different page loads:

image

Initializing imperatively gives us more precise control over when execution of the widget begins.

kandy:
@DrewML, can I ask you to  describe the real problem? It is normal for asynchronous code to load in a different sequence. It is only important that the code be executed only when all dependencies are available, and the order of loading itself is not important.

DrewML:
@kandy Described in [this performance write-up](https://github.com/magento/magento2/issues/14667#issuecomment-415813263)

kandy:
@DrewML,   all that you say in the bug definitely make sense. But I personally think that it is will be the match valuable to improve the declarative mechanism of initialization instead of adding inline javascript

DrewML:
To those still watching: for now, I am going to withdrawal this proposal.

There were two primary motivators for me when I wrote the proposal:

1. Speeding up product page image performance
2. Improving DX

For 1, https://github.com/magento/magento2/commit/56e56ceee884a830e93b953ba34de348484ce5c0 greatly improved the performance of initial product page rendering, allowing the browser to begin fetching the product photo immediately. The work we're doing on bundling with [`baler`](https://github.com/DrewML/baler/) will continue to improve these numbers

With regards to DX, I've had a slight change of opinion. With `pwa-studio` being worked on as the future of Magento front-ends, I am more hesitant to make breaking changes to current themes unless the benefits are too hard to ignore. In this case, the benefits are more debatable, since those that don't want to use Fotorama have many free (and paid) alternatives for m2.

Thanks to everyone that participated and provided input on this!

PR #34, opened 2018-10-17
Add redeem method to the Gift card API

AlexMaxHorkun:
Redeem functionality is missing from the gift card API

### DIFF
--- a/design-documents/add-redeem-method.md
+++ b/design-documents/add-redeem-method.md
@@ -1,0 +1,15 @@
+## Add GiftCardAccountManagementInterface::redeem() method
+#### Why?
+Redeeming a gift card is an existing part of the gift card module functionality,
+API should reflect that and allow for client code to execute it and/or use it as Web API.
+#### Situation
+Right now "redeem" method only exists in the Giftcardaccount model and
+actually does not contain all the logic required to redeem a gift card - only Customer\Index constroller has all the logic -
+it would be much better if a service had a method for this functionality with all the required logic in it
+and to remove the logic from the controller by leaving only the GiftCardAccountManagementInterface::redeem() call.
+Also there's additional logic regarding gift card usages limits which is included in other methods of GiftCardAccountManagementInterface,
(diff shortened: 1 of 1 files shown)


melnikovi:
Let's introduce new interface for redeem method. Please confirm change with PO.

design-documents/add-redeem-method.md
```diff
@@ -0,0 +1,13 @@
+## Add GiftCardRedeemInterface interface as a service contract
+#### Why?
+Redeeming a gift card is an existing part of the gift card module functionality,
+API should reflect that and allow for client code to execute it and/or use it as Web API.
+#### Situation
+Right now "redeem" method only exists in the Giftcardaccount model and
+actually does not contain all the logic required to redeem a gift card - only Customer\Index constroller has all the logic -
+it would be much better if a service had a method for this functionality with all the required logic in it
+and to remove the logic from the controller by leaving only the GiftCardRedeemInterface::redeem() call.
```
melnikovi: Name should be noun. How about `GiftCardRedeemInterface`? I would prefer `RedeemInterface` because GiftCard is in the namespace and it's clear but might be better to keep it because other interfaces in this directory have it.

AlexMaxHorkun: @melnikovi I did name it GiftCardRedeemInterface, did you mean to write something else?

AlexMaxHorkun:
@melnikovi The PO approved this change

design-documents/add-redeem-method.md
melnikovi: I'm sorry, I meant to write `GiftCardRedeemableInterface` or `RedeemableInterface`.

AlexMaxHorkun: @melnikovi Maybe RedeemManagerInterface? RedeemableInterface sounds like an entity that can be redeemed

melnikovi: Let's call it GiftCardRedeemerInterface.

AlexMaxHorkun: ok

PR #35, opened 2018-10-18
Remove __sleep and __wakeup

AlexMaxHorkun:
Object serialization should be avoided

### DIFF
--- a/design-documents/Remove-__sleep-and-__wakeup.md
+++ b/design-documents/Remove-__sleep-and-__wakeup.md
@@ -1,0 +1,35 @@
+### Remove __sleep and __wakeup methods from our code
+#### Why?
+* Serializing objects requires classes to always consider serialization process
+and it is easy for multiple developers working on the same class to forget about it.
+* PHP serialization opens up our code to potential security issues (Remote code execution).
+* During serialization objects of classes that do not expect serialization may be serialized
+anyway causing unexpected behaviour.
+
+#### Current situation
+The only place we serialize objects (as far as I am aware of) is storing User and Acl objects
+in admin session for convenience during admin auth process. Indirect serialization caused
+us to add __sleep and __wakeup methods to ~33 classes.
+
+#### What has to be done
+##### Admin session
+I propose creating 2 interfaces in the Magento\Backend\Spi namespace:
+* SessionUserSerializerInterface
+* SessionAclSerializerInterface
+ 
+Both interfaces will have 2 methods:
+* serialize(User|Acl $user|$acl): array - which will extract given object's
+data into an array of scalars
+* unserialize(array $data): User|Acl - taking array of scalars
+and turning it into objects
+ 
+Only arrays of scalars will be stored in session. getUser, setUser, getAcl, setAcl methods
+must be added to the Magento\Backend\Model\Auth\Session class to prevent serializing
+objects into admin session.
+ 
+##### CodeMess rule
+Add a PHPMD rule which will be triggered by classes
+having methods called __wakeup and __sleep
+ 
+##### Existing code
+Remove __sleep and __wakeup from our code


orlangur:
> serialize(User|Acl $user|$acl): array - which will extract given object's
data into an array of scalars

Sounds like data hydrator rather than serializer, isn't it?

> Add a PHPMD rule which will be triggered by classes
having methods called __wakeup and __sleep

PHPCS is better suitable for this purpose as PHPMD rule has to look in every single token.

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
+### Remove __sleep and __wakeup methods from our code
+#### Why?
+* Serializing objects requires classes to always consider serialization process
```
kandy: Can be fixed by writing an integration test

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
...
+and it is easy for multiple developers working on the same class to forget about it.
+* PHP serialization opens up our code to potential security issues (Remote code execution).
+* During serialization objects of classes that do not expect serialization may be serialized
+anyway causing unexpected behaviour.
+
+#### Current situation
+The only place we serialize objects (as far as I am aware of) is storing User and Acl objects
```
kandy: Any object stored in the session should be serializable

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
+### Remove __sleep and __wakeup methods from our code
+#### Why?
+* Serializing objects requires classes to always consider serialization process
+and it is easy for multiple developers working on the same class to forget about it.
+* PHP serialization opens up our code to potential security issues (Remote code execution).
```
buskamuza: Unserialize has been already prohibited and removed whenever possible.

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
...
+anyway causing unexpected behaviour.
+
+#### Current situation
+The only place we serialize objects (as far as I am aware of) is storing User and Acl objects
+in admin session for convenience during admin auth process. Indirect serialization caused
+us to add __sleep and __wakeup methods to ~33 classes.
+
+#### What has to be done
+##### Admin session
+I propose creating 2 interfaces in the Magento\Backend\Spi namespace:
```
melnikovi: What would be a benefit in having these interfaces part of module API?

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
...
+* SessionUserSerializerInterface
+* SessionAclSerializerInterface
+ 
+Both interfaces will have 2 methods:
+* serialize(User|Acl $user|$acl): array - which will extract given object's
+data into an array of scalars
+* unserialize(array $data): User|Acl - taking array of scalars
+and turning it into objects
+ 
+Only arrays of scalars will be stored in session. getUser, setUser, getAcl, setAcl methods
```
melnikovi: Could you explain how these getters will be used? Do they need to be public?

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
...
+must be added to the Magento\Backend\Model\Auth\Session class to prevent serializing
+objects into admin session.
+ 
+##### CodeMess rule
+Add a PHPMD rule which will be triggered by classes
```
melnikovi: I would say this is lower priority. There tests that check code for usages of `unserialize` and `\Magento\Framework\Serialize\Serializer\Serialize`, so I think probability of adding `__wakeup` and `__sleep` is low. Do you see it happening regularly?

design-documents/Remove-__sleep-and-__wakeup.md
```diff
@@ -0,0 +1,35 @@
...
+us to add __sleep and __wakeup methods to ~33 classes.
+
+#### What has to be done
+##### Admin session
+I propose creating 2 interfaces in the Magento\Backend\Spi namespace:
+* SessionUserSerializerInterface
+* SessionAclSerializerInterface
+ 
+Both interfaces will have 2 methods:
+* serialize(User|Acl $user|$acl): array - which will extract given object's
```
melnikovi: Agree with @orlangur that these are hydrators and not serializers.

design-documents/Remove-__sleep-and-__wakeup.md
melnikovi: We should avoid storing objects in session, do you agree?

design-documents/Remove-__sleep-and-__wakeup.md
AlexMaxHorkun: No, indirect serialization cannot be covered without developers remembering to update serialization tests as well as __sleep __wakeup methods

design-documents/Remove-__sleep-and-__wakeup.md
AlexMaxHorkun: Unserialize is triggered for data stored in sessions, I've specified the case when we store objects in session

design-documents/Remove-__sleep-and-__wakeup.md
AlexMaxHorkun: There may be additional logic needed for 3rd party devs who extend User or Role classes, so they would implement their own hydrators. We can't just use getData() and setData() on those objects and expect that to be enough for everyone.

design-documents/Remove-__sleep-and-__wakeup.md
AlexMaxHorkun: Currently authenticated user and acl objects are being accessed via magic getters and setters mentioned above by client code. To trigger our new hydration instead of default PHP's unserialize we need to actually create those getters and setters

design-documents/Remove-__sleep-and-__wakeup.md
AlexMaxHorkun: if someone starts storing some new kind of objects in session they wouldn't use unserialize anywhere explicitly and would just prepare their classes for being serialized by adding __wakeup and __sleep

AlexMaxHorkun:
> > serialize(User|Acl $user|$acl): array - which will extract given object's
> > data into an array of scalars
> 
> Sounds like data hydrator rather than serializer, isn't it?
> 
> > Add a PHPMD rule which will be triggered by classes
> > having methods called __wakeup and __sleep
> 
> PHPCS is better suitable for this purpose as PHPMD rule has to look in every single token.

I don't think adding a code mess rule will produce too much performance issues

AlexMaxHorkun:
> > serialize(User|Acl $user|$acl): array - which will extract given object's
> > data into an array of scalars
> 
> Sounds like data hydrator rather than serializer, isn't it?
> 
> > Add a PHPMD rule which will be triggered by classes
> > having methods called __wakeup and __sleep
> 
> PHPCS is better suitable for this purpose as PHPMD rule has to look in every single token.

I don't think a code mess rule will impact static tests run time too much

AlexMaxHorkun:
I think a new code mess rule wouldn't impact static tests run time too much and we might need to suppress it in some cases

mslabko:
Please, take care of possible performance degradation. Use medium or bigger profile during tests.

AlexMaxHorkun:
> Please, take care of possible performance degradation. Use medium or bigger profile during tests.

sure, sounds reasonable

design-documents/Remove-__sleep-and-__wakeup.md
melnikovi: We don't have anywhere Spi as part of namespace, I propose to discuss adding this convention separately. To add more data to User and Acl, you would need to modify User and Acl and hydrators. Can we make hydrators generic? Maybe to extract data, we can have toArray method and creation of the entity will be done using constructor (haven't look at the code, don't know if it's feasible)?

PR #36, opened 2018-10-18
Sessions to be used only in controllers and blocks

AlexMaxHorkun:
It is unsafe to use sessions anywhere but controllers and blocks due to multiple web APIs in Magento

### DIFF
--- a/design-documents/sessions-only-for-html.md
+++ b/design-documents/sessions-only-for-html.md
@@ -1,0 +1,20 @@
+### Magento\Framework\Session\SessionManagerInterface should be used only in controllers and blocks
+#### Why?
+Web APIs (REST (by definition), SOAP, GraphQL) do not employ sessions,
+relying on sessions in classes that are not directly linked to preparing HTML
+may cause sessions being used during Web API requests unintentionally causing
+unexpected behaviour during and negative performance impact.
+ 
+#### Current situation
+Sessions are being used in model classes.
+ 
+#### What has to be done
+Add PHPMD rule which will be triggered when a class has a constructor parameter of type
+SessionManagerInterface (and any of the implementations) and is not an instance of
+ActionInterface or AbstractBlock.
+The rule's message should explain that UserContextInterface should be used to get current
+user instead of sessions.
+ 
+Removing all existing session usages from non-controller and non-block classes would
+prove a difficult process so we just have to warn developers not to use them
+in the future with the PHPMD rule.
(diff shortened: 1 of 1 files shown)


buskamuza:
> 6.2.2. Request, Response, Session, Store Manager and Cookie objects MUST be used only in the Presentation layer.

https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html

So, yes, this is correct and is in guidelines. And it would be great to cover it with a test.

design-documents/sessions-only-for-html.md
```diff
@@ -0,0 +1,20 @@
...
+may cause sessions being used during Web API requests unintentionally causing
+unexpected behaviour during and negative performance impact.
+ 
+#### Current situation
+Sessions are being used in model classes.
+ 
+#### What has to be done
+Add PHPMD rule which will be triggered when a class has a constructor parameter of type
+SessionManagerInterface (and any of the implementations) and is not an instance of
+ActionInterface or AbstractBlock.
```
buskamuza: 1. Also a plugin for a block/controller should be OK to use sessions.
1. Would be good to expand the rule to cover "6.2.2. Request, Response, Session, Store Manager and Cookie objects MUST be used only in the Presentation layer.". Can be implemented as separate rules.
1. Also, data providers for UI components are part of presentation layer, so they can depend on session.

AlexMaxHorkun: 1. makes sense, but I don't know how a block plugin can be identified in a PHPMD rule. Controller plugin can be sort of identified by having before/after/aroundExecute/Dispatch but I'm not sure if that would be correct. I think it would be OK if devs would just have to put SuppressWarnings for those cases
2. Ok, I'll update the proposal
3. Sure, instances of Magento\Framework\View\Element\UiComponent\DataProvider\Document and Magento\Framework\View\Element\UiComponent\DataProvider\DataProviderInterface will be ignored by the rule then

buskamuza: 1. The plugin has `$subject` which can go through the same checks on whether it's a block or controller.

Vinai:
In general I agree, but I do have to say that `\Magento\Customer\Model\Authorization\CustomerSessionUserContext` makes REST API usage on the store front very convenient.  

I think It should stick around for at least a few years still, until the GraphQL integration is mature enough and has effectively replaces the frontend REST API uses in extensions.

AlexMaxHorkun:
@Vinai CustomerSessionUserContext is an exception case when we can't identify automatically whether a session object is used properly - CustomerSessionUserContext is fine and will stay

PR #37, opened 2018-10-26
Remote Code Invoker Design Doc

AlexMaxHorkun:
Design for remote code execution as a part of distributed Magento

### DIFF
--- a/design-documents/remote-invoker.md
+++ b/design-documents/remote-invoker.md
@@ -308,6 +308,13 @@ the results ASAP
 * currently it only supports string argument and will need to be updated heavily anyway - no time saved
 comparing to writing a proper HTTP gateway based RPC designed for internal services communications
 
+#### Invoker based on RESTful web API
+To differentiate outer web APIs and inner web APIs their configuration must differ, there are 2 ways to do it:
+  * add _inner_ attribute to _route_ (webapi.xsd) element which when _true_ will mean that inner call mode would be
+  enabled for Magento when the API is used (requests would have to be signed, call ID provided etc.) and
+  some additional sub-elements/attributes maybe used/omitted (like not having to use _url_ attribute for such APIs)
+  * add another .xsd and new config files to modules to describe inner web APIs
+
 ### Explanations
 #### Need for an abstraction around network calls
 Say we're writing a Proxy version of a ProductRepositoryInterface service using HTTP client and RESTful API.
--- a/design-documents/remote-invoker.md
+++ b/design-documents/remote-invoker.md
@@ -4,22 +4,87 @@ Remotely calling services in proxy versions of modules
 ### Overview
 This documents describes remote invoking mechanism, it's functional requirements, the module's API and SPI.
 It also describes a recommended RPC-style networking gateway to call services remotely, but there can be
-multiple _Remote Code Invoker_ implementations based on existing Web APIs, Message Queue API or any other.
+multiple _Remote Code Invoker_ implementations based on existing Web APIs, Message Queue RPC API or any other.
+### How are remote services are going to look like with Remote Code Invoker
+\<Module name\>Proxy module will contain \<Module name\>API module's implementations that would look like this:
+```
+class ServiceProxy implements \Magento\ModuleNameApi\ServiceInterface
+{
+
+....
+
+    public function doStuff(\Magento\ModuleNameApi\Data\DTOInterface $dto): \Magento\ModuleNameApi\Data\ResultDTOInterface
+    {
+        $remoteResult = null;
+        $promise = $this->invoker->invoke(
+            'catalog',
+            \Magento\ModuleNameApi\ServiceInterface::class,
+            'doStuff',
+            [$dto]
+        );
+        $promise->then(function (\Magento\ModuleNameApi\Data\ResultDTOInterface $result) use (&$remoteResult) {
+            $remoteResult = $result;
+        });
+        $promise->wait();
+        
+        return $remoteResult;
+    }
+
+....
+
+}
+``` 
+ 
+di.xml in \<Module name\>Proxy module will specify that ServiceProxy
+is the preference for \Magento\ModuleNameApi\ServiceInterface.
+
+So when a local service uses \<Module name\>API's service it will not be aware whether the service is actually
+remote and it won't affect how we write code.
+Example:
+```
+class LocalServiceA implements ServiceAInterface
+{
+
+    /**
+     * @var \Magento\BModuleApi\ServiceBInterface
+     */
+    private $serviceB;
+
+....
+
+    public function doAStuff(): void
+    {
+        ....
+
+        $stuff = $this->serviceB->doBStuff($someVar);
+        
+        ....
+    }
+
+....
+
+}
+```
+
+Depending on whether _BModule_ or _BModuleProxy_ is deployed on the local node _ServiceBInterface_ will either
+work locally or remotely.
 ### Why not just use cURL and existing Web APIs
 We need a mechanism, an abstraction around an actual network interactions that would provide a retries mechanism,
 a standard to pass local call's arguments, a request signing mechanism, authentication mechanism which would allow
-to perform authentication only once on the node that accepts user requests, and not on each node etc
+to perform authentication only once on the node that accepts user requests and not on each node, data consistency
+mechanism etc.
 _([see explanation](#need-for-an-abstraction-around-network-calls))_.
 ### Functional requirements
 * Invoker's code is a part of Magento\Framework namespace
 * Invoker can call classes and method that were specifically enabled for remote calls via a config
+* Invoker provides a way for services to written as usual, not knowing whether other modules' services are remote or not
 * Multiple remote sources can be configured and assigned unique IDs
 * Each HTTP request to remote sources is signed using unique source's key, timestamp and request's attributes
 * Each remote call must be given unique ID to be used by retries mechanism
 * Timeout limit and retries count can be configured globally and for each call
 * Serialized arguments string includes arguments instances' classes information for proper
 deserialization _([see explanation](#inability-to-just-use-interfaces-for-arguments-and-return-types))_
-(serializer employed for RESTful gateway my be reused)
+(serializer employed for RESTful gateway will probably be reused)
 * Serialized remote method return values string must contain type information for
 proper deserialization by the caller
 * Remote call response can provide either serialized method call result or serialized exception to be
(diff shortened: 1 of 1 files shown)


design-documents/remote-invoker.md
```diff
@@ -0,0 +1,187 @@
...
+performed every time a proxy service calls remote service unlike when a service
+is called locally
+* Proxy services would have to generate auth information of already authorized users (tokens etc) to access REST/SOAP
+services which is a redundant process, also some of those REST/SOAP endpoints would have limitations in
+place for certain type of users which would be useless in case of remote execution
+(like requiring an admin user)
+* Other application state spoofing may be required for seamless remote execution of services
+### Functional requirements
+* Invoker's code is a part of Magento\Framework namespace
+* Invoker is able to call any class' method
```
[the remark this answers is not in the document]
AlexMaxHorkun: @antonkril Updated the doc

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,187 @@
...
+* Multiple remote sources can be configured and assigned unique IDs
+* Each HTTP request to remote sources is signed using unique source's key, timestamp and request's attributes
+* Each remote call must be given unique ID to be used by retries mechanism
+* Timeout limit and retries count can be configured globally and for each call
+* Serialized arguments string includes arguments instances' classes information for proper deserialization
+* Serialized remote method return values string must contain type information for
+proper deserialization by the caller
+* Remote call response can provide either serialized method call result or serialized exception to be
+rethrow in order to emulate local method call
+* Remote calls are done in a synchronous manner because in order for
```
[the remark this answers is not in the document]
AlexMaxHorkun: @antonkril I've added explanations for this issue

AlexMaxHorkun:
I have added "Explanations" section to the document where some of the controversial points are being explained

design-documents/remote-invoker.md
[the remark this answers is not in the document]
antonkril: I understood your explanation, but the conversation is about making new services asynchronous, not existing ones.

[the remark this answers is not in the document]
AlexMaxHorkun: You mean you want our new service contracts look like below and client code always using other services in asynchronous manner?
```
interface NewEntityRepositoryInterface
{
    public function save(NewEntityInterface $entity): Promise;
}
```

Local implementations would almost always be executed in a synchronous manner and using them would just severely complicate our code. PHP does not support async code execution in it's core, and making some of the code async will most likely force all the client code working with such services be async. I mean look at node.js and it's callback hell, do we want that in Magento?

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,618 @@
...
+    }
+ 
+....
+ 
+}
+```
+If the remote call to _ProductRepositoryInterface_ service was done via RESTful (or GraphQL) web API the _incrOrdederedCount_ would have
+to be exposed as Web API and could be requested via HTTP by anyone and independently from context which we cannot afford since we can't just increase the counter without an order happening so only another service responsible for orders may call _incrOrdederedCount_.
+ 
+If done via the Invoker the _incrOrdederedCount_ wouldn't be exposed to be requested by HTTP clients via web API and wouldn't show up on
```
buskamuza: There is no difference because Invoker makes the method exposed to all services, if I understand the proposal correctly.
Keep in mind that client-facing Web APIs are not the same as service-facing (inter-network) Web APIs. Only limited list of Web APIs (explicitly declared) will be exposed to the client on the BFF side. All other APIs will be available for services only.

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,618 @@
...
+    }
+ 
+....
+ 
+}
+```
+If the remote call to _ProductRepositoryInterface::decrOrderedCount_ was done via RESTful web API the exposed endpoint would have
+ACL configured for it so it requires _Catalog edit access_ from admins, but our current admin only has access to _orders_ and the call would fail. However if the Catalog domain modules were not remote and the _ProductRepositoryInterface::decrOrderedCount_ call was made then no ACL validation would occur and we would save the order successsfully.
+ 
+With Invoker such ACL validations wouldn't occur and the remote call would be completed just as if it was local.
```
buskamuza: ACL for Web API should be revisited. It's not necessary to introduce another type of APIs (invoker-based) to modify access.
Authentication and authorization can be done on API Gateway level, and in-network calls between services can be done w/o additional validation. Or for better security, there can be additional validation on each service, which will be probably similar to what you propose for the invoker. It depends on how we want to implement it.

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,618 @@
...
+* Proxy services would have to generate auth information of already authorized users (tokens etc) to access REST/SOAP
+services which is a redundant process, also some of those REST/SOAP endpoints would have limitations in
+place for certain type of users which would be useless in case of remote execution
+(like requiring an admin user) _([see explanation](#redundant-authorization-validation-when-using-existing-web-api))_
+* Other application state spoofing may be required for seamless remote execution of services
+* RESTful Web API provides access to service contracts by assigning a unique human-readable
+HTTP method/URL pair for external systems - we don't need that when writing a proxy service since we
+would always already have class and method's names, the only other thing existing RESTful Web API mechanism
+provides is it's serializer and it will be reused in the _Remote Code Invoker_
+* When an exception occurs during a service call via RESTful API the response does not contain
```
buskamuza: Looks more like an issue with error reporting for Web APIs.

design-documents/remote-invoker.md
AlexMaxHorkun: With RPC style communication there would be no need to re-declare our web API endpoints and there would be a clear differentiation between web API endpoints for outside world and inner-network

design-documents/remote-invoker.md
AlexMaxHorkun: not really, clientside and 3rd-party systems wouldn't care for exception classes in responses of RESTful APIs

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,631 @@
...
+    }
+
+....
+
+}
+```
+
+Depending on whether _BModule_ or _BModuleProxy_ is deployed on the local node _ServiceBInterface_ will either
+work locally or remotely.
+### Why not just use cURL and existing Web APIs
```
melnikovi: I was thinking that service mesh will handle retries

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,631 @@
...
+mechanism etc.
+_([see explanation](#need-for-an-abstraction-around-network-calls))_.
+### Functional requirements
+* Invoker's code is a part of Magento\Framework namespace
+* Invoker can call classes and method that were specifically enabled for remote calls via a config
+* Invoker provides a way for services to written as usual, not knowing whether other modules' services are remote or not
+* Multiple remote sources can be configured and assigned unique IDs
+* Each HTTP request to remote sources is signed using unique source's key, timestamp and request's attributes
+* Each remote call must be given unique ID to be used by retries mechanism
+* Timeout limit and retries count can be configured globally and for each call
```
melnikovi: I think service mesh should handle these things.

AlexMaxHorkun: service mash is a separate system which is acting as a proxy between inter-service communications and, yes, provides retries and data consistency. I haven't found a mention of such system in Anton's docs and nor do we need such system - they only make sense when different services use different languages/frameworks etc. And we are going to use PHP + Magento on all of our services.

AlexMaxHorkun: If what you were meaning to say is that we need a separate entity in our codebase to handle retries, data consistency etc - this document is in fact describing such entity (invoker = service mesh)

melnikovi: Service mesh can solve many problems related to deployment and communication for us: service registration, routing, load balancing, circuit breaking, retries, statistics and monitoring, etc. Retries logic is complicated and I would suggest not to implement it ourselves.

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,631 @@
...
+        $promise = $this->invoker->invoke(
+            'catalog',
+            \Magento\ModuleNameApi\ServiceInterface::class,
+            'doStuff',
+            [$dto]
+        );
+        $promise->then(function (\Magento\ModuleNameApi\Data\ResultDTOInterface $result) use (&$remoteResult) {
+            $remoteResult = $result;
+        });
+        $promise->wait();
```
melnikovi: I wonder if we need to have different interfaces for synchronous and asynchronous communication. Developer would know how communication happens. And we will not have to wait for result of asynchronous operation, one of the benefits of async operations is that we don't need to wait for the response and it can be performed quickly.

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,631 @@
...
+* Serialized remote method return values string must contain type information for
+proper deserialization by the caller
+* Remote call response can provide either serialized method call result or serialized exception to be
+rethrow in order to emulate local method call
+* Remote calls can be done in a synchronous manner to ease remote invoking for
+existing services _([see explanation](#using-synchronous-remote-invoking-in-existing-services))_
+and in an asynchronous manner for new services to be able to interact with other services with remote invoking in mind
+* Each Magento instance that can be used for remote calls (_Magento server_)
+has it's own unique key for signing requests
+* Remote calls can be disabled in a Magento server
```
melnikovi: What is Magento server, instance/installation?

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,631 @@
...
+* Authenticated user's information (ID and type) is passed from caller
```
melnikovi: Will it be added by the remote implementation of the service?

design-documents/remote-invoker.md
```diff
@@ -0,0 +1,631 @@
...
+
+#### RPC-style gateway
+
+HTTP protocol will be used. A controller will be added to process a single URL like _magento-base-url_/invoker.
+Requests' body will contain requested class name, method name, arguments and it all encoded in json.
+Requests' headers will contain information such as authenticated user ID, call ID, issued timestamp, signature.
+Response body will contain serialized service call result or a serialized thrown exception encoded in json.
+
+#### Why RESTful web API should not be used as the network gateway for the invoker
+* Not all service contracts are exposed as web API and nor should they be (for security and workflow reasons)
```
melnikovi: We will have BFF that would expose public API. API of the services can be exposed and will be protected by private network.

buskamuza:
Requires tech vision for inter services communication first.

melnikovi:
Closing as topic is covered in https://github.com/magento/architecture/pull/50.

PR #39, opened 2018-11-02
dependency generation tool

adeptofvoltron:


buskamuza:
@adeptofvoltron , it's a great proposal and is something discussed here an there. But I think it's more a question to the product team of whether this can fit to the backlog.
So I'm going to pass this ticket to our Product team, and close it.
Also, a similar tools is mentioned in the [Services Isolation](https://github.com/magento/architecture/blob/master/design-documents/service-isolation.md#instance-builder) technical vision.

PR #42, opened 2018-11-09
Standard Promises

AlexMaxHorkun:
Standard promises to use across Magento

### DIFF
--- a/design-documents/promises.md
+++ b/design-documents/promises.md
@@ -23,8 +23,14 @@ This interface will be used as the return type of methods returning promises.
 ```php
 interface PromiseInterface
 {
+    /**
+     * @throws PromiseProcessedException When callback was alredy provided.
+     */
     public function then(callable $callback): void;
     
+    /**
+     * @throws PromiseProcessedException When callback was alredy provided.
+     */
     public function otherwise(callable $callback): void;
     
     public function wait(): void;
--- a/design-documents/promises.md
+++ b/design-documents/promises.md
@@ -10,7 +10,7 @@ operation, it will receive a single argument - a _Throwable_
 * Promises can be used in a synchronous way to prevent methods that use methods returning promises
 having to return a promise as well ([see explanation](#callback-hell)); This will be done by promises having _wait_ method
 * _wait_ method does not throw an exception if the promise is rejected
-nor does it return the result of the resolved promise
+nor does it return the result of the resolved promise ([see explanation](#wait-not-unwrapping-promises))
 * If an exception occurs during an asynchronous operation and no _otherwise_ callback is
 provided then it will just be rethrown
 ### API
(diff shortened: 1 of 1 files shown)


antonkril:
Please add comparison of existing libraries.

design-documents/promises.md
```diff
@@ -0,0 +1,112 @@
+### Why?
```
melnikovi: After service separation we will have a lot of communication between services. Separating checkout adds > 20 additional requests. These brings two issues: increases loads (number of requests) and decreases performance (more communication). We will have to reduce amount of communication and optimize services themselves to minimize/avoid these issues.

Invoker in this proposal, from what I understand, will generate even more requests (`wait` method) which are not cheap. It will also increase execution time by much because we can send requests to get result of operation only with certain interval, which in our case can not be small. If we have composite operation, total time will be even bigger. The interval will be very hard to get right, as it should depend on load and average response time.

Because of the concerns above, I would like to understand better where where will we use this API and make sure it's good fit for these use cases.

design-documents/promises.md
```diff
@@ -0,0 +1,112 @@
...
+and it's time to introduce a standard Promise to Magento for asynchronous operations to employ
+### Requirements
+* Promises CANNOT be forwarded with _then_ and _otherwise_ ([see explanation](#forwarding))
+* _then_ accepts a function that will be executed when the promise is resolved, the callback will
+receive a single argument - result of the execution
+* _otherwise_ accepts a function that will be executed if an error occurs during the asynchronous
+operation, it will receive a single argument - a _Throwable_
+* Promises can be used in a synchronous way to prevent methods that use methods returning promises
+having to return a promise as well ([see explanation](#callback-hell)); This will be done by promises having _wait_ method
+* _wait_ method does not throw an exception if the promise is rejected
```
melnikovi: Could you explain what `wait` method does?

design-documents/promises.md
```diff
@@ -0,0 +1,112 @@
+### Why?
+Service contracts in the future will often be executed in an asynchronous manner
+and it's time to introduce a standard Promise to Magento for asynchronous operations to employ
+### Requirements
+* Promises CANNOT be forwarded with _then_ and _otherwise_ ([see explanation](#forwarding))
+* _then_ accepts a function that will be executed when the promise is resolved, the callback will
```
melnikovi: When promise is resolved? When we received response on our request or in case of async operation we retried result of our operation? How do we retrieve result of our operation? Can we have API that would allow to create multiple promises and execute code only when all of them resolved?

design-documents/promises.md
AlexMaxHorkun: waits when promise is resolved (resolve method called by initiator). Please see guzzle promises repo for examples

design-documents/promises.md
AlexMaxHorkun: you will be retreiving promised results as the 1st argument to _then_ callback. Creating an util for waiting for multiple promises doesn't seem necessary - you can just use _wait_ on promises you want to wait for

kirmorozov:
Fellows, please come back PHP, 
PHP has no shared memory or loops between threads.
PHP has no execution time frame after delivery of request.
PHP has no execution time before handling request.
There is no event loop. 

Please explain use-case,
and why it cannot be handled with simple while loop where needed.

kirmorozov:
@AlexMaxHorkun What problem are you solving?
Why do you need any promises? 
Do you have any runtime? 
What are you waiting for during short lifetime of php? (average response time must stay under 300ms)

PR #48, opened 2018-11-15
Services authentication & authorization

joni-jones:


design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+
+And the basic scenario might be as following:
+
+![Auth Basic Scenario](auth-basic-scenario.png)
+
+Also, there are a few variants of the proposed schema like to have separate services for token validation and renewal:
+
+![Token Validation Service](token-validation-service.png)
+
+Now, each service validates a token through `Token Service` it reduces logic duplication but adds additional requests
```
buskamuza: > it reduces logic duplication

Logic duplication can be reduced by creating a shared library that will be installed on all services. This will reduce number of network calls, and also will allow to rollout updates of the library gradually, if necessary (though this shouldn't be a frequent task).
Having a single service, it becomes a bottleneck, so if it goes down (or is in upgrade process), all services will be blocked.

buskamuza:
Could you add more to the permissions (authorization) topic? Do we work with scenario-based permissions (e.g., this customer can perform Add to Cart scenario) or with action-based ones (e.g., this customer can request QTY for a product in this store)? I assume we're talking about action-based permissions, as we're going to validate the token on each service. But wanted to clarify. And if we indeed talk about action-based permissions, how to avoid situation when a customer places an order and suddenly (after placing the payment), the system realizes that it can't reduce QTY for the purchased products because this customer doesn't have permissions for this?

design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+which might impact on performance.
+
+According to the proposed schema, user-Monolith/BFF scenario could use JWE tokens, Monolith/BFF - service could use JWS
+tokens to make the communication more secure.
+
+### OAuth 2.0
+
+As `Auth Service` is the single entry for authentication & authorization it should store roles and privileges for all
```
melnikovi: I propose not to validate privileges when we have service to service communication. All operations will be defined on the BFF/monolith and these are operations we need to control access to. To perform these operations monolith may talk to one or more services which in turn can talk to other services. For these communication I propose not to check privileges. It's more work and we will be granting privileges to operations defined in BFF/monolith. We can do authentication if we want to make communication within the network more secure. Some people don't do it.

AlexMaxHorkun:
There is no point for services to pass a token around - services cannot be accessed directly by clients and received requests can be considered as geniune. Services would just pass enought data to initialize a UserContextInterface (user type and id). If we allow scenario where nodes are not within a closed network we would have to sign whole requests - not just auth data (JWT) because other sensitive data maybe passed around as well (User roles, area code etc.). It would be an overhead to keep auth services outside front nodes (BFFs for admins and customers) because for most client requests that service would be accessed remotely so it's best to keep auth services local for front nodes to avoid sending remote service calls to auth service each time.

AlexMaxHorkun:
if you want to change tokens we use right now for RESTful web API to JWTs that would not be profitable. Encoding user info in tokens is useful to avoid making heavy operations (e.g. DB queries) for every request, but since tokens can be manually revoked (for customers, admins and 3rd party integrations) we would have to check the customer/user tables everytime anyway to determine whether provided tokens were revoked

joni-jones:
> There is no point for services to pass a token around - services cannot be accessed directly by clients and received requests can be considered as geniune.

This document doesn't contain any mentions that services are available from the outside network, vice versa, the schema shows all services under the protected network.

> we would have to check the customer/user tables everytime anyway to determine whether provided tokens were revoked

You don't need to do it, it's enough to implement the mechanism for key publication between services, and after revoking a token, just publish a new key. Each service will decide if it's enough to validate signature or additional validation (like permissions checking).

design-documents/service-isolation/authentication-authorization.md
joni-jones: As at least now, all our services will be on the same stack - a separate validation service doesn't make sense. But potentially, services might be implemented using a different stack, that's why I've described the additional approach. For now shared library it's ok.

design-documents/service-isolation/authentication-authorization.md
joni-jones: How we can guarantee that service could execute a request from another service?

AlexMaxHorkun:
That is my points - there is no reason pass around a token which would require decryption and validation. Services can just pass user type and id

AlexMaxHorkun:
Why would we have a keyat all when services contact other services?

joni-jones:
> That is my point - there is no reason pass around a token which would require decryption and validation.

A validation allows checking if a token wasn't modified. Services won't do decryption, it's enough to validate the signature (the JWT library does all validation). 

> Services can just pass user type and id

Not all services should have user type and user id to execute a request, for example, Payment Service to execute authorization transaction doesn't require the user context but somehow the service should no if the request initiator has permissions to do it.

> Why would we have a key at all when services contact other services?

The main idea of using a token for each request is avoiding [Confused Deputy Problem](https://en.wikipedia.org/wiki/Confused_deputy_problem) and pass all additional data, like permissions via request in the header as a simple string. To verify the token different strategies might be used:
- HMAC - uses a private key for token issuing and validation
- RSA - uses a private key for token issuing and a public key for a token validation

AlexMaxHorkun:
> A validation allows checking if a token wasn't modified. Services won't do decryption, it's enough to validate the signature (the JWT library does all validation).

Modified by who? Data is being exchanged within closed network by services - there's nobody to falsify the data.



> Not all services should have user type and user id to execute a request, for example, Payment Service to execute authorization transaction doesn't require the user context but somehow the service should no if the request initiator has permissions to do it.

Usually we check permissions for the whole endpoints (like POST /V1/products requires Magento_Catalog::products) and do not check permissions for each service contract call required to process client requests underneath. In cases when we have to check ACL resources access for some sub-operation it makes sense to pass ACL role information between services but, again, we don't need tokens for that - there's nobody to falsify requests and we can just add additional information as is to requests between services.

design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,146 @@
...
+As each service should be agnostic to a source of a request (in general, for a service it does not matter who sends
+a request) service-to-service communication also requires a token. The next diagram describes a possible flow:
+
+![Service to Service](service-to-service.png)
+
+Before making a call to the needed service, the first service makes a call to `Auth Service` to retrieve token for the future
+communication. The second service makes a call for token validation in the same way as for other types of requests, so
+for the second service, it does not matter who sends original request.
+
+### Guest-to-Service Communication
```
melnikovi: We also going to have service to monolith communication at the beginning. How would we distinguish these requests and enable our authorization mechanism for them?

design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,146 @@
...
+
+ - Anyway, each service should check if a user is allowed to perform an action (for example, if the current user is allowed
+ editing profile details because only service knows the profile owner)
+ - Search WEB API operations like `getList` should be filtered by a request initiator (a customer can retrieve only his
+ orders, a guest does not have customer ID, admin can get all orders, a company user - according to the allowed hierarchy)
+ - Roles with hierarchy probably might use [Open Policy Agent](https://www.openpolicyagent.org/)
+
+## Implementation Approach
+
+On the first iteration, BFF/Monolith could be used as `Auth Service` and as a single point for roles & privileges storage.
```
melnikovi: How authorization will be enabled on each service and monolith? Do we enable authorization for all requests or only for post operations and ones that contain sensitive data (I think this would be better)?

AlexMaxHorkun:
About services checking ACL privileges: if auth service will be local for BFF there's no point of a service to send a request back to it when checking for authorization - it will be more efficient for BFF to check access to resources when accepting a request considering such check will be performed for nearly most of the endpoints. Also currently we state ACL resource for an endpoint in webapi.xml - those files will reside on BFF, for a service to know an endpoints assigned resource it would either have to have *Webapi modules installed as well, or a separate xml files just with resources assigned. That doesn't seem great

AlexMaxHorkun:
Also, forwarding allowed ACL resources between services may not be possible since authorization is not always so straightforward - we have different rights for different company users for instance, as in company admin may edit only customers from their own company. That's just 1 example. I think auth service (which will be a part of BFF) should be solely responsible for performing authorization based on required resource access defined for endpoints in webapi.xml's present on BFF

design-documents/service-isolation/authentication-authorization.md
joni-jones: I think for GET requests authorization also should be enabled because service should know if request initiator is allowed to request such data.

joni-jones:
> Modified by who? Data is being exchanged within closed network by services - there's nobody to falsify the data.

A service should not rely on infrastructure, for a separate service it doesn't matter who is a request initiator.

> Usually we check permissions for the whole endpoints (like POST /V1/products requires Magento_Catalog::products) and do not check permissions for each service contract call required to process client requests underneath.

That's why the current permission infrastructure doesn't have granularity and you can't recognize if a user has permissions to get only entities which available for him because it has access for a whole resource.

> In cases when we have to check ACL resources access for some sub-operation it makes sense to pass ACL role information between services

In that case, all services should know about all roles and managing dynamic roles will be a challenge.

> we have different rights for different company users for instance, as in company admin may edit only customers from their own company.

That's why ACL is not the best way to do authorization, the ACL doesn't allow to recognize such granular operations.

design-documents/service-isolation/authentication-authorization.md
joni-jones: What kind of operation do you mean? When service should be an initiator for such request? In general, it's the same type of communication as service-to-service.

AlexMaxHorkun:
So do you propose also to change our whole authorization mechanism?

joni-jones:
> So do you propose also to change our whole authorization mechanism?

Yes, exactly because ACL doesn't allow achieving needed requirements.

AlexMaxHorkun:
what requirements is that?

AlexMaxHorkun:
> When admin cancels an order, the `order/cancel/request` will be initiated according to the admin role, the request initiator doesn't know anything about repositories or other application layers it just calls an accessible WEB API endpoint. The validation will check if a current role with current user (admin, customer, etc) is allowed to perform such action.

That is my point - we need to check only "order/cancel/request" rights, not internal OrderRepository::save() rights. OrderRepository::save() is a service, which when exposed as public API would require order::edit permissions, but since it's initiated by another service we don't need to check those right because it's an indirect call

AlexMaxHorkun:
> This is a good point. At the same time, a 3rd party system can interact with inventory and there are should be permissions for that system. So, anyway, the service should check if the request can be processed. For the described case, API Gateway/BFF might have own permissions based on a customer request permissions to make internal calls.

user is not responsible to how his requests are actually processed. So, in the situation I've described you assume that customers will be given some inventory::manage rights? they really shouldn't

AlexMaxHorkun:
As a part of Magento security team I can tell you that you're fighting windmills and those security holes, if they are still present, left in Magento regarding accessing specific entities that don't belong to users are solved in a different way

joni-jones:
> user is not responsible to how his requests are actually processed. So, in the situation I've described you assume that customers will be given some inventory::manage rights? they really shouldn't

No, I did not mention that customers will be given rights to the inventory services.

> As a part of Magento security team I can tell you that you're fighting windmills and those security holes, if they are still present, left in Magento regarding accessing specific entities that don't belong to users are solved in a different way

Actually, from time to time we get requests from customers related to such issues. Now, each service layer should solve this issue again and again. There is no standard mechanism to resolve it.

design-documents/service-isolation/authentication-authorization.md
melnikovi: I'm not sure we need to perform authorization/authentication for all get requests. Number of calls and ability to efficiently cache requests is major concern right now. Caching in the application will have much lover performance and higher overhead than caching by Varnish. I think we don't need to perform authorization for GET requests that don't expose sensitive data.

joni-jones: I partially agree but, for example, order items details are not sensitive data but probably we won't provide an ability to expose ordered items from one customer to another.

design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,146 @@
...
+![Auth Basic Scenario](auth-basic-scenario.png)
+
+A token validation per each request on `Auth Service` allows keeping authorization process in the centralized place, only
+`Auth Service` knows about all users, roles and assigned privileges. Each service during deployment should expose their API
+to `Auth Service`, which can automatically assign API to existing roles based on default rules. The token validation includes
+not only signature verification but authorization.
+
+### Services Deployment
+
+`Auth Service` stores allowed operations for roles and, during deployment, each service should send a list of available
```
melnikovi: We could have simple (operation performed by one service) and composite operations (operation performed by multiple services). Composite operation can be initiated by the service (for instance Checkout service calls Totals service) or BFF. How are we going to give permissions to these operations? Seems like we need to have some sort of configuration.

design-documents/service-isolation/authentication-authorization.md
melnikovi: I would consider order items sensitive data as it's part of order data. Not sure we should have API to get only order items data. We can cache for instance product, reviews and tax data.

design-documents/service-isolation/authentication-authorization.md
joni-jones: Probably we need to define the list of "standard" operations for a separate service and define the privileges for them. Other permissions can be added manually.
As service owners, we know about all flows (which includes requests to other services) and can define all operations pretty accurate.

design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,146 @@
...
+
+![Authentication and Authorization General Schema](auth-general-schema.png)
+
+1. An unauthorized user sends a request with credentials to retrieve an access token, BFF/Monolith proxies this request to
+`Auth Service` which performs authorization & authentication and issues access token based on a user role.
+2. The user sends the request to get a needed resource with the access token, BFF/Monolith proxies this request to the appropriate
+service and there are few multiple strategies for token validation:
+    - BFF/Monolith just verifies the token signature (each service sends a token for validation to `Auth Service`)
+    - BFF/Monolith sends the token to `Auth Service` for validation
+3. A service sends the token for validation to `Auth Service` and, if a request is allowed, the service processes the request.
```
melnikovi: Add to cart operation may involve communication between more than 5 different services. Having each of these services send request to Auth service to validate the token is highly undesirable. If we implement Auth service using existing technologies, this would add a lot of delay.

design-documents/service-isolation/authentication-authorization.md
```diff
@@ -0,0 +1,146 @@
...
+
+As Magento should provide [multiple BFF](https://github.com/magento/architecture/blob/master/design-documents/service-isolation.md#backends-for-frontends)
+it makes sense to extract authentication and authorization flow to a separate `Auth Service` to avoid credentials
+validation and duplication of roles & privileges data. And the general schema might look like this:
+
+![Authentication and Authorization General Schema](auth-general-schema.png)
+
+1. An unauthorized user sends a request with credentials to retrieve an access token, BFF/Monolith proxies this request to
+`Auth Service` which performs authorization & authentication and issues access token based on a user role.
+2. The user sends the request to get a needed resource with the access token, BFF/Monolith proxies this request to the appropriate
```
melnikovi: Can BFF handle tokens? This would allow to save 1 network call and will not require clients know about tokens. To receive product information, for example, client would send request to BFF. BFF will request the token and send the request to Catalog service.

design-documents/service-isolation/authentication-authorization.md
melnikovi: For instance, Totals service after first iteration can still be part of monolith application. To add product to cart, Checkout service will need to send request to monolith application.

After more reading I noticed that you described this in the proposal, thank you.

design-documents/service-isolation/authentication-authorization.md
melnikovi: It would be good to define cases when do we need to check if service can call other service to perform operation. How it's different from monolith deployment?

Some people don't perform this validation. Communication between services is trusted and secured from public by private network.

design-documents/service-isolation/authentication-authorization.md
joni-jones: Actually, now we have WEB API endpoint to get order items. In general, order item just contains product details (which are not sensitive). Mentioned example just to show that not always we can strictly divide some data to sensitive and not.

design-documents/service-isolation/authentication-authorization.md
joni-jones: > Some people don't perform this validation. Communication between services is trusted and secured from public by private network.

That was our first discussion, probably we need to discuss such approach again about private/public services communication.

design-documents/service-isolation/authentication-authorization.md
joni-jones: Probably we need to discuss requests segregation to private and public again.

melnikovi:

PR #49, opened 2018-11-30
Add front-end tech vision doc

jedmao:
The JavaScript ecosystem can be daunting when presented with so many options. Which libraries do you choose and why this one over that one?

<a href="https://hackernoon.com/how-it-feels-to-learn-javascript-in-2016-d3a717dd577f">so many JavaScript options</a>

This document is intended to set you off on the right foot. Think of it as a generic guide for new front-end applications. Your specific project will most likely have needs of its own that deviate from this guide. That's totally fine and expected.

### DIFF
--- a/design-documents/frontend/technical-vision.md
+++ b/design-documents/frontend/technical-vision.md
@@ -25,10 +25,10 @@ Creating a new React project has never been easier with [CRA][], which comes out
 ### Technologies
 - Create a new React project with [CRA][].
 - [Redux](https://redux.js.org/) as a predictable state container.
-   - Handling of data fetching and side effects now have first class support with [React Suspense, React.lazy](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactlazy-code-splitting-with-suspense) and [hooks (proposal)](https://reactjs.org/docs/hooks-intro.html).
+   - Handling of data fetching and side effects now have first class support with [React Suspense, React.lazy](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactlazy-code-splitting-with-suspense) and [hooks](https://reactjs.org/docs/hooks-intro.html).
    - Logging with [redux-logger](https://github.com/evgenyrodionov/redux-logger).
 - **Internationalization** with [react-intl](https://www.npmjs.com/package/react-intl).
-   - _Disclaimer: the documented path here is to couple the components to this library. This will be fine for most projects, but if ultimate **portability** is required, you may want to decouple with the [container paradigm][]._
+   - _Warning: the documented path here is to couple the components to this library. This will be fine for most projects, but if ultimate **portability** is required, you will need to decouple the component from the [i18n](https://en.wikipedia.org/wiki/Internationalization_and_localization) library._
 
 ## [TypeScript][]
 
@@ -46,8 +46,11 @@ _Disclaimer: though possible to introduce [TypeScript][] into an existing [CRA][
 
 ### Technologies
 - [Adding TypeScript to CRA](https://facebook.github.io/create-react-app/docs/adding-typescript).
-- Code **consistency** is enforced via [configuration](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html) and [linting with TSLint](https://palantir.github.io/tslint/).
-   - _Disclaimer: use the most strict rules before you write a single line of code, because it's much harder to introduce these rules later._
+- Code **consistency** is enforced via [configuration](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html), [linting with TSLint](https://palantir.github.io/tslint/) and formatting with [Prettier](https://prettier.io/).
+   - Use the most strict tsconfig rules before you write a single line of code, because it's much harder to introduce these rules later in a project.
+   - Use [`tslint-config-prettier`](https://www.npmjs.com/package/tslint-config-prettier) to ensure there is no overlap between TSLint and Prettier.
+   - Extend `tslint-recommended` as a starting point. Introduce new rules with extreme hesitation to prevent bikeshedding over trivial and subjective preferences.
+   - With [`husky`](https://www.npmjs.com/package/husky), ensure errors are caught before CI with a [`pretty-quick`](https://www.npmjs.com/package/pretty-quick) pre-commit hook and a pre-push hook for linting.
 - Generate [TypeScript][] declarations for [CSS Modules](https://github.com/css-modules/css-modules) via [css-modules-typescript-loader](https://www.npmjs.com/package/css-modules-typescript-loader).
 - Documentation via [TSDoc](https://github.com/Microsoft/tsdoc#tsdoc).
 
@@ -65,11 +68,9 @@ On top of sharing code between different environments, we should also be concern
 For **performance** reasons, JavaScript should only be used for cheap operations on the front end, reserving expensive operations for a services layer. It's worth calling out that we could achieve even more code sharing if this API were written in JavaScript. For practical reasons, however, it makes more sense to leverage internal PHP resources for this task.
 
 ## Browser Support
-All users should have a **consistent** user experience. For the best **performance**, those with the latest browsers should not be penalized by incurring an additional payload hit than those with older browsers. To achieve this goal, we download just the [polyfills](https://remysharp.com/2010/10/08/what-is-a-polyfill) needed for our sites, tailored to each browser via [Polyfill.io](https://polyfill.io) feature detection.
+All users should have a **consistent** user experience. For the best **performance**, those with the latest browsers should not be penalized by incurring an additional payload hit than those with older browsers.
 
-> We test browser support with a test suite for each feature. A browser is considered compliant only if it passes all the tests, so this is a reasonably comprehensive test rather than a naïve feature-detect.
-
-_TODO: Investigate building two separate [TypeScript][] builds, one that compiles down to ES5 and another lighter-weight modern build, specifically for modern browsers. Depending on how well Polyfill.io works, the modern build might be all we need._
+There are [various ways](https://polyfill.io) to achieve this goal; however, an in-house approach might be warranted to prevent the stability issues of relying on 3rd party resources. One idea would be to publish multiple build targets, tailored to various environments.
 
 ## React UI Component Library
 
@@ -80,11 +81,13 @@ We should lean against downloading various UI components from different sources 
 - **Inconsistent** implementations, documentation, theming and customization.
 - Limited **accessibility** and support.
 
-[KendoReact | UI for React Developers](https://www.telerik.com/kendo-react-ui/) is a great option for those wishing to get React UI components from the same source. Still, there will be times when you need to download a 3rd-party package.
+If you have the budget to pay for [a product](https://www.telerik.com/kendo-react-ui/), it might be a good choice in this space. Why? Because it's much better to get all of your components from a single source than to have completely different paradigms scattered everywhere.
+
+_Disclaimer: there may be licensing limitations that prevent an open-source product from implementing a paid-for product._
 
 ### Downloading a 3rd-party package
 
-You may have to download a 3rd party component. When this happens, consider the following criteria before making a final choice:
+There are times when you may consider downloading a 3rd party package (e.g., a component). When this happens, consider the following criteria before making a final choice:
 
 1. First thing's first – what is the license?
    - Is it open source?
@@ -93,7 +96,7 @@ You may have to download a 3rd party component. When this happens, consider the 
    - Popularity – are there a significant number of weekly downloads?
    - Is the downloads line graph going up or down?
 1. Look for tests, ideally with 100% test coverage.
-1. Look at install size with [Package Phobia](https://packagephobia.now.sh/).
+1. Look at the bundle size with [Bundle Phobia](https://bundlephobia.com/).
 1. Does it have good/clear documentation?
 1. Is it currently maintained?
    - Look at the number of issues vs. pull requests.
@@ -107,7 +110,9 @@ You may have to download a 3rd party component. When this happens, consider the 
 
 ## Styles
 
-Bundle optimization works best when CSS Modules are imported directly into the JavaScript components that use them (a better [TypeScript][] experience is being talked about in [this CRA issue](https://github.com/facebook/create-react-app/issues/5677)). The coupling here is a bit unfortunate for some projects (e.g., libraries), but pretty standard for applications. If there is a specific reason for ultimate **portability** in your project, consider the [container paradigm][] to decouple the styles from the components, each component defining its own styles via a styles interface.
+Bundle optimization works best when CSS Modules are imported directly into the JavaScript components that use them (a better [TypeScript][] experience is being talked about in [this CRA issue](https://github.com/facebook/create-react-app/issues/5677)).
+
+The coupling here is a bit unfortunate for some projects (e.g., libraries), but pretty standard for applications. For ultimate **portability**, reusable components should be provided their styles externally.
 
 ### Resources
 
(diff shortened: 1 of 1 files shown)


design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+  - [**Quality** by Code Climate](https://codeclimate.com/quality/) provides automated code review for test coverage, **maintainability** and more so that you can **save time** and merge with confidence.
+  - Code Climate’s engineering process insights and automated code review for GitHub and GitHub Enterprise help you ship **better software, faster**.
+
+### Resources
+- [Code Climate for TypeScript is here!](https://codeclimate.com/changelog/5a147fcac7d08102b700084e/)
+- [Code Climate | TSLint](https://docs.codeclimate.com/docs/tslint)
+
+## Performance Testing
+
+_TODO: Is there a way for us to determine if components overuse actions or API calls?_
```
jedmao: @kandy I reached out to you on Slack about this on Dec. 14th, but no response. Do you have any more details as to what your expectations are for performance testing a front-end application here? Personally, I think this is more of a DevOps concern, outside of what CodeClimate's tools already provide (mentioned above in the "Code Quality" section).

kandy: I expect to see such hi-level elements here:
- All S0, S1 scenarios from end to end *MUST* be automated and should be monitored and compare metrics as total load time, average action time / step. Degradation of metrics *MUST* be prohibited.
- All common reusable component should be covered with separate isolated performance test.
- All measurement SHOULD be tested on average user data  (like Medium profile) and on edge case (like 10x) to see scalability gaps

jedmao: @kandy are you aware of any specific tools that would aid in meeting your requirements? I'm looking for more specific guidance here.

I agree that performance tests should be separate/isolated, but can you be specific about what libraries you would use for this? Also, when you say separate, do you mean a separate repo too?

If we can't lock down anything more specific would you be OK with me copying what you write above directly into this "Performance Testing" section and tagging you for more information?

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+
+Creating a new React project has never been easier with [CRA][], which comes out-of-the-box with [hot reloading](https://webpack.js.org/concepts/hot-module-replacement/). This increases **productivity** by giving developers [short, positive feedback loops](https://www.youtube.com/watch?v=avKBrgIQXzk). Hot reloading is [also achievable in React Native](https://facebook.github.io/react-native/blog/2016/03/24/introducing-hot-reloading.html). Another opportunity for short, positive feedback loops is testing with [Jest][] and [Enzyme](https://airbnb.io/enzyme/). Tests are fastest when writing unit tests with the [Shallow Rendering API](https://airbnb.io/enzyme/docs/api/shallow.html) and reserving the [Full Rendering API](https://airbnb.io/enzyme/docs/api/mount.html) for integration tests. The better the development experience in these cycles, the better **quality** we achieve as a result. [Jest's snapshot testing](https://jestjs.io/docs/en/snapshot-testing) is another way to ensure **quality** in that we don't inadvertently introduce breaking changes.
```
DrewML: `[CRA][]` here is missing the link value

design-documents/frontend/technical-vision.md
jedmao: @kandy I have some more questions about your post:
1. What are S0, S1 scenarios? I'm no performance architect.
   - How do you automate them, exactly?
   - Wouldn't prohibiting degradation of metrics prevent almost any feature from being added? Doesn't seem realistic.
1. Define "common reusable component". Wouldn't that be 99% of the components, if written properly and decoupled?

design-documents/frontend/technical-vision.md
jedmao: Look at line 138.

design-documents/frontend/technical-vision.md
kandy: >  are you aware of any specific tools that would aid in meeting your requirements? I'm looking for more specific guidance here.

I am not a front-end specialist, so I can only offer sitespeed.io, but at a high level, even if the tool does not exist, it should be developed.

> I agree that performance tests should be separate/isolated, but can you be specific about what libraries you would use for this? 

Don't know why you ask me :)

> Also, when you say separate, do you mean a separate repo too?

I did not think about it and it does not relate directly to the performance. But when you asked me, I noticed that there was nothing in the document about managing dependencies between packages, but it should there

> What are S0, S1 scenarios? How do you automate them, exactly?

High severity scenario that MUST work in the product (for example in Magento it's "add product to cart and checkout flow"). 
We have Performance Acceptance Build that run for every PR, and compare execution time for the scenarios on base branch vs PR branch

> Wouldn't prohibiting degradation of metrics prevent almost any feature from being added? Doesn't seem realistic.

It prohibited without additional approve of Performance Architect and Product Owner.

> Define "common reusable component". Wouldn't that be 99% of the components, if written properly and decoupled?

I believe "will be used 3+ times" is good definition.

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+- Extensibility
+- Maintainability
+- Consistency
+- Performance
+- Accessibility
+- Internationalization
+- Productivity (Efficiency)
+- Quality (Reliability)
+
+## [React](https://reactjs.org/)
```
kandy: I believe that better to create requirement and not provide the solution as the requirement. The almost all that you will say about "React" will be the same for Vue.Js

design-documents/frontend/technical-vision.md
jedmao: I think after this conversation I can add a small paragraph that reads something like this:
> Ideally, for each front-end project, there will be a separate project that covers both performance and critical-path testing. This project is separate so that it doesn't cripple the short feedback loops of feature development. Additionally, we would do well to have additional performance testing around the most commonly used components (e.g., used in 3+ places).

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+
+[Type definitions](https://www.typescriptlang.org/docs/handbook/declaration-files/consumption.html) are provided for [many of the most popular npm packages](http://definitelytyped.org/), meaning developers can increase **productivity** by staying in-editor instead of constantly task switching and looking up documentation online. Even internally, there is less need to open a local file when you can extract most of the relevant information you need via [Intellisense](https://code.visualstudio.com/docs/editor/intellisense#_intellisense-features).
+
+_Disclaimer: though possible to introduce [TypeScript][] into an existing [CRA][] app and though [TypeScript][] can co-exist with JavaScript, it may not make sense to introduce it into an existing project if you feel it may hinder momentum._
+
+### Technologies
+- [Adding TypeScript to CRA](https://facebook.github.io/create-react-app/docs/adding-typescript).
+- Code **consistency** is enforced via [configuration](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html) and [linting with TSLint](https://palantir.github.io/tslint/).
```
DrewML: Can we add in a section for Prettier and be more clear about what category of rules we expect to be enabled with TSLint? We do too much bikeshedding on formatting in existing projects and I'd like to leave that as a relic of the past

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Business Logic Repository
+1. Written and [documented](https://github.com/Microsoft/tsdoc#tsdoc) in [TypeScript][].
+1. [Semantic versioning](https://semver.org/) to ensure that progress can be made without breaking things for existing - consumers.
+1. 100% unit test coverage via [Jest][] for **consistency**.
+
+For **performance** reasons, JavaScript should only be used for cheap operations on the front end, reserving expensive operations for a services layer. It's worth calling out that we could achieve even more code sharing if this API were written in JavaScript. For practical reasons, however, it makes more sense to leverage internal PHP resources for this task.
+
+## Browser Support
+All users should have a **consistent** user experience. For the best **performance**, those with the latest browsers should not be penalized by incurring an additional payload hit than those with older browsers. To achieve this goal, we download just the [polyfills](https://remysharp.com/2010/10/08/what-is-a-polyfill) needed for our sites, tailored to each browser via [Polyfill.io](https://polyfill.io) feature detection.
```
DrewML: Since we haven't used `Polyfill.io` in a product yet, I'm hesitant to provide an explicit requirement for it. I think we can leave that in as a possible route to accomplish the goal, but just be more clear about the real intent here. Regarding this line:

> those with the latest browsers should not be penalized by incurring an additional payload hit than those with older browsers

I think this won't be clear to those unfamiliar with transpilation/polyfills _why_ the extra hit for modern browsers would happen.

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+## React UI Component Library
+
+_Consistency, Productivity, Scalability, Maintainability, Accessibility._
+
+We should lean against downloading various UI components from different sources for the following reasons:
+
+- **Inconsistent** implementations, documentation, theming and customization.
+- Limited **accessibility** and support.
+
+[KendoReact | UI for React Developers](https://www.telerik.com/kendo-react-ui/) is a great option for those wishing to get React UI components from the same source. Still, there will be times when you need to download a 3rd-party package.
```
DrewML: I think we ended up discussing reasons offline that Kendo wasn't going to work out well for us and our OSS code. I think we should remove the advocacy for it here since this document will be used more often than not for OSS projects.

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+You may have to download a 3rd party component. When this happens, consider the following criteria before making a final choice:
+
+1. First thing's first – what is the license?
+   - Is it open source?
+   - You may need to reach out to your legal department if things are a bit blurry.
+1. Does the package have a bright future?
+   - Popularity – are there a significant number of weekly downloads?
+   - Is the downloads line graph going up or down?
+1. Look for tests, ideally with 100% test coverage.
+1. Look at install size with [Package Phobia](https://packagephobia.now.sh/).
```
DrewML: Probably meant Bundle Phobia here

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+   - When was the last comment made on them and by whom?
+   - **Reliability** – how serious are the issues?
+1. Search and compare this package with other similar packages to gauge which package is the best choice and feel free to reach out to other team members for guidance.
+
+### Resources
+- [spectrum-css](http://opensource.adobe.com/spectrum-css/2.6.0/docs/)
+
+## Styles
+
+Bundle optimization works best when CSS Modules are imported directly into the JavaScript components that use them (a better [TypeScript][] experience is being talked about in [this CRA issue](https://github.com/facebook/create-react-app/issues/5677)). The coupling here is a bit unfortunate for some projects (e.g., libraries), but pretty standard for applications. If there is a specific reason for ultimate **portability** in your project, consider the [container paradigm][] to decouple the styles from the components, each component defining its own styles via a styles interface.
```
DrewML: Think it would make sense to remove the reference to the container paradigm here. From what I'm seeing, it's going to quickly disappear in lieu of similar things that can be done with hooks.

Opinion: For now, this section should stick to the _what_, rather than the _how_.

design-documents/frontend/technical-vision.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Technologies
+- Create a new React project with [CRA][].
+- [Redux](https://redux.js.org/) as a predictable state container.
+   - Handling of data fetching and side effects now have first class support with [React Suspense, React.lazy](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactlazy-code-splitting-with-suspense) and [hooks (proposal)](https://reactjs.org/docs/hooks-intro.html).
+   - Logging with [redux-logger](https://github.com/evgenyrodionov/redux-logger).
+- **Internationalization** with [react-intl](https://www.npmjs.com/package/react-intl).
```
DrewML: There are a _massive_ number of options in this space, some specific to React and some a bit more abstract. What's the motivation for recommending this particular lib?

design-documents/frontend/technical-vision.md
jedmao: Right, not for us, but this document is public, so I had to remove certain information, remember? That said, KendoReact is still a great alternative for those who have no other options. The ultimate point here is that if they have the ability to pay for the product, it's much better to get all your components from a single source than to have completely different paradigms scattered everywhere.

design-documents/frontend/technical-vision.md
jedmao: I like that suggestion.

design-documents/frontend/technical-vision.md
jedmao: I actually learned recently that we (Adobe) don't "trust" polyfill.io:

> polyfill.io won’t detect usage, so it’ll just download all polyfills for your target browser, even if you don’t use like 90% of them.

☝️ A bit out of place, as you can tweak your Polyfill.io request to request only the features you use.

> But I’d argue the performance hit from the additional bandwidth is less than the perf hit from having an additional connection to polyfill.io (which will also break your website if it goes down!)

☝️ Good argument, but a bit subjective on the first point.

> If polyfill.io is loading external scripts, I think we’d run into PCI compliance issues.

☝️ That's something with which I can't really argue.

In conclusion, I'll generalize this point instead of the specific usage of polyfill.io.

design-documents/frontend/technical-vision.md
jedmao: Also, to be clear, the solution here and really for most of this document isn't really a set of requirements so much as it is a guide to help people sift through the plethora of technology choices available in the JS ecosystem. The first sentence of this document states that it is "a guide", so I think that should be pointed out.

design-documents/frontend/technical-vision.md
jedmao: I didn't want to litter this document with all the plethora of choices available. Instead, this document is a guide. Here's a viable solution that we already use and has worked for us. Maybe you find something better in the future or even today that warrants a discussion point against this technology choice. By all means, present that choice as a better alternative to react-intl and defend it with argument points and we can change this item to something else.

This document doesn't have to be 🔐. I see it as a living document that:
1. Highlights or implies what we want (the general requirement).
1. Provides a viable solution to that requirement so you can hit the ground running without going on a technology hunt every time you need a technology in your application.

From there, I'm open to critical feedback against any proposed technology, so long as you pitch a good alternative.

design-documents/frontend/technical-vision.md
jedmao: Yep. I need to do a bit more research here as to getting TSLint and Prettier and EditorConfig to work well together. All of this will change, of course, once [these TypeScript roadmap items](https://github.com/Microsoft/TypeScript/issues/29288#roadmap-overview) are addressed.

design-documents/frontend/technical-vision.md
DrewML: > The ultimate point here is that if they have the ability to pay for the product, it's much better to get all of your components from a single source than to have completely different paradigms scattered everywhere.

I think you just wrote up the better generic description to replace a recommendation of a specific product :)

jedmao: If you're going to pay for a React component library though, I really do recommend KendoReact. It's a solid product.

design-documents/frontend/technical-vision.md
jedmao: Addressed in 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

design-documents/frontend/technical-vision.md
jedmao: Addressed in 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

design-documents/frontend/technical-vision.md
jedmao: Addressed in 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

design-documents/frontend/technical-vision.md
jedmao: Addressed in 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

design-documents/frontend/technical-vision.md
jedmao: Addressed in 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

design-documents/frontend/technical-vision.md
jedmao: Addressed in 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

jedmao:
@kandy, @DrewML see 4f625a47b50e0c0ac67ccbf5fed03eff17b94d40

PR #50, opened 2018-12-03
inter-service-communication doc

AlexMaxHorkun:
This document describes inter-services communications


design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+[here](invoker.md).
+ 
+#### Gateway
+For actualy delivering remote service requests RESTful API will be used - it's an established way to execute requested methods and to
+serialize arguments and return values. For sending HTTP requests [Guzzle](https://github.com/guzzle/guzzle) library will be used - it's
+a stable and well supported library which is capable of sending concurrent requests and allows wide customization.
+Guzzle will be configured to use cUrl (cUrl multi) as a backend. When asynchronous requests are sent via Guzzle it returns a promise which can easily
+be converted to Magento's promise.
+#### Lookup
+The information on how to reach services will be provided in env.php like this:
```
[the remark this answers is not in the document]
AlexMaxHorkun: in such deployment base URLs would lead to the same address. I think described approach fits both cases - with service mesh and without one.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+ 
+#### The invoker
+The invoker will be a part of Magento\Framework and will be present in every node installation. Invoker's job is to find a remote node's
+address based on given class name and method, send the request, receive response and convert it to either requested service method return type
+or an exception if it was thrown by the remote service. It returns a promise to support asynchronous services and because
+the requests to remote services will be asynchronous (depends on implementation). More details on the invokers API and SPI
+[here](invoker.md).
+ 
+#### Gateway
+For actualy delivering remote service requests RESTful API will be used - it's an established way to execute requested methods and to
```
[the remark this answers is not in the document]
AlexMaxHorkun: I think right now we use what we already have - adopting the new technology to call certain classes is unnecessary right now. We already have like 3 ways to do that

buskamuza:
Duplicate of https://github.com/magento/architecture/pull/42
Doesn't explain basics of communication between services. Do we always use async calls, or are there cases (or transition period) when we will use sync calls? How errors should be handled? Do we use service mesh or other approach? What functionality should be handled by the framework (errors handling, authorization and authentication involvement, retries and timeouts, etc) and how exactly?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
+### Inter-service communications
+This documents describes how exactly services will be communicating in a distributed setup.
+#### Service declaration
+Services will be declared in Api modules using interfaces. These modules will be installed on all nodes.
```
melnikovi: Do you mean installed on all nodes of particular service? For instance, ServiceBApi module will be installed only on ServiceB? Please clarify.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+  
+    public function getName(): string;
+}
+```
+The API modules will just contain these declarations and nothing else (no di.xml, schema.xml, webapi.xml etc).
+ 
+Even services that are not written using Magento (or even PHP) will be declared in this way for nodes using Magento.
+ 
+#### Publicly available Web APIs
+To enable publicly available APIs (that frontend and 3rd party systems can use) the _front node_ will have to have -Module-Webapi
```
melnikovi: BFF will expose aggregated operations and operations that needs to be proxied to a service. As I understand this point is about operations that need to be proxied? Maybe we can solve this with configuration?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+        'Magento\\Catalog' => 'http://catalog.mydomain.com/',
+        'Magento\\Checkout' => 'https://checkout.mydomain.com/',
+        'Magento\\Customer' => 'http://192.168.1.1/',
+    ],
+];
+```
+The invoker will look up the information based on provided services' class names.
+ 
+#### Changes to RESTful Web API mechanism
+We need to able so send and receive more information when using RESTful web API for inter-service communication than for regular
```
melnikovi: What do you mean by regular endpoints, public APIs? Public API will be defined on the BFF. Services define private APIs. Only other services and BFF can use APIs of the services. Based on this we have clear distinction.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+<?xml version="1.0"?>
+<services xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Webapi:etc/inter_webapi.xsd">
+
+    <!-- Product Service -->
+    <service class="Magento\Catalog\Api\ProductRepositoryInterface" method="save"/>
+</services>
+```
+ 
+We will have to pass authenticated user information to remote services so we'll add _X-Magento-User-Type_ and _X-Magento-User-Id_
```
melnikovi: Propose to move this to proposal about [authentication and authorization](https://github.com/magento/architecture/pull/48). Current idea is that authentication will be performed on BFF and each service will perform authorization.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+headers to requests for a _UserContextInterface_ instance to be initialized. There's no need for nodes to validate auth information of
+incoming
+requests since those requests will come from other nodes in a closed network or they will be encrypted. These headers will only be
+accepted by endpoints declared in _inter_webapi.xml_ files.
+ 
+The way exceptions thrown by services are serialized right now for RESTful web API needs to be changed for inter-service endpoints.
```
melnikovi: Would it be possible to reuse existing configuration for web API?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+Exceptions are serialized without their type information provided - this info is crutial for services to communicate with each other so
+the type information will be added to the output.
+ 
+_inter_webapi.xml_ files will be contained in separate <ModuleName>InterApi modules and will have to be installed
+on nodes accepting remote inter-service requests.
+ 
+##### Changes for retries mechanism
```
melnikovi: I think we shouldn't try to implement this functionality ourselves. What are the reasons to not use service mesh?

design-documents/promises.md
```diff
@@ -0,0 +1,138 @@
+### Why?
```
melnikovi: Promises related part please move to separate document.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+Front node is a Magento installation with Framework and Webapi/GraphQl modules.
+It will handle actual authentication and authorization of incoming requests and pass the auth info to remote services
+as well as accepting requests from clients (frontend, 3rd party systems).
+ 
+This is the node that will be available on a public network.
+
+#### Flow
+Entities:
+* Front node
+  * Modules installed:
```
melnikovi: Putting all code in one module to enable remote communication for a module makes it easier to develop but has deployment complications. As there could be changes that are relevant to multiple services, changing module for one service will require redeploying other services. Need to think if we can make it more granular to avoid this issue.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+ 
+#### Changes to RESTful Web API mechanism
+We need to able so send and receive more information when using RESTful web API for inter-service communication than for regular
+communications so to distinguish regular endpoints and endpoints meant for inter-service communications we will declare them in a
+seperate yet similar xml files - _inter_webapi.xml_. Since we declare services by declaring PHP interfaces there is no reason to provide
+human-readable URLs for endpoints meant for inter-service communications, thus we'll make it possible to omit _url_ attribute of the
+_route_ element in inter_webapi.xml and have it autogenerated for us. This way _the invoker_ will be able to determine URL
+for a service based on class name/method name provided. The _method_ attribute will be dropped because otherwise both nodes
+who are calling remote services and nodes hosting the services will have to have the inter_webapi.xml installed.
+Also caching service call results via
```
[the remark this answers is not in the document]
AlexMaxHorkun: I think public and inner API endpoints should differ in that regard - when calling a remote service we expect the same
fresh result as we would by using it locally, or at least explicitly cached by the service

design-documents/service-isolation/inter-service-communication.md
AlexMaxHorkun: I am not following

design-documents/service-isolation/inter-service-communication.md
AlexMaxHorkun: if a service adds a remote API (endpoint) that will not affect other services communicating with it using an older declarations

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,294 @@
...
+* Checks inter_webapi.xml in _CatalogInterApi_ module to find if ProductRepositoryInterface::save() is available
+* Reads user type and user ID information from the request and creates a UserContextInterface instance
+* Deserializes body into the ProductInterface instance (this node also has _CatalogApi_ modules and gets the service's arguments it provides)
+* Catalog\Model\Product is actually used as the argument because this node has _Catalog_ module installed
+* Catalog\Model\ProductRepository is used as the service instance because this node has _Catalog_ module with it's di.xml
+and no _CatalogProxy_ module
+* ProductRepository::save() is executed
+* ProductRepository::save() uses CustomerProductManagerInterface::process from _CustomerApi_ module
+* this node has only _CustomerProxy_ module so an instance of CustomerProxy\CustomerProductManagerProxy is used
+* it sends a request to the _customer node_ using the invoker
```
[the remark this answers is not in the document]
AlexMaxHorkun: it's just an example of inter-service communications, let's just assume product repo uses CustomerProductManagerInterface service for some business logic

design-documents/service-isolation/inter-service-communication.md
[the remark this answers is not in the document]
melnikovi: Not sure that using hostnames to enable load balancing is typical for deployments. I'm more inclined to make it generic resource identifier. Let me look into it.

design-documents/service-isolation/inter-service-communication.md
melnikovi: What is the purpose of default implementation? Development?

design-documents/service-isolation/inter-service-communication.md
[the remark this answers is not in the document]
melnikovi: gRPC has benefits that can be useful for us
* Allows for low latency connections, connection can be reused
* No limitations on the  set of primary function calls (for rest, it's read, replace, modify, delete)
* Seems like it's specifically designed to mask difference between local and remote calls

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,301 @@
...
+namespace Magento\ServiceBApi\Data;
+
+interface BDataInterface
+{
+    public function getId(): string;
+  
+    public function getName(): string;
+}
+```
+The API modules will just contain these declarations and config files (like acl.xml, config.xml) describing module
```
melnikovi: acl and config files only needed on BFF, they probably be in *Admin modules.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,301 @@
...
+without describing actual implementation.
+ 
+Even services that are not written using Magento (or even PHP) will be declared in this way for nodes using Magento.
```
melnikovi: What do you mean here, could you elaborate?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,301 @@
...
+ 
+#### Publicly available Web APIs
+To enable publicly available APIs (that frontend and 3rd party systems can use) the _front node_ will have to have -Module-Webapi
+or -Module-Graphql modules installed.
+ 
+#### Real and proxy modules
+If a node wants to use services locally then it will just have real modules installed (e.g. "Catalog"), if a node is supposed
+to access services remotely it will have proxy modules installed (e.g. "CatalogProxy").
+ 
+Proxy modules will contain services/DTOs implementations and a di.xml file to declare preferences for proxy implementations to be
```
melnikovi: DTO implementations should be in *API modules.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,301 @@
...
+```php
+namespace Magento\ServiceBProxy;
+
+class BManagerProxy implements Magento\ServiceBApi\BManagerInterface
+{
+    ....
+    
+    public function save(BDataInterface $data): BDataInterface
+    {
+        $promise = $this->invoker->invoker(
```
melnikovi: What is the benefit of using promise here?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,301 @@
+### Inter-service communications
```
melnikovi: Another idea someone mentioned is making service a demon. This would allow save time on application bootstrap.

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,107 @@
...
+Creates HTTP request based on request data
+```php
+interface RequestGeneratorInterface
+{
+    public function generate(InvokeRequestInterface $invokeRequest): \Psr\Http\Message\RequestInterface;
+}
+```
+##### Transport
+```php
+interface TransportInterface
```
melnikovi: Need more information. Who is client of this interface?

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,107 @@
...
+interface ServiceSourceRepositoryInteface {
+    public function find(string $class, string $method): ServiceSourceInterface;
+    
+    public function getAll(): ServiceSourceInterface[];
+}
+```
+##### Remote request data
+Information about the request to be sent for a remote service
+```php
+interface InvokeRequestInterface {
```
melnikovi: Does developer need to know about this interface? Need to provide more details for each interfaces, not clear how they are used.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,309 @@
...
+we will avoid executing duplicate calls when retries mechanism is applied.
+ 
+The timeout and tries limit will be configured via di.xml (see the invoker API and SPI [here](invoker.md)).
+    
+#### Service mashes and balancers
+Since we going to contact services via HTTP it is possible for configured service base URLs to actually lead to service mashes or
+balancers - that way the retries and lookup mechanism may be delegated.
+
+#### MQ RPC for gateway
+Magento has another way to remotely call services - RPC via a message queue. The problem with it is that right now it can only work
```
melnikovi: Could you elaborate more on this. What kind of gateway you propose to use message queue for?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,309 @@
...
+with strings as arguments for service contracts and daemons written in PHP to process queues will show slower results in handling
+multiple requests than web servers like Nginx or Apache when using RESTful web API for gateway.
+
+#### Authentication and authorization
+More details [here](distributed-auth.md)
+
+#### Front node (BFF)
+Front node is a Magento installation with Framework, *Api, *Proxy and *Webapi/*GraphQl modules.
+It will handle actual authentication and authorization of incoming requests and pass the auth info to remote services
```
melnikovi: Propose to remove this, this is covered in other proposal.

design-documents/service-isolation/inter-service-communication.md
AlexMaxHorkun: We would have to rewrite every module. DTOs in API modules are just declarations of data types this module's service contracts accept/return, other modules should not care what the actual implementations are. Plus we allow some services to be implemented using different programming languages - implementations for DTOs for those service contracts in PHP would be redundant.

design-documents/service-isolation/inter-service-communication.md
AlexMaxHorkun: to be able to work with asynchronous operations - the invoker sends requests to other services in an asynchronous manner

design-documents/service-isolation/invoker.md
AlexMaxHorkun: invoker will use it after using RequestGeneratorInterface to generate an HTTP request based on InvokeRequestInterface

design-documents/service-isolation/invoker.md
AlexMaxHorkun: I'll add more use cases

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,328 @@
...
+* ProductRepository::save() uses CustomerProductManagerInterface::process from _CustomerApi_ module
+* this node has only _CustomerProxy_ module so an instance of CustomerProxy\CustomerProductManagerProxy is used
+* it sends a request to the _customer node_ using the invoker
+* the result is used by the ProductRepository::save() method
+* ProductRepository::save() finishes and returns persisted Model\Product instance
+* RESTful API mechanism serializes Model\Product to CatalogApi\Data\ProductInterface since CatalogApi\ProductRepositoryInterface::save()
+states that the return type is ProductInterface
+* HTTP response is generated
+ 
+_front node_
```
melnikovi: Please rename to BFF everywhere.

melnikovi:
Nee to describe how do we pass correlation identifier for all calls that are part of the same flow.

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,122 @@
+### Invoker API and SPI
+Responsible for contacting remote services
+#### API
+##### Remote call response
+Result of executing the call.
+Can be used by 3rd party devs to add data to remote calls results or read it when they choose to modify
+remote calling mechanism.
+```php
+interface InvokeResponseInterface {
```
melnikovi: Are we going to have many response interfaces in the library? Can be just ResponseInterface probably.

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * @return mixed Value that remote service returned.
+     */
+    public function getResult();
+    
+    /**
+     * Exception that remote service has thrown, if any.
+     *
+     * @return \Throwable|null
```
melnikovi: Please double check that current framework will be able to extract/hydrate \Throwable.

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,122 @@
...
+    public function getException(): ?\Throwable
+}
+```
+##### Remote Invoker
+Used to make a call to a Magento server.
+Can be used by 3rd party devs to write proxy services and modify invoker functionality/gateway.
+```php
+interface InvokerInterface {
+    /**
+     * @param string $class
```
melnikovi: This should names an interface?

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,122 @@
...
+     ): PromiseInterface;
+}
+```
+ 
+#### SPI
+##### Base URL information
+Nodes data.
+Used to access nodes and provide requests for them.
+```php
+interface ServiceSourceInterface {
```
melnikovi: I think we discussed, but it would be good to provide more description why do we need this and ServiceSourceRepositoryInteface interfaces.

design-documents/service-isolation/invoker.md
melnikovi: Did you add more use cases?

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,122 @@
...
+    public function getUrl(): string;
+    public function isSignatureRequired(): bool;
+}
+```
+##### Remote service sources reposit.ory
```
melnikovi: repository, not reposit.ory. Please describe why do we need this interface.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,328 @@
...
+* Invoker looks at env.php to find Magento\Catalog's URL (something like https://catalog.mydomain.com/ or http://182.1.1.2)
+* Invoker serializes the argument as the argument type provided in ProductRepositoryInterface::save() declaration
+* Invoker gets user type and user ID from the UserContextInterface instance
+* Invoker composes URL to access inter-service endpoint on the _Catalog_ node
+like __http://catalog.mydomain.com/rest/inter/catalog/product-repository/save__
+* Invoker sends request to _Catalog_ node containing serialized ProductInterface argument, user type, user ID
+  and returns a promise that will be fulfilled when the request finishes and return value is unserialized
+* the request is sent using Guzzle library and it's sent asynchronously (concurrently with other requests)
+ 
+_catalog node_
```
melnikovi: Let's use service instead of node to be consistent with other documents.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+        }
+        
+        return $result->getResult();
+    }
+    
+    ....
+}
+```
+
+Proxy services will use _InvokerInterface_ to call remote services. It will return a promise which will pass a service execution result
```
melnikovi: Some information about promises left here that need to be removed.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+        $this->name = $name;
+        $this->id = $id;
+    }
+    
+    public function getId(): ?string
+    {
+        return $this->id;
+    }
+    
+    public function setId(string $id): void
```
melnikovi: Setters in DTOs are optional and can be omitted.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+ 
+#### The invoker
+The invoker will be a part of Magento\Framework and will be present in every node installation. Invoker's job is to find a remote node's
+address based on given class name and method, send the request, receive response and convert it to either requested service method return type
+or an exception if it was thrown by the remote service. It returns a promise to support asynchronous services and because
+the requests to remote services will be asynchronous (depends on implementation). More details on the invoker's API and SPI
+[here](invoker.md).
+ 
+#### Gateway
+For actually delivering remote service requests RESTful API will be used - it's an established way to execute requested
```
melnikovi: REST probably is not the best protocol for communication between services. We also can't violate protocol specification.

I suggest to mention some pros and cons for each of the options (REST, gRPC, RPC + REST, GraphQL), say that we need to investigate more gRPC and decision will be make in separate proposal. You can copy and paste what I have below.

REST
Pros
* We already have framework that provides support for it

Cons
* Not suitable for communication between services because it has limitations on the set of primary function calls (read, replace, modify, delete)
* Has large size of the message

gRPC
Pros
* Specifically designed for communication between services
* Small size of the message
* Support of HTTP 2
* Client code generation support for many programming languages (need to confirm if it works for PHP)
* Potentially can save on network communication if services would be demons that invokes service application

Cons
* There are some support limitations in PHP (rapidly growing though)

RPC + REST
Pros
* No limitations on the set of primary function calls 

Cons
* Response code for all requests is 200, which makes it hard to cache
* No versioning
* No strict schema

GraphQL
Pros
* ?

Cons
* Has complication logic behind it

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+class' methods and to
+serialize arguments and return values. Existing web API gateway will be used as an RPC to save time on developing new
+gateway - there is no need to preserve RESTful principles (like using certain HTTP methods for certain types of operations)
+for inter-service communications since requests will be generated automatically and no human developer will be working
+with inter APIs explicitly.
+ 
+For sending HTTP requests [Guzzle](https://github.com/guzzle/guzzle) library will be used - it's
```
melnikovi: Not relevant until we decide on protocol.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+a stable and well supported library which is capable of sending concurrent requests and allows wide customization.
+Guzzle will be configured to use cUrl (cUrl multi) as a backend. When asynchronous requests are sent via Guzzle it
+returns a promise which can easily be converted to Magento's promise.
+#### Lookup
+The information on how to reach services will be provided in env.php like this:
+```php
+return [
+    'remote_deployment' => [
```
melnikovi: I propose something like this ['services' => ['catalog' => ['address' => 'catalog']]]. Whether we need to specify protocol is not clear at this point (depends on communication protocol we going to use), so I would suggest to omit it for now or make separate parameter.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+        'nodes' => [
+            'Magento\Catalog' => ['url' => 'http://catalog.mydomain.com/'],
+            'Magento\Checkout' => ['url' => 'https://checkout.mydomain.com/'],
+            'Magento\Customer' => ['url' => 'http://192.168.1.1/'],
+        ]
+    ],
+];
+```
+The invoker will look up the information based on provided services' class names.
```
melnikovi: Interface.

design-documents/service-isolation/inter-service-communication.md
melnikovi: I would suggest to call it public, because it could be confusing.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+
+#### MQ RPC for gateway
+Magento has another way to remotely call services - RPC via a message queue. The problem with it is that right now it can only work
+with strings as arguments for service contracts and daemons written in PHP to process queues will show slower results in handling
+multiple requests than web servers like Nginx or Apache when using RESTful web API for gateway.
+See _Magento\Framework\MessageQueue\Rpc\*_ and _Magento\Framework\MessageQueue\UseCase\RpcCommunicationsTest_ for
+more information. It is not a finished mechanism and it will take longer to adapt it for inter-service communications.
+ 
+#### gRPC for gateway
+Remote calls to services must coexist with local calls, and on the service it would make sens to use an RPC for that but
```
melnikovi: Did you find that it's not possible to add headers when using gRPC?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+Front node is a Magento installation with Framework, *Api, *Proxy and *Webapi/*GraphQl modules.
+ 
+This is the node that will be available on a public network.
+
+#### Flow
+Entities:
+* Front node
+  * Modules installed:
+    * Magento\Framework
+    * Magento\Authorization (implementation, APIs for auth are moved to Framework)
```
melnikovi: Maybe we should leave this module as is, I'm not sure that how authorization should change.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,319 @@
...
+* Catalog node
+  * Modules installed:
+    * Magento\Framework
+    * CatalogApi
+    * Catalog
+    * CatalogInterApi
+    * CustomerApi
+    * CustomerProxy
+  * env.php has base URL for Magento\Customer
+* Customer node
```
melnikovi: Need to provide better example, because it could be hard to understand why Customer module need to know about Catalog.

design-documents/service-isolation/invoker.md
AlexMaxHorkun: don't want to mix it with actual network responses

design-documents/service-isolation/invoker.md
AlexMaxHorkun: even if it doesn't we can create a middleware DTO to pass exception information between services

design-documents/service-isolation/invoker.md
```diff
@@ -0,0 +1,122 @@
...
+##### Remote Invoker
+Used to make a call to a Magento server.
+Can be used by 3rd party devs to write proxy services and modify invoker functionality/gateway.
+```php
+interface InvokerInterface {
+    /**
+     * @param string $class
+     * @param string $method
+     * @param array $arguments Keys - argument names.
+     * @return PromiseInterface Will provide InvokeResponseInterface when resolved,
```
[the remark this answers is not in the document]
AlexMaxHorkun: Invoker must return a promise, that proposal is a separate one - I don't specify where that interfaces comes from. But it has to be a promise

design-documents/service-isolation/inter-service-communication.md
ravmenon: Mentioning my  comments as discussed during the last meeting:

I do find REST with JSON easier to debug - you may be able to use the browser directly or tools like Postman and/or curl, wireshark etc...

However there is no doubt gRPC with protobuffs is more efficient in space and time.  If we do go this route few things to confirm:

- With protobuffs, what happens if field ordering is changed as new fields are added on one side first (client or server)?. JSON can easily deal with this, but not sure if binary protocols are that flexible.  
- As most of the payload data types might be string type (e.g. in catalog with fields like long description, sku etc..), will there still be size benefits with protobuffs?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+In default .htaccess and nginx.conf.sample files requests to /rest/inter/* will be disabled because it would be dangerous to
+allow calling any installed services from outside - only other service nodes must be able to do this. That's why it will
+be recommended to have service nodes reside in a private network (or have firewalls allowing only connections
+from trusted nodes) and then enabling requests to /rest/inter/* in .htaccess/nginx.conf.
+ 
+For cases when a service node must be publicly accessible (like BFF having some services deployed locally) developers
+would have to create SSL client-side certificates and then create a separate virtual host processing requests to /rest/inter/*
+that would require all client service nodes to use this certificates to sign their requests.
+ 
+This measures will prevent anyone other than service nodes using service comm gateway.
```
paliarush:
```suggestion
This measures will prevent anyone other than service nodes using service comm gateway.
```
Seems incomplete

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+    
+    ....
+}
+```
+
+Proxy services will use _InvokerInterface_ to call remote services. It will return a promise which will pass a service execution result
+to _then_ callback and an exception to _otherwise_ callback if something during the remote call went wrong (like a network error). The
+service execution result may contain the return value of a service or an exception it has thrown. The promise that invoker returns may
+be used in both synchronous and asynchronous ways to be able to work with both types of services. More details about the promise
+[here](../promises.md).
```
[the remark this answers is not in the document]
AlexMaxHorkun: the files is in the different design doc which will be merged right after this one.

Requests to another services will be asynchronous, but only our new service contracts will be async - we need to support the old ones. Promises are a great mechanism to support both async and sync operations

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+```
+Usually DTOs implementations are models in real modules but we need a class to be initialized when client code communicates with a
+remote service and for that simple DTOs will suffice.
+ 
+#### The invoker
+The invoker will be a part of Magento\Framework and will be present in every node installation. Invoker's job is to find a remote node's
+address based on given interface name and method, send the request, receive response and convert it to either requested service method return type
+or an exception if it was thrown by the remote service. It returns a promise to support asynchronous services and because
+the requests to remote services will be asynchronous (depends on implementation). More details on the invoker's API and SPI
+[here](invoker.md).
```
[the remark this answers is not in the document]
AlexMaxHorkun: works fine for me

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+```php
+namespace Magento\ServiceBProxy;
+
+class BDataProxy implements BDataInterface
+{
+    private $id;
+    
+    private $name;
+    
+    public function __construct(string $name, ?string $id) {
```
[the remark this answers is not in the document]
AlexMaxHorkun: what do you mean?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+#### Possible gateways
+__REST__
+
+Pros
+* We already have framework that provides support for it
+* Established way to call service contracts
+* Easy to adopt by developers creating services with tools other then Magento
+
+Cons
+* Not suitable for communication between services because it has limitations on the set of primary function calls (read, replace, modify, delete)
```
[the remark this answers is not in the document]
AlexMaxHorkun: I think those endpoints would not be 100% compliant with the standard

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+* Has large size of the message
+* Not all endpoints should be exposed to frontend developers
+(like getting a Gift card account by ID, service for sending emails, service for sending a payment)
+* Mix of endpoints meant for frontend devs and endpoints for other services would potentially create security problems
+* Multiple endpoints for the same service interface/method may exist with different parameters enforced - invoker
```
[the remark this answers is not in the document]
AlexMaxHorkun: How does this contradict with what I've written? Multiple public endpoints may use the save service contract

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+* Would require more development time comparing to using existing gateways
+* Developers that are not using Magento would have to have a daemon processing requests from Magento instances
+* gRPC servers would have to be daemons and PHP doesn't have native async code support which would make daemons processing
+requests much slower than with an HTTP server like apache/nginx
+(daemon would use 1 core, HTTP servers effectively distribute processes between cores)
+* Magento would have to be adopted to be run as a daemon
+* only supports 1 dimensional maps (with values of the same type) and one dimensional arrays which makes sending
+arbitrary structures (like template vars to an email service) impossible
+
+__RPC + REST__
```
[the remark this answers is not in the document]
AlexMaxHorkun: example of arbitrary structures? Template vars in this case

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+* We wouldn't be able to utilize GraphQL's ability to strip properties or load multiple entities at once since we would
+have no way to know what properties client service wanted because client services wouldn't know whether they're calling
+a  remote service or a local one
+ 
+#### Gateway
+For actually delivering remote service requests RESTful API will be used - it's an established way to execute requested
+class' methods and to
+serialize arguments and return values. Existing web API gateway will be used as an RPC to save time on developing new
+gateway - there is no need to preserve RESTful principles (like using certain HTTP methods for certain types of operations)
+for inter-service communications since requests will be generated automatically and no human developer will be working
```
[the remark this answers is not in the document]
AlexMaxHorkun: If I'm not mistaken Anton's document says that in Magento service contracts are declared via interfaces, in a monolith deployment a module can use any service contract of another installed module - same should be preserved for distributed deployment, I don't see any reason to redeclare service contracts in an xml

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+    ],
+];
+```
+The invoker will look up the information based on provided services' interface names.
+ 
+#### Changes to RESTful Web API mechanism
+In services communications we would have to send and receive additional metadata (like forwarding ID of the user who
+originally made the request to BFF etc, providing more detailed information about occurred exceptions) so we'd have to
+clearly distinguish requests from other services from requests from front-ends and 1st thing to do is to accept
+inter-service requests on a different URL - _/rest/inter/*_. A controller will be created similar to
```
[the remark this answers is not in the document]
AlexMaxHorkun: How come? We wouldn't have to manually expose endpoints that are only meant for inter-service communications. We still would have process inter-service requests differently - accept specific headers, serialize exceptions differently etc. Avoiding adding another controller seems like more work than adding one.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+    }
+    
+    ....
+}
+```
+
+Proxy services will use _InvokerInterface_ to call remote services. It will return a promise which will pass a service execution result
+to _then_ callback and an exception to _otherwise_ callback if something during the remote call went wrong (like a network error). The
+service execution result may contain the return value of a service or an exception it has thrown. The promise that invoker returns may
+be used in both synchronous and asynchronous ways to be able to work with both types of services. More details about the promise
```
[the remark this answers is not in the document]
AlexMaxHorkun: Only new service contracts will be returning promises and client services would be able to utilize async service requests. For existing services proxy implementations will always call wait() on promises returned by invoker to receive results in a synchronous manner

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+services written using other technologies than Magento, all the requests to services will generated automatically by
+invoker. Therefore there is no need to assign URLs to service contracts manually - they can be generated automatically
+based on interfaces/methods names.
+
+HTTP methods in RESTful API are used to provide self-documentation for developers writing clients and for browsers to
+adjust their behaviour when processing requests to RESTful APIs. They also can be used for caching rules for web
+servers/proxies like Varnish. In services communications requests are being generated automatically, they don't
+come through browsers and caching rules for servers must be assessed on a case-to-case basis and cache needs to be
+carefully invalidated to avoid breaking services' business logic, therefore there is no need in HTTP methods for
+services communications and invoker will always send POST requests to remote services. Services will accept any
```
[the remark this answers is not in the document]
AlexMaxHorkun: Specific HTTP method doesn't matter, we can use get for inter-service requests just as well

design-documents/service-isolation/inter-service-communication.md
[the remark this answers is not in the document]
joni-jones: Why do we need to separate calls for private and public? For service, it doesn't matter who calls it. And anyway, all "external" communication will be done through the BFF. BFF can add all needed headers and data for the future inter-service calls.

joni-jones:
How the services versioning will be supported by invoker if it uses direct names of interfaces and methods?

design-documents/service-isolation/inter-service-communication.md
joni-jones: Not all programming languages support interfaces. If "tomorrow" we need to write a new service on a different stack, how invoker will call it if the invoker?

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+HTTP method though so developers of services not based on Magento can send requests with methods they feel
+suite a requests best.
+
+While URLs for services will be autogenerated and POST always used to make it easier for Magento's invoker it will still
```
joni-jones: Seems like POST is used to make invoker implementation easier but communication shouldn't depend on implementation details.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+be possible for developers to write services without using Magento (or even PHP) - they would be able to get see URLs
+for certain contracts with a CLI utility and while using POST for every request looks odd, functionally it wouldn't matter.
+So developers could contact Magento services just like any other microservice with RESTful web API.
```
joni-jones: This is controversial to the previous sentence there are all requests will be via the POST method.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+To make it easier for debuging and to perhaps later to be used for data consistency _X-Magento-Correlation-Id_ header will
+be added to requests to remote services with ID generated on BFF before a 1st request to a service. The ID will be
+forwarded between services to identify a single operation.  
+ 
+##### Retries mechanism
+The invoker implementation based on RESTful web APIs will offer default retries mechanism, but developers will be able to disable it
+by using _di.xml_ (see Invoker API and SPI [here](invoker.md)) for when they would want to delegate this responsibility
+to a service mash or a balancer. 
+The timeout and tries limit will be configured via di.xml (see the invoker API and SPI [here](invoker.md)).
+This default implementation may be used by developers who choose to avoid service mashes or for development environments.
```
joni-jones: I think a retries mechanism should a part of the infrastructure, not an invoker's part. For development purposes, probably, you don't need retries at all. And not sure if we are building a system where services could be deployed and located separately, how we can avoid the usage of some kind of service mesh.
For development purposes, we should provide an environment not only for 3rd party developers but for core teams also.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+#### Data consistency
+Mechanisms to preserve data consistency will described later in another document and can be built on
+top of the communications described in this one.
+ 
+    
+#### Service mashes and balancers
+Since we going to contact services via HTTP it is possible for configured service base URLs to actually lead to service mashes or
+balancers - that way the retries and lookup mechanism may be delegated.
+
+#### Deployment
```
joni-jones: Probably the deployment topic should be described in a separate document as it much bigger than described in this proposal and at that moment we still don't have the final solution for services communication.

design-documents/service-isolation/inter-service-communication.md
```diff
@@ -0,0 +1,398 @@
...
+    * Magento\Framework
+    * QuoteApi
+    * Quote
+    * QuoteInterApi
+    * CatalogApi
+    * CatalogProxy
+  * env.php has base URL for Magento\Catalog
+  
+  
+ What will happen:
```
joni-jones: I would like proposing to replace this section by UML sequence diagram

melnikovi:

PR #52, opened 2018-12-05
Caching layer for Magento services

viktym:
Draft proposal


design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+ - In case of "Remote caches dedicated to each service" model service should provide entry point for full cache invalidation
+ - In case of "Remote caches shared across the services" model full cache invalidation can be performed by any service
+
+#### Technologies 
+Today most key/value stores such as Memcached and Redis can store terabytes worth of data. Redis also provides high availability and persistence features. 
+Redis and Memcached offer high performance. Memcached is designed for simplicity while Redis offers a rich set of features that make it effective for a wide range of use cases.
+Unlike Memcached, Redis offers snapshots, replication, transactions, advanced data structures. So I propose to use Redis for application data caching.
+
+### Resolutions
+ - Model "Remote cache dedicated to each service" is most suitable.
```
buskamuza: How about a mixed approach? Remote cache makes sense for scaled services. If a service is not impacted with high load, the chances are high that it will have only one instance. Local cache may be more effective in this case.

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+   - the information stored within an individual cache node cannot be shared with other application servers
+   - not follow the 12-factor principles (any data that needs to persist must be stored in a stateful backing service)
+ 
+ ![Local cache](img/local-app-cache.png)
+ 
+#### Remote caches dedicated to each service
+ - Pros: 
+   - scalable, adheres to 12-factor principles
+   - externalize application state
+   - possibility dynamically change the number of application instances
```
buskamuza: "application instances" - I assume you mean "service instances"?

buskamuza:
On the diagrams, "Application" should be a "Service" I assume?
"Application" is entire Magento application consisting of all the services.

Could you please also add interaction between services to the diagrams and description? In the case of remote cache storage, do we have another reverse proxy so the service load reduced or not?

design-documents/service-isolation/caching-layer.md
viktym: Of course, we should make some performance measurements for our architecture.
But in most cases, the added network overhead associated with a remote cache is of little concern given that a request to it is generally fulfilled in sub-millisecond performance. So to unify cache model for all services I propose to use more advanced option that fit "12-factor app" principles - "Remote cache dedicated to each service"

viktym:
> On the diagrams, "Application" should be a "Service" I assume?
> "Application" is entire Magento application consisting of all the services.
On the diagrams, I treated the service as a standalone application, not in the context of whole Magento application. Actually, I mean PHP-FPM (can be scaled separately from web-server) or Apache/Nginx+PHP instance.
But I agree, better operate in terms of the statement that service is a part of Magento service-oriented application. Then let's clarify what service includes in our case.  If I understand correctly it is Apache/Nginx+PHP.
MySql, Elasticsearch, Redis, RabbitMQ, outbound email service are attached resources. Service consumes attached resources but not includes them. 

> Could you please also add interaction between services to the diagrams and description? In the case of remote cache storage, do we have another reverse proxy so the service load reduced or not?
The reverse proxy (Fastly) is situated between API-gateway and client in external network. Remote cache storage (like Redis) is in private network. In this document, I didn't consider the situation with reverse proxy between services inside the private network. Or what interaction do you mean?

design-documents/service-isolation/caching-layer.md
viktym: Yes, it will be more accurate. Fixed, thanks

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Overview
+
+Cache is one of the most important parts in a modern web application architecture.
+Magento caching layer should effectively improve:
+ * Performance
+ * Availability
+ * Scalability
+
+Caching should be implemented on several levels:
```
melnikovi: I would also add here caching responses of the services (by reverse proxies).

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+
+[BFF](https://github.com/magento/architecture/blob/master/design-documents/service-isolation.md#backends-for-frontends) http-responses contains static assets, **public** and **private** dynamic content.
+Static assets should be cached using CDN. Reverse proxy should cache public content.
+It could be Varnish (acts as reverse proxy) or Fastly (combines reverse proxy and CDN functions).
+Varnish and Fastly are already integrated with Magento. Reverse proxy uses lazy caching approach.
+Beside performance this type of cache has next benefits:
+ * Protection against outages - can optionally serve stale content when there is a problem with origin server
+ * Scalability - number of caching nodes can be increased
+ * Flexibility – Varnish Configuration Language (VCL) builds customized solutions, rules and modules
+ * Shielding - the shield node is the only node to get the content from origin server, other cache nodes will fetch from the shield node
```
melnikovi: Not sure this is the benefit, this is more of implementation detail.

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+
+A private or visitor-specific content should be stored on client-side (browser). After POST requests the cache 
+in the web browser will be flushed and an AJAX call will be done to fetch an updated copy of the private content.
+Also private content cache on client-side should expire according to TTL.
+
+ ![Private content caching](img/private-cache.jpg)
+ 
+ In general, the same caching approach works in current monolith Magento architecture.
+ For service oriented architecture next issues should be solved:
+  * The http-response header should include a set of tags for effective cache purging on reverse-proxy.
```
melnikovi: What would be an example of request performed by multiple services?

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+ 
+#### Remote caches shared across the services
+ - Pros: 
+   - centralized cache invalidation is easy
+   - scalable
+   - adheres to 12-factor principles
+ - Cons: 
+   - cache nodes contain mixed data from different services
+ 
+ ![Remote cache shared across the services](img/remote-cache-cluster.png)
```
melnikovi: Are Service 1 and Service 2 different services or these are instances of the same service? If first, this contradicts best practices. Please clarify.

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+ 
+Local caches are isolated on the individual nodes and cannot be leveraged within a cluster of application servers.
+Remote distributed caches on the other hand, provide low latency and higher levels of availability, when employed with read replicas and can provide a shared environment for all application servers to utilize the cached data.
+Remote cache dedicated to each service allows to store only data of particular service. 
+But there is no possibility (needs implementation) to flush all cache instances in case of upgrade or other huge changes.
+Remote caches shared across the services are easy to flush centralized, but mix data from different services.
+
+
+#### Cache invalidation
```
melnikovi: How can we invalidate cache for all services and BFF? Each service can expose API for cache invalidation, just like you suggest. But better approach would be is to have services and BFF listen to cache invalidation events. This way system would be less coupled.

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+ - In case of "Remote caches shared across the services" model full cache invalidation can be performed by any service
+
+#### Technologies 
+Today most key/value stores such as Memcached and Redis can store terabytes worth of data. Redis also provides high availability and persistence features. 
+Redis and Memcached offer high performance. Memcached is designed for simplicity while Redis offers a rich set of features that make it effective for a wide range of use cases.
+Unlike Memcached, Redis offers snapshots, replication, transactions, advanced data structures. So I propose to use Redis for application data caching.
+
+### Resolutions
+ - Model "Remote cache dedicated to each service" is most suitable.
+ - REST entry point on each service for cache invalidation should be created.
```
melnikovi: Have you consider invalidating cache by publishing events in the message queue? What are the benefits of using API?

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,105 @@
...
+ - Service mustn't store in cache data retrieved from other Magento service
+ - In case of "Remote caches dedicated to each service" model service should provide entry point for full cache invalidation
+ - In case of "Remote caches shared across the services" model full cache invalidation can be performed by any service
+
+#### Technologies 
+Today most key/value stores such as Memcached and Redis can store terabytes worth of data. Redis also provides high availability and persistence features. 
+Redis and Memcached offer high performance. Memcached is designed for simplicity while Redis offers a rich set of features that make it effective for a wide range of use cases.
+Unlike Memcached, Redis offers snapshots, replication, transactions, advanced data structures. So I propose to use Redis for application data caching.
+
+### Resolutions
```
melnikovi: Also need to describe BFF full page cache invalidation, I suggest to include in this proposal high level overview.

design-documents/service-isolation/caching-layer.md
viktym: > Are Service 1 and Service 2 different services or these are instances of the same service? If first, this contradicts best practices. Please clarify.

On this diagram Service 1 and Service 2 are different services. It is only one of the possible variants and we decided that it's not the best option.

> For application cache I would propose each service has it's own cache this cache should not be shared with other services. If there are multiple instances of the same service, they will use the same cache.

So actually this is model "Remote caches dedicated to each service" that we agreed to use. I'll add statement that  multiple instances of the same service should use the same cache instance.

 > I also propose to use reverse proxy to cache GET requests that don't contain sensitive data.

So you propose to set reverse proxy in front of each service? In that case is there sense of setting reverse proxy in front of BFF?

design-documents/service-isolation/caching-layer.md
viktym: As I remember Anton said that we should go away from service bus in interservice communication as it can be a single point of failure. So I didn't include this way in document

design-documents/service-isolation/caching-layer.md
viktym: If you talk about "Shielding" I agree. It's OOTB Fastly feature, I included since we using Fastly for our cloud solution. I'll remove it

design-documents/service-isolation/caching-layer.md
viktym: So it's a case when GET request after BFF goes to service1 and service2 performs a request to service2 for some data.
For example this one goes to Checkout service and than request data from Payment service:

` <route url="/V1/guest-carts/:cartId/payment-information" method="GET">
        <service class="Magento\Checkout\Api\GuestPaymentInformationManagementInterface" method="getPaymentInformation"/>
        <resources>
            <resource ref="anonymous" />
        </resources>
    </route>`

design-documents/service-isolation/caching-layer.md
viktym: Added

design-documents/service-isolation/caching-layer.md
viktym: Added proposal about considering cache invalidation via message queue

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,113 @@
...
+It could be Varnish (acts as reverse proxy) or Fastly (combines reverse proxy and CDN functions).
+Varnish and Fastly are already integrated with Magento. Reverse proxy uses lazy caching approach.
+Beside performance this type of cache has next benefits:
+ * Protection against outages - can optionally serve stale content when there is a problem with origin server
+ * Scalability - number of caching nodes can be increased
+ * Flexibility – Varnish Configuration Language (VCL) builds customized solutions, rules and modules
+
+ ![Public content caching](img/public-cache.jpg)
+
+A private or visitor-specific content should be stored on client-side (browser). After POST requests the cache
```
DrewML: > After POST requests the cache in the web browser will be flushed 

It's not super clear what is meant by this. Can you expand?

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,113 @@
...
+[BFF](https://github.com/magento/architecture/blob/master/design-documents/service-isolation.md#backends-for-frontends) http-responses contains static assets, **public** and **private** dynamic content.
+Static assets should be cached using CDN. Reverse proxy should cache public content.
+It could be Varnish (acts as reverse proxy) or Fastly (combines reverse proxy and CDN functions).
+Varnish and Fastly are already integrated with Magento. Reverse proxy uses lazy caching approach.
+Beside performance this type of cache has next benefits:
+ * Protection against outages - can optionally serve stale content when there is a problem with origin server
+ * Scalability - number of caching nodes can be increased
+ * Flexibility – Varnish Configuration Language (VCL) builds customized solutions, rules and modules
+
+ ![Public content caching](img/public-cache.jpg)
```
DrewML: Note that these diagrams show GraphQL sitting behind a cache/proxy. This is fine, but note that an HTTP cache will _not_ actually handle caching of GraphQL requests/responses, as they're all done as a `POST`

viktym: Thanks, added a note about it

design-documents/service-isolation/caching-layer.md
melnikovi: As I remember we agreed to use message queue for cache invalidation to avoid coupling between services. @akaplya can you confirm?

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,117 @@
...
+ - In case of "Remote caches shared across the services" model full cache invalidation can be performed by any service
+
+#### Technologies 
+Today most key/value stores such as Memcached and Redis can store terabytes worth of data. Redis also provides high availability and persistence features. 
+Redis and Memcached offer high performance. Memcached is designed for simplicity while Redis offers a rich set of features that make it effective for a wide range of use cases.
+Unlike Memcached, Redis offers snapshots, replication, transactions, advanced data structures. So I propose to use Redis for application data caching.
+
+### Resolutions
+ - Model "Remote cache dedicated to each service" is most suitable for application data cache.
+ - Don't use reverse proxy for caching requests in service-to-service communication until it's needed
```
melnikovi: I suggest to remove this from design document.

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,117 @@
...
+ - Pros: fast, no network traffic associated with retrieving data
+ - Cons: 
+   - shares sources with application
+   - can lead to data inconsistency when multiple instances of application servers are created
+   - the information stored within an individual cache node cannot be shared with other application servers
+   - not follow the 12-factor principles (any data that needs to persist must be stored in a stateful backing service)
+ 
+ ![Local cache](img/local-service-cache.png)
+ 
+#### Remote caches dedicated to each service
```
melnikovi: Could you specify, that this is what we agreed on?

design-documents/service-isolation/caching-layer.md
```diff
@@ -0,0 +1,117 @@
...
+  * Caching on reverse-proxy should be aligned with request authorization approach. The presence of an authorization token should be taken into account.
+
+### Caching results of GET-requests in service-to-service communication (private web-server)
+
+Services can produce a lot of GET requests to each other in the private network.
+First of all service application should be as fast as possible to avoid using reverse proxy as cache layer.
+But if application is not able to hold the load a reverse proxy can be used to reduce the overall load on service origin.
+Varnish is offered as a proven solution for such cases. Should be aligned with request authorization approach.
+
+ ![Private content caching](img/reverse-proxy-service1.png)
```
melnikovi: Need to remove Varnish from diagram.

design-documents/service-isolation/caching-layer.md
viktym: It means that we flushing the web browser cache every time a HTTP POST request is made

PR #54, opened 2018-12-12
Update general principles section

melnikovi:


design-documents/service-isolation.md
```diff
@@ -239,6 +239,9 @@ General principles to follow for service isolation:
 * All service operations MUST BE stateless
 * There MUST BE NO data dependencies between services.
 * Command & query responsibility segregation – storefront APIs for data immutable in storefront (catalog) should be optimized for data retrieval
+* Each request to service MUST contain correlation identifier to allow identify requests related to the same flow during debugging
```
buskamuza: The same is already mentioned as "Correlation IDs must be added to internal calls to make tracing easier.".

design-documents/service-isolation.md
```diff
@@ -239,6 +239,9 @@ General principles to follow for service isolation:
...
+* Service mesh MUST BE used for communication between services: routing, load balancing, circuit breaking, statistics
```
buskamuza: I wouldn't put it as a principle. It is a solution. It would be good if it goes to a separate document about services communication.

design-documents/service-isolation.md
```diff
@@ -239,6 +239,7 @@ General principles to follow for service isolation:
 * All service operations MUST BE stateless
 * There MUST BE NO data dependencies between services.
 * Command & query responsibility segregation – storefront APIs for data immutable in storefront (catalog) should be optimized for data retrieval
+* Services MUST NOT manage session, operations have to explicitly declare all required arguments, and those arguments should be passed in request
```
paliarush: There must be no knowledge about `Request` object in service either.

PR #57, opened 2018-12-20
Use default XML encoding

milindsingh:


buskamuza:
Approved. But keep in mind that it doesn't provide any commitment from Magento side to implement it, this depends on the product priorities. It looks like a good task for the community to contribute though.

PR #58, opened 2019-01-03
Change schema of Import sources to AsyncImport module

nuzil:
After several discussion we change input body of Import Source call and also changed Endpoint

### DIFF
--- a/design-documents/asynchronous-import/base-extension.md
+++ b/design-documents/asynchronous-import/base-extension.md
@@ -13,7 +13,7 @@ With phase 1 we are planning to develop next functionality:
 ## File Upload Endpoint
 ### Main endpoint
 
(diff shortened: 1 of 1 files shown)


design-documents/asynchronous-import/base-extension.md
```diff
@@ -24,13 +24,10 @@ This request can accept files from different sources:
 
 ```
 {
-  "fileEntry": {
-    "file_id": 0,
-    "source": {
+  "source": {
       "import_data": "/var/www/html/bulk-api/async-import/var/catalog_product.csv",
```
kandy: We shouldn't support importing from the outside of  Magento root

nuzil: Thanks, its true

PR #64, opened 2019-01-11
Framework components as first-class citizens

antonkril:
## Problem

Magento framework components are parts for `magento/framework` package, and shipped with this package. There is no way for a lightweight application to only specify components that it will use.

Also framework components do not have ability to specify magento application configuration. This leads to application framework configuration to be stored in `app/etc` and modules (mostly `di.xml`). `app/etc/di.xml` is more than 1700 lines long.

## Solution

To provide the immediate solution to the 2 problems mentioned above, we need to split application framework to separate composer packages, and allow framework components to provide application configuration in `/etc` folder. This will allow to move all declarations from `app/etc/di.xml` to corresponding framework components.

## Open questions

1. Should there be a limitation of what types of configuration can be provided by framework components? Current answer - no, there should be no limitation for simplicity.

## POC
POC is implemented in https://github.com/magento-architects/magento2/tree/split-framework.

Use https://github.com/magento-architects/magento-project for easy installation of a "hello world" application.

### DIFF
--- a/design-documents/components.md
+++ b/design-documents/components.md
@@ -2,7 +2,7 @@
 
 1. Every framework component is a separate composer package with type `magento2-library` and name prefix `magento2/library-`
 2. Every `\Magento\Framework\App` component is a separate package.
-2. For backward compatibility in GitHub repo classes that used to live in `Magento\Framework` and `Magento\Framework\App` namespaces are moved to root folders of corresponding components, while all other classes of those components are moved into folder with the name of component.
+2. Classes that used to live in `Magento\Framework` and `Magento\Framework\App` namespaces are moved to root folders of corresponding components, while all other classes of those components are moved into folder with the name of component. Root namespace for such components is configured to be `\Magento\Framework` and `\Magento\Framework\App` to avoid renaming classes.
 
 *Before*
 ```
--- a/design-documents/components.md
+++ b/design-documents/components.md
@@ -1,0 +1,29 @@
+# Framework components
+
+1. Every component is a separate composer package
+2. Every `\Magento\Framework\App` component is a separate package.
+2. For backward compatibility in GitHub repo classes that used to live in `Magento\Framework` and `Magento\Framework\App` namespaces are moved to root folders of corresponding components, while all other classes of those components are moved into folder with the name of component.
+
+*Before*
+```
+Magento/
+  Framework/
+    App/
+      Action/
+        AbstractAction.php
+        ...
+      ActionInterface.php
+```
+*Now*
+```
+Magento/
+  Framework/
+    App/
+      Action/
+        Action/
+          AbstractAction.php
+        ActionInterface.php
+        composer.json
+```
+
+3. Every framework component can specify `/etc` directory where application configuration can be provided.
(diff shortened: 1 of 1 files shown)


antonkril:
I like the `/src` folder idea, but that is different from all other Magento modules. Also, how would we treat classes that belong to a component but sit outside of its folder (like ActionInterface)?

davidverholen:
ok I see, this would require namespace changes. But maybe, it could work by introducing additonal bc components, that extend these classes and reside in the original namespace. 

Since they are doing nothing but extending the original class / interface, they should be compatible with any future magento version as long as the moved class is not removed / moved again.



**magento/magento2-library-app-action**
composer type: magento2-library
psr4-prefix `\Magento\Framework\App\Action\`

`/src/ActionInterface.php` contains the Interface which means it changed the namespace resulting in `\Magento\Framework\App\Action\ActionInterface`



**magento/magento2-library-app-action-bc**
composer type: library
psr4 prefix:`\Magento\Framework\App\`

`/src/ActionInterface.php` contains new interface in old namespace that extends from `\Magento\Framework\App\Action\ActionInterface`



these modules could just be left out as soon as magento does not want to support the old namespaces anymore. And even then, integrators could install these themselves if they still need bc for a component.

This would mean introducing many more new composer packages. But in the long run you would end up with a cleaner directory structure for the framework. 

Not sure if I'm missing other bc implications maybe

navarr:
I think _moving_ to src is a good idea.  I've seen before that Magento typically recommends a structure like:

```
Vendor_Module/<files>
Vendor_Module2/<files>
```

If remaining with a monolithic approach, and the goal is to get everything into Magento components of some sort (as opposed to living in the project, for example), this seems like the direction you'd be headed in anyway.

Regarding things like ActionInterface, is there a large con to just having say..

`Magento_LibraryAppAction/` be PSR-4 mapped to `\Magento\Framework\App`, and then inside it you have `ActionInterface` and an `Action\` folder?

antonkril:
@navarr i don't disagree with `/src`. It is a good topic but separate from this. 

Regarding the PSR-4 mapping - this is effectively what we are doing.

joni-jones:
Not sure about Framework directory structure in POC, as an example:
```
Image
  - etc
     di.xml
  - Image
    -- Adapter
      AbstractAdapter.php
    -- Test
      -- Unit
         AdapterFactoryTest.php
    AdapterFactory.php
    Factory.php
    Readme.md
  composer.json
```

Maybe it makes sense to move `Test` directory on the same level as source code?

melnikovi:
@joni-jones isn't test directory already on the level of source code (see [example](https://github.com/magento-architects/magento2/tree/split-framework/lib/internal/Magento/Framework/Encryption/Encryption))? Or what do you mean?

design-documents/components.md
```diff
@@ -0,0 +1,29 @@
+# Framework components
```
melnikovi: Could you describe structure for new framework components.

joni-jones:
> @joni-jones isn't test directory already on the level of source code (see [example](https://github.com/magento-architects/magento2/tree/split-framework/lib/internal/Magento/Framework/Encryption/Encryption))? Or what do you mean?

Actually, you attached a link to the module with nested structure:
```
Encryption
    - Encryption
       - Adapter
       - Helper
       - Test
           - Unit
    composer.json
```
The `Test` directory is a child directory of source code.

PR #71, opened 2019-01-24
Standard Promises

AlexMaxHorkun:
Standard promises to be used with asynchronous operations in Magento. Operations like calling a remote service or just sending an HTTP request asynchronously.


melnikovi:

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=71) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=71) before we can accept your contribution.<br/>

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
...
+          $anotherPromise->wait();
+      }
+  }
+  ```
+
+### Using promises for existing code
+We have a standard HTTP client - Magento\Framework\HTTP\ClientInterface, it can benefit from allowing async requests
+functionality for developers to use by employing promises.
+ 
+This client is being used in Magento\Shipping\Model\Carrier\AbstractCarrierOnline to create package shipments/
```
joni-jones: Could you make a POC to see how API will be applicable and to find potential gaps before introducing this API?

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
...
+$result = null;
+$doStuffOperation->then(function ($response) use (&$result) { $result = $response; });
+$simultaneousOperation = $this->otherObject->processStuff();
+$doStuffOperation->wait();
+$updated = null;
+$saveOperation = $this->repo->save($result);
+$saveOperation->then(function ($result) use (&$updated) { $updated = $result; });
+
+//Waiting for all
+$saveOperation->wait();
```
joni-jones: Should we call `wait` for both async and sync approaches?

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
+### Why?
+Service contracts in the future will often be executed in an asynchronous manner
```
melnikovi: Please elaborate more on the purpose and provide examples where it can be used now. As I understand, the purpose is to allow call APIs in parallel when we need to call multiple APIs consequentially and as a result improve performance?

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
...
+{
+    public function resolve($value): void;
+    
+    public function reject(\Throwable $exception): void;
+}
+```
+
+### Implementation
+A wrapper around [Guzzle Promises](https://github.com/guzzle/promises) will be created to implement the APIs above. Guzzle Promises fit
+most important criteria - they allow synchronous execution as well as asynchronous. It's a mature
```
melnikovi: What are the benefits of the interface that would allow us to call something both ways?

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
...
+          $anotherPromise = $this->runner->run(SomeServiceInterface::class, 'doStuff', []);
+          //Waiting for both results
+          $promise->wait();
+          $anotherPromise->wait();
+      }
+  }
+  ```
+
+### Using promises for existing code
+We have a standard HTTP client - Magento\Framework\HTTP\ClientInterface, it can benefit from allowing async requests
```
melnikovi: Does it mean that we will have to change the interface? This might be backwards incompatible.

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
...
+* Async web API uses status requests to check whether operations are completed which is alright for large numbers
+  of operations but for a small number (like 2, 3) it will just generate more requests than just sending 1 request
+  for each operation
+
+So to allow execution of multiple service contracts from different domains it's best to send 1 request per operation
+and to let client code to chain, pass and properly receive promises of results of operations.
+ 
+#### How will it look?
+There are to ways we can go about using promises for asynchronous execution of service contracts:
+* Service interfaces themselves returning promises for client code to use
```
melnikovi: How will this work in the existing system? In web API for instance?

design-documents/promises.md
```diff
@@ -0,0 +1,240 @@
...
+  interface AsynchronousRunnerInterface
+  {
+      public function run(string $serviceName, string $serviceMethod, array $arguments): PromiseInterface;
+  }
+  ```
+  _regular service_:
+  ```php
+  interface SomeRepositoryInterface
+  {
+    public function save(DTOInterface $dto): void;
```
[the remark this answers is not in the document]
AlexMaxHorkun: They can, that's just an example meant to show that 2nd approach does not require a particular return type

design-documents/promises.md
AlexMaxHorkun: you would call 'wait' to wait for an operation to complete if it's required, you can just provide a callback to 'then' method if you don't need waiting

design-documents/promises.md
AlexMaxHorkun: public API methods don't have to return promise since we'll be calling them via HTTP. If you mean that when a service contract implementation will actually be calling web API on another server then yes, if we go with the 1st route they'd have to return a promise

design-documents/promises.md
AlexMaxHorkun: if a promise would only allow 'then' and 'otherwise' callback then all the client code using would have to be returning promises as well because there would be no way to actually wait for an async operation to finish

design-documents/promises.md
AlexMaxHorkun: we would have to add methods or add an interface that will extend the existing one

design-documents/promises.md
AlexMaxHorkun: web API controllers would just have to updated to expect promises, calling wait on them and then rendering results

AlexMaxHorkun:
Decided that futures serve us better

joni-jones:
@melnikovi , @akaplya do you have any comments/objections?

design-documents/promises.md
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @return bool
+     */
+    public function isDone(): bool;
+}
+```
+
+Advanced interface that allows canceling asynchronous operations that have been started.
+```php
+interface CancelableDeferredInterface extends DeferredInterface
```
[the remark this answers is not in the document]
AlexMaxHorkun: added the explanation to the document

PR #85, opened 2019-02-06
Proposal: Allow class attribute for referenceBlock type in layout xml

sivaschenko:
# Allow class attribute for referenceBlock type in layout xml

## Problem

It is **impossible to substitute a block class** and/or customize block method(s) implementation **on a specific page keeping layout configuration of the block** (position, sorting in the layout, child blocks, arguments).

Currently available block class substitution approaches do not resolve the problem:

| Currently available approach                                                      | Why the approach does not resolve the problem  |
| --------------------------------------------------------------------------------- |:----------------------------------------------:|
| Introduce a preference for block class in di.xml                                  | Block is substituted globally, not only on the desired page |
| Remove the existing block from the layout and declare a new one                   | Layout configuration related to this block is lost |
| Use plugins                                                                       | Block functionality is affected for all instances, not only on the desired page |
| Use plugins that change method functionality only for specific page/layout handle | This approach is imperative, implicit, prone to mistakes, bugs and implementation inconsistency, it is not compatible with layout updates | 

## Solution

Allow the `class` attribute of `blockReferenceType` in `lib/internal/Magento/Framework/View/Layout/etc/elements.xsd`

See: https://github.com/magento/magento2/pull/18933

### Pros

* Possible to substitute a block class for a specific page easily and in declarative way preserving layout configuration of the block

### Cons

* Allows developer to substitute a block class with a class that is not compatible with a template

## POC

POC is implemented in public pull request: https://github.com/magento/magento2/pull/18933

## Requested Reviewers

@leandrommedeiros - contributor
@antonkril 
@maghamed 
@akaplya 
@kandy

### DIFF
--- a/design-documents/class-attibute-for-referenceBlock-type-in-layout-xml.md
+++ b/design-documents/class-attibute-for-referenceBlock-type-in-layout-xml.md
@@ -1,0 +1,32 @@
+# Allow class attribute for referenceBlock type in layout xml
+
+## Problem
+
+It is **impossible to substitute a block class** and/or customize block method(s) implementation **on a specific page keeping layout configuration of the block** (position, sorting in the layout, child blocks, arguments).
+
+Currently available block class substitution approaches do not resolve the problem:
+
+| Currently available approach                                                      | Why the approach does not resolve the problem  |
+| --------------------------------------------------------------------------------- |:----------------------------------------------:|
+| Introduce a preference for block class in di.xml                                  | Block is substituted globally, not only on the desired page |
+| Remove the existing block from the layout and declare a new one                   | Layout configuration related to this block is lost |
+| Use plugins                                                                       | Block functionality is affected for all instances, not only on the desired page |
+| Use plugins that change method functionality only for specific page/layout handle | This approach is imperative, implicit, prone to mistakes, bugs and implementation inconsistency, it is not compatible with layout updates | 
+
+## Solution
+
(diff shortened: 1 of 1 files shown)


maghamed:
> Allows developer to substitute a block class with a class that is not compatible with a template

This is called Duck Typing, as the contract of a block is not defined in advance it's defining by the template/child blocks which use it

sivaschenko:
@maghamed It is possible to implement structural typing for this proposal by restricting the class defined in `class` attribute of `referenceBlock` to be instance or child of the original block class defined in `block` node.

orlangur:
@maghamed,

> Structural typing is preferable in OOP in comparison with Duck Typing which this proposal incorporates

Absolutely not applicable here as we allow complete removal of a particular node and re-adding it back.

This proposal is just a syntactic sugar to avoid some something as ugly as https://github.com/magento/magento2/pull/11709/files

Vinai:
I think this would be a nice addition. This gives layout XML more leverage, that is, once learned, more can be achieved with the tool.

johnhughes1984:
Agreed this would be a nice addition. 

Also I don’t believe the ‘con’:

> Allows developer to substitute a block class with a class that is not compatible with a template

Is any different from a developer currently being able to replace the existing template with one that is not compatible with the block class or that you can add / replace / remove arguments in layout XML that could equally break something.

PR #91, opened 2019-02-19
MC-13864: Consumer always read config from memory

vzabaznov:
## Problem

Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to wrong im-memory config/objects states and we need to introduce the way to reinit them
## Solution

Introduce poison pill - some message stored in Data Storage that signals that consumer must be reinited

## Requested Reviewers

@kandy @paliarush @vrann

### DIFF
--- a/design-documents/asynchronous-operations/poison-pill.md
+++ b/design-documents/asynchronous-operations/poison-pill.md
@@ -5,7 +5,7 @@ poison pill
 
 ### Overview
 
-Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to wrong im-memory config/objects states and we need to introduce the way to reinit them.
+Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to outdated im-memory config/objects states and we need to introduce the way to reinit them.
 
 ### Design
 
--- a/design-documents/asynchronous-operations/poison-pill.md
+++ b/design-documents/asynchronous-operations/poison-pill.md
@@ -1,20 +1,36 @@
 ## Glossary
 
 poison pill
-: message in Data Storage that signals that consumer should die
+: message that added after list of events in Data Storage that signals that consumer should die.
 
 ### Overview
 
-Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to outdated im-memory config/objects states and we need to introduce the way to reinit them.
+Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it could happens that our consumer keep outdated im-memory config/objects states. Considering that, we need to introduce the way to reinit them.
 
 ### Design
 
-Before process new message consumer should ask Data Storage if there are poison pill for him. If yes, consumer dies and next cron:run will start new instance of consumer.
+Basic implementation of how the pill will be added should be implemented inside Magento Framework. We will share the api that let each module call whenever it needed.
+
+Api Interface - *PoisonPillInterface* will be inside Magento Framework, implementation in MessageQueue module.
+
+PoisonPillInterface
+```php
+/**
+ * Put poison pill inside DB.
+ *
+ * @throws \Exception
+ * @return int
+ */
+public function put();
+```
+
+Before process new message, consumer will compare version of PoisonPill in-memory state with latest in DB. If version is different consumer dies and next cron:run will start new instance of consumer.
 
 #### Acceptance Criteria Fulfillment
 
-1. If consumer takes message to process after poison pill was placed, consumer must be reinited
+1. If consumer takes message to process after poison pill was placed, consumer must be reinited.
 
 #### Extension Points and Scenarios
 
-We will put poison pill into Data Storage based on common Magento extension points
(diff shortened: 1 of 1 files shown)


design-documents/asynchronous-operations/poison-pill.md
```diff
@@ -0,0 +1,20 @@
...
+
+Before process new message consumer should ask Data Storage if there are poison pill for him. If yes, consumer dies and next cron:run will start new instance of consumer.
+
+#### Acceptance Criteria Fulfillment
+
+1. If consumer takes message to process after poison pill was placed, consumer must be reinited
+
+#### Extension Points and Scenarios
+
+We will put poison pill into Data Storage based on common Magento extension points like events and plugins
```
joni-jones: As our goal is modules decoupling, we should not add new dependencies between modules via plugins or observers.

design-documents/asynchronous-operations/poison-pill.md
```diff
@@ -0,0 +1,20 @@
+## Glossary
+
+poison pill
+: message in Data Storage that signals that consumer should die
```
joni-jones: What kind of consumer do you mean? A message queue, cron?

design-documents/asynchronous-operations/poison-pill.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Overview
+
+Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to wrong im-memory config/objects states and we need to introduce the way to reinit them.
+
+### Design
+
+Before process new message consumer should ask Data Storage if there are poison pill for him. If yes, consumer dies and next cron:run will start new instance of consumer.
```
joni-jones: How a consumer will communicate with data storage?

design-documents/asynchronous-operations/poison-pill.md
vzabaznov: Basically, we are not going to use in-process communication

design-documents/asynchronous-operations/poison-pill.md
vzabaznov: I think yeas, we are going to introduce the basic implementation inside our framework and other modules will use it

design-documents/asynchronous-operations/poison-pill.md
```diff
@@ -0,0 +1,20 @@
+## Glossary
+
+poison pill
+: message in Data Storage that signals that consumer should die
+
+### Overview
+
+Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to wrong im-memory config/objects states and we need to introduce the way to reinit them.
```
buskamuza: Could you clarify what "wrong" means? Is it a state modified by the processes, so it is changed?

vzabaznov: wrongs means that if someone will change the state, consumer will running with outdated in cache, yeah outdated is more accurate meaning right here

design-documents/asynchronous-operations/poison-pill.md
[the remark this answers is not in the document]
vzabaznov: The message will be added after some event is triggered, e.g. website was created

design-documents/asynchronous-operations/poison-pill.md
joni-jones: Could you add more details about API which should be introduced?

design-documents/asynchronous-operations/poison-pill.md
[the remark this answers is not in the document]
vzabaznov: Adding variable on cloud triggers redeploy, which will restart consumers by default

design-documents/asynchronous-operations/poison-pill.md
```diff
@@ -0,0 +1,20 @@
...
+poison pill
+: message in Data Storage that signals that consumer should die
+
+### Overview
+
+Right now our consumer depends on config/objects in-memory state. After application was bootstrapped it initialize a lot of object that are always kept in memory after first load - config, websites, etc.., since our consumers are long-living and their live time limited only by message count it leads to outdated im-memory config/objects states and we need to introduce the way to reinit them.
+
+### Design
+
+Before process new message consumer should ask Data Storage if there are poison pill for him. If yes, consumer dies and next cron:run will start new instance of consumer.
```
igrybkov: If there is a possibility that several consumers would be started simultaneously, how are you going to deliver your poison pill for each of them?

vzabaznov: Each of them will compare version of poison pill in his memory vs latest in DB

design-documents/asynchronous-operations/poison-pill.md
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Put poison pill inside DB.
+ *
+ * @throws \Exception
+ * @return int
+ */
+public function put();
+```
+
+Before process new message, consumer will compare version of PoisonPill in-memory state with latest in DB. If version is different consumer dies and next cron:run will start new instance of consumer.
```
paliarush: Does it mean that there will be at least one MySQL DB query for each consumed message?

vzabaznov: Yes

paliarush: Please add warning to the design document for the team implementing this feature to verify performance implications.

tariqjawed83: @paliarush @vzabaznov I am concerned about this point specifically, if for each read we are hitting the database query, then what is the purpose of storing configurations in-memory? we might lose the benefits of in-memory optimizations. We should think differently here, on invalidating the cache on event based.

PR #93, opened 2019-02-20
Update service decomposition guidelines

melnikovi:
- working with data
- deployment of services


joni-jones:
Not sure that services should dispatch any events, like for reporting purposes. It's a responsibility of reporting service to obtain all needed data from services.

melnikovi:
Reporting services will not know about updates to the data. They will have to constantly pull many services which generates load on them.

PR #96, opened 2019-02-25
Services orchestration

akaplya:
## Solution
This proposal describes how to orchestrate communication between isolated services. And the way how to move from the current state to the desired state.


design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+
+```
+Observer can not return values into `outputPath` or start a new execution branch.
+As a result, observer execution can be delayed or even processed async in background.
+Observer can be declared `before` of `after` the step.
+
+![Figure 2](services-orchestration/events.png)
+
+## Sync and Async adapters
+Ideally, all scenarios should not be run synchronously.
```
melnikovi: I think some scenarios can be only synchronous. And some scenarios (add to cart for instance) would be best to keep synchronous. What store front scenarios except modifying when totals being calculated and placing order would you make asynchronous?

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+[Configuration schema](https://github.com/akaplya/mage-state-machine/blob/master/lib/internal/Magento/Framework/StateMachine/etc/routines.xsd)
+
+## Backward compatibility
+
+Such a mechanism can be useful with the evolution of an existing application.
+Major scenarios can be described with such language.
+For monolithic app the scenario rewrote such way will be backward compatible because scenario will run the same operation in the monolith execution context.
+### Real life example
+
+Implementation of [\Magento\Sales\Api\OrderManagementInterface::place](https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/Sales/Model/Service/OrderService.php#L196)
```
melnikovi: Where this configuration would live?

design-documents/services-orchestration.md
akaplya: Scenario declaration does not depend on the manner how the scenario could be executed. Sync or async mode will be chosen by a corresponding scenario invoker (adapter).

design-documents/services-orchestration.md
akaplya: Modules and configuration decomposition is aligned with the base document. So the main scenario should be declared in the scope of the corresponding module group. For instance `place order` scenario will be declared in SalesWorkflow. Payment extension will come from SalesPaymentWorkflow module.

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+![Figure 1](services-orchestration/so1.png)
+
+The main question, which must be addressed prior to many others is who will be responsible for organizing such communication? As soon as one service is starting to call other services through the network it becomes responsible for handling responses, managing timeouts, retries etc. As the result, increases overall complexity. Since this moment this service becomes responsible for handling network communication and receives knowledge about of called service like knowledge about foreign service signature. As a result, the service becomes more fragile. Changes that were made at another end may lead to original service invalidation.
```
melnikovi: What do you mean here by service signature and service invalidation?

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+
+Another drawback which should be considered, when one service encapsulates calls to another service such situation may affect the service replaceability. For instance, we have service `AddProductToCart` which encapsulate call to service `GetProductCustomizableOptions`. We can imagine a situation when a developer who knows about service `AddProductToCart` implementation details may build an extension with an assumption that `GetProductCustomizableOptions` will be always called inside of `AddProductToCart`. Replacing the service `AddProductToCart` by service `MyAddProductToCart` which is doing the same but does not call `GetProductCustomizableOptions` may unintentionally break logic of such extension.
+
+Possible "accidental" performance degradation also is a point of consideration. In case we make it possible and convenient to call one service from another service we could loose track of cases when it happens. Network call must not be convenient to place it anywhere in the code (For instance, in a plug-in of a frequently called method).
```
melnikovi: I think this could be an issue in both cases, when services know about each other and when communication between services orchestrated. I would propose to remove this statement.

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+
+Possible "accidental" performance degradation also is a point of consideration. In case we make it possible and convenient to call one service from another service we could loose track of cases when it happens. Network call must not be convenient to place it anywhere in the code (For instance, in a plug-in of a frequently called method). 
+
+
+Without proper testing, such code may cause performance degradation that could be found after deploying on production. 
+
+Also, debugging the application that may invoke remote services which also may invoke remote services with unlimited nesting is challenging(too much).
```
melnikovi: Add whitespace after challenging.

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+## Routines
+Routines configuration specifies a callable resource to be used in scenario steps.
+
+Actually we have to support two base types of routines:
+* Direct PHP Calls.
+* Magento WebAPI  REST calls.
+
+### Routines declaration example
+```
+    <routine xsi:type="MagentoWebAPIRoutine"
```
melnikovi: We may need to separate url param into host, resource or endpoint and uri.

design-documents/services-orchestration.md
melnikovi: Will this module be deployed on a orchestration service?

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+              routine="Magento:Sales:OrderManagement:saveOrder"
+        />
+        <step name="OrderSuccessfullyPlaced" xsi:type="Success" outputPath="params/order"/>
+    </scenario>
+```
+For monolith installation this scenario will exactly the same methods with the same arguments as it is now.
+In the same time, such declaration provides new ways how to evolve such service.
+
+## Extensibility
+All steps are declared as XML node.
```
melnikovi: I'm not sure we would be able to refactor all existing scenarios to follow this workflow.

design-documents/services-orchestration.md
```diff
@@ -0,0 +1,254 @@
...
+    * Changing step, route to a function with an interface that is different from core one:
+```xml
+<step xsi:type="Task" name="PlacePaymentForOrder" next="PersistOrder"
+      inputPath="params/order"
+      outputPath="params/order"
+      routine="Magento:MySales:OrderManagement:placeOrder"
+/>
+```
+
+*
```
melnikovi: Doublecheck formatting here.

melnikovi:

PR #106, opened 2019-02-28
Proposal for components decoupling

joni-jones:
Proposal for components decoupling


melnikovi:

PR #108, opened 2019-03-01
Eliminate GraphQL config elements instantiation on the fly

akaplya:
Eliminate GraphQL config elements instantiation on the fly

### DIFF
--- a/design-documents/graph-ql/framework/eliminate-config-elements-instantiation-on-the-fly.md
+++ b/design-documents/graph-ql/framework/eliminate-config-elements-instantiation-on-the-fly.md
@@ -27,7 +27,7 @@ To achieve this we can introduce specific serializer for config data. `Magento\F
         </arguments>
     </virtualType>
 ```
-In a fact, Magento knows all config data object.
+In a fact, Magento knows all config data objects.
 Magento does not use external resources as a configuration source.
 So we can create a lightweight serializer which may utilize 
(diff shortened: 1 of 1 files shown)


larsroettig:
@akaplyaI have a pr for a similar thing https://github.com/magento/graphql-ce/pull/314

paliarush:
Proposal looks good, please test performance to make sure this change gives us any improvement.

design-documents/graph-ql/framework/eliminate-config-elements-instantiation-on-the-fly.md
```diff
@@ -0,0 +1,36 @@
...
+            <argument name="serializer" xsi:type="string">Magento\Framework\GraphQl\Config\Data\Serializer</argument>
+        </arguments>
+    </virtualType>
+```
+In a fact, Magento knows all config data objects.
+Magento does not use external resources as a configuration source.
+So we can create a lightweight serializer which may utilize 
+[serialize](https://secure.php.net/manual/en/function.serialize.php)
+/*
+[unserialize](https://secure.php.net/manual/en/function.unserialize.php) functions with all known GraphQL configuration data objects mentioned in *`allowed_classes` option.
```
buskamuza: We can't use `unserialize` from security perspective.

paliarush:
Closing due to high risk/reward ratio. Proposed approach can be reconsidered once again in the future.

PR #109, opened 2019-03-05
Tech upgrade strategy

joni-jones:
## Problem

Magento has dependencies to 3rd party components and during each release, we need to spend a lot of time to update all components-dependable code. Before release, we need to check all dependencies manually and find the latest compatible versions.

## Solution

Provide a tool, which can find the latest version of 3rd party components.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=109) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=109) before we can accept your contribution.<br/>

design-documents/tracking-tool.md
```diff
@@ -0,0 +1,80 @@
...
+### Packages source 
+
+The most of packages are specified in `composer.json` but some of them are not specified, like database, or can be added in the future not via composer.
+So we need to provide a possibility to extend source of packages versions.
+
+The tool should support multiple input types like CLI, configuration files, URL to composer.json file.
+
+#### Github REST API
+
+Unfortunately, [Packagist API](https://packagist.org/apidoc) does not provide a possibility to retrieve the package latest version but, as the of packages are hosted on Github, [Github REST API](https://developer.github.com/v3/repos/releases/) might be used for such purposes.
```
igrybkov: “as the of packages” - typo here

joni-jones: Thanks, fixed.

design-documents/tracking-tool.md
```diff
@@ -0,0 +1,80 @@
...
+### Packages source 
+
+The most of packages are specified in `composer.json` but some of them are not specified, like database, or can be added in the future not via composer.
+So we need to provide a possibility to extend source of packages versions.
+
+The tool should support multiple input types like CLI, configuration files, URL to composer.json file.
+
+#### Github REST API
+
+Unfortunately, [Packagist API](https://packagist.org/apidoc) does not provide a possibility to retrieve the package latest version but, as the most of packages are hosted on Github, [Github REST API](https://developer.github.com/v3/repos/releases/) might be used for such purposes.
```
igrybkov: I see on your link to Packagist API that it allows to get package metadata which contains all available package versions: https://packagist.org/apidoc#get-package-data
Seems like it’s what you’re looking for as it could be parsed and compared against defined update rules like `^`

igrybkov: And they have field called “version_normalized” which, as I understand, allows you to not parse all those prefixes on your own

joni-jones: Not exactly, because you should have a possibility to find different versions like the latest patch, minor, major.

joni-jones: > And they have field called “version_normalized” which, as I understand, allows you to not parse all those prefixes on your own

I have seen it and probably it can be used but I didn't find the information why it contains 4 digits.

igrybkov: Yes, I just wanted to point out that this line is not completely correct. And in my opinion it’s better to stay close to the way how composer gets its info as it’s how we’d finally get the dependency. It could be published on GitHub, but not available on composer for example - not sure ho composer would be looking for dependency in such case.
Also, probably we could avoid implementing and supporting some things if we’d be close to composer - it already has everything we want, probably not always behind the right interface.
Here’s an example of the composer plugin I’m using which doing such checks: https://github.com/soullivaneuh/composer-versions-check

joni-jones: > A package could be published on GitHub, but not available on composer for example - not sure ho composer would be looking for dependency in such case.

Mysql it's a good example, we don't have dependency in `composer.json` but using it. This proposal about all dependencies which exist not only in `composer.json`.

> composer-versions-check is a plugin for Composer.
It warns user for outdated packages from last major versions after update command.

We are interested not only in major versions. Not always we can use the latest version, as an example some zend components but anyway the have to use the latest compatible version.

DrewML:
This write-up doesn't mention any prior art on the subject. I have to assume that, with PHP being one of the most popular programming languages on the web, we're not the first to attempt a solve for this problem.

I think it would be valuable for the document to identify how other large PHP frameworks are handling this.

design-documents/tracking-tool.md
```diff
@@ -0,0 +1,83 @@
...
+ - `/repos/:ownwer/:repo/releases` - get a list of all published releases
+ - `/repos/:owner/:repo/latest` - get the latest published release
+ - `/repos/:owner/:repo/releases/:id` - get release details
+ - `/repos/:owner/:repo/tags` - get all repository tags
+ 
+The `/repos/:owner/:repo/latest` allows getting all needed details but the problem that not all, event popular, organizations publish their releases (as example, https://github.com/symfony/console/releases, https://github.com/php/php-src/releases) which are available on Github but not via REST API.
+
+The `/repos/:owner/:repo/tags` allows retrieving all tags (in the most cases, tags and releases the same) but the tags will be returned unsorted and some cases a repository contains not only versions tags.
+
+WEB crawlers can be used as a possible solution for repositories which do not publish releases as Github shows sorted by date releases based on tags list. The tool can try using Github REST API, if result is an empty array, it will use WEB crawler to parse the releases page.
```
DrewML: Tools that scrape the DOM of a site are always going to be incredibly flaky. I highly recommend avoiding scraping pages. The Github DOM has no versioning guarantees and can break at any time.

design-documents/tracking-tool.md
```diff
@@ -0,0 +1,83 @@
...
+## Overview
+
+Magento uses different 3rd party components and libraries like Symfony, Zend, hash generators, SDK, etc. And during Magento release we need to update components versions manually. Also, updates to the latest package version might require some code migration in Magento, which might impact release dates.
+To avoid such problem, we need to have a mechanism to track latest versions of 3rd party components.
+
+As most of dependencies specified in `composer.json`, such process can be automated.
+
+## Solution
+
+Dependencies, specified in `composer.json`, can be parsed and according to the dependencies list, we can find the latest version of the package.
```
DrewML: > Dependencies, specified in `composer.json`, can be parsed and according to the dependencies list, we can find the latest version of the package.

This seems to not account for the content under "Packages source" below:

> The most of packages are specified in `composer.json` but some of them are not specified, like database, or can be added in the future not via composer.
So we need to provide a possibility to extend source of packages versions.

> The tool should support multiple input types like CLI, configuration files, URL to composer.json file.

joni-jones: Why? This section exactly explains why we need to support multiple sources and it's a child section of the "Solution" section

design-documents/tracking-tool.md
joni-jones: You are right but as an alternative solution, it has the right to exist as there is no other API to get such details.

joni-jones:
> This write-up doesn't mention any prior art on the subject. I have to assume that, with PHP being one of the most popular programming languages on the web, we're not the first to attempt a solve for this problem.

I've found multiple solutions, most of them are outdated and don't maintain anymore. Some of them don't satisfy our requirements, like mentioned https://github.com/soullivaneuh/composer-versions-check. Some of them have only notifications about the latest versions, like https://coderelease.io/ or https://libraries.io. But they aggregate data from the same source - packagist, github, etc.

> I think it would be valuable for the document to identify how other large PHP frameworks are handling this. If we're going to write and maintain our own tool, it's worth being clear about why we couldn't use an off the shelf solution.

I didn't find information about other popular frameworks, maybe because their `composer.json`don't contain so many dependencies to 3rd party components like in Magento because they are frameworks not a platform like Magento.

DrewML:
> I've found multiple solutions, most of them are outdated and don't maintain anymore. Some of them don't satisfy our requirements, like mentioned soullivaneuh/composer-versions-check. Some of them have only notifications about the latest versions, like coderelease.io or libraries.io. The newreleases.io - looks suitable but it still in beta and doesn't have a documentation yet.

A super quick Google search brings up more options than that.

- `composer outdated` command (manual, but respects semver)
- https://violinist.io/ (has a Github bot integration)
- https://renovatebot.com/ (has a Github bot integration, supports more than just composer, allows custom plugins)
- https://dependabot.com/php/ (can handle composer or github sources)

I think we should really strive to try and find an off-the-shelf solution (or at least an existing one we can contribute to) before we start hand-rolling more of our own tools, as maintenance of stuff like this has a real cost.

joni-jones:
Have you checked the proposed solutions according to specified requirements?

I agree that it always better using some existing tool and don't reinvent the wheels but the solution should be suitable for our requirements.

DrewML:
> Have you checked the proposed solutions according to specified requirements?

Oh I'm not proposing them - I'm highlighting that a proposal for Magento authoring a new tool should (imo) include why off-the-shelf solutions were not viable, justifying the time we'd have to sink into building something from scratch.

joni-jones:
@DrewML will do more investigation and add a section for the existing solutions.

buskamuza:
For composer dependencies we can also use https://getcomposer.org/doc/03-cli.md#outdated

DrewML:
> will do more investigation and add a section for the existing solutions.

Thanks for posting the update! Good to know our options

DrewML:
In case you didn't see, Github acquired Dependabot

design-documents/tracking-tool.md
davidverholen: there is the composer repo api, to get sorted version information.
You can also access that on the command line, e.g.:
composer show -a symfony/console

Or accessing the packagist api, by just adding .json to the packagist package url:
https://packagist.org/packages/symfony/console.json

I think the versions are already sorted, you only have to filter out dev versions

Issue #88, opened 2019-02-14
Service Isolation -> Minimising package renaming impact


fooman:
As per [this](https://github.com/magento/architecture/blob/master/design-documents/service-isolation.md#split-modules) existing modules will be split into multiple packages.

This has a follow-on impact on existing extensions (ie nearly 100% of extension will require not only a new release but also a separately maintained branch). One idea to minimise the impact would be to keep the package name for CatalogAPI for existing extensions that only use the existing service contracts (ie current best practice extension get less impacted)

As an example an extension today has the following `composer.json`
```
    "require": {
        "magento/module-catalog": "^103.0.0",
```
and only uses the API interfaces. After splitting the modules ideally this should only be 
```
    "require": {
        "magento/module-catalog": "^103.0.0 | ^104.0.0",
```

This would allow the extension to be useful across both versions instead of requiring two separate versions to support both branches:

```
    "require": {
        "magento/module-catalog": "^103.0.0",
```
and
```
    "require": {
        "magento/module-catalog-api": "^100.0.0",
```

Further to allow transitioning into the end state of service isolation it would be great if this could be phased in. Once we know the split and what package names the new modules will have we can add them to the existing Catalog module.

So the existing package `magento/module-catalog` adds the following extra packages:
```
    "provide": {
        "magento/module-catalog-implementation": "self.version",
        "magento/module-catalog-admin": "self.version",
        "magento/module-catalog-storefront": "self.version",
    }
```
while the code is not yet moved out. That way it will be possible to release an extension that prepares for the split and is able to work across major release branches.

buskamuza:
Thanks, @fooman . That's also what we're thinking about to reduce the impact for extensions.
Will be reflecting this in a proposal as we work on it.
cc @joni-jones

joni-jones:
> So the existing package `magento/module-catalog` adds the following extra packages:

The new packages should depend on `module-catalog`, not vice versa. We want to split the business domain from the presentation layer, dependencies to other modules, etc. But the backward compatibility and extensions it's another point to discuss.

fooman:
@joni-jones The idea is to introduce the package split gradually. The above comment is pre actual code split.

So something like this. In the 2.3 branch we have
magento/module-catalog

In 2.4 we might have 
magento/module-catalog
- provides magento/module-catalog-implementation
- provides magento/module-catalog-admin
- provides magento/module-catalog-storefront

Bonus points here for deprecation notices .

[...]

PR #113, opened 2019-03-07
Split composer dependencies for modules/extensions

joni-jones:
## Problem

Magento modules and extensions have multiple dependencies to other Magento modules. This proposal describes strategies how to split and specify new modules in `composer.json`.

Relates to https://github.com/magento/architecture/issues/88.


design-documents/service-isolation/split-extensions.md
```diff
@@ -0,0 +1,81 @@
...
+
+ ```json
+"require": {
+    "magento/module-catalog": "^103.0.0|^104.0.0"
+}
+```
+
+Pros:
+ - No changes in the dependant code
+ - Allows specify the
```
melnikovi: Unfinished sentence.

design-documents/service-isolation/split-extensions.md
```diff
@@ -0,0 +1,81 @@
...
+}
+```
+
+Pros:
+ - No changes in the dependant code
+ - Allows specify the 
+  
+Cons:
+ - `composer.json` anyway should be changed to specify needed version of module
+ - All modules and extensions still have coupling to all dependant code if ti does not use
```
melnikovi: They don't use?

proposals/rich-content-graphql-api.md
```diff
@@ -1,380 +0,0 @@
-# Common GraphQL Representation of Rich Content
```
melnikovi: This proposal need to be moved to design-documents folder, not removed.

design-documents/service-isolation/split-extensions.md
```diff
@@ -0,0 +1,81 @@
...
+ - Extensions anyway should split their modules in the same way as core modules and specify the correct dependencies
+ - Such approach won't get benefits of modules decoupling as module|extension still depends on the whole component
```
melnikovi: Maybe at some point we could stop including meta package that aggregates all extensions by default. But if some extensions have dependency on meta package, they would download it and continue to work. There are would be no benefits from modularity, but it would be backwards compatible.

design-documents/service-isolation/split-extensions.md
```diff
@@ -0,0 +1,81 @@
...
+Pros:
+ - Introduce "right" dependencies
+ - Dependant code use only needed modules
+ - No changes in the dependant code
+  
+Cons:
+ - Dependant modules and extensions should update `composer.json`
+ - Extensions should split their modules in the same way as core modules and specify the correct dependencies
+
+### New module - old composer
```
melnikovi: Is this an option with meta package? If so, can we clarify this and say that it's similar to previous option, but we create meta package with old name for backwards compatibility?

joni-jones: It's only about composer dependencies between modules

design-documents/service-isolation/split-extensions.md
joni-jones: `Component` means for instance Catalog functionality, after splitting there will be multiple modules which aggregate the same component behavior

melnikovi: What do you mean by catalog functionality? Modules that Catalog module will be separated on or all of the modules related to catalog (including Bundle, ConfigurableProduct, etc)?

We can also aggregate modules main module will be split on using meta package that will have main module name. If the new module keep old namespace, it would be backwards compatible. Could you add it also to the list?

joni-jones: > What do you mean by catalog functionality? Modules that Catalog module will be separated on or all of the modules related to catalog (including Bundle, ConfigurableProduct, etc)?

Both, new modules from existing Catalog modules + Bundle, Configurable and all related as they represent catalog functionality.

> We can also aggregate modules main module will be split on using meta package that will have main module name. If the new module keep old namespace, it would be backwards compatible. Could you add it also to the list?

The last strategy describes this approach but based on composer packages, not a metapackage. I'm not sure about metapackage approach as 3rd party extensions depend on magento-module packages, not metapackages.

design-documents/service-isolation/split-extensions.md
joni-jones: Thanks, fixed.

design-documents/service-isolation/split-extensions.md
joni-jones: Thanks, fixed.

proposals/rich-content-graphql-api.md
joni-jones: Fixed

melnikovi:
Remaining comments
* Fix conflicts in design-documents/rich-content-graphql-api.md
* Specify that in New module - old composer we will use metapackage to aggregate extensions and keep old namespaces.
* Add that metapackages will not be part of the new project we release. And that you get benefits of decomposition only if all of your extensions split according to our recommendations. If you install extension that is not split, it will install metapackage and you loose all of the benefits of decomposition (you will not be able to deploy ui, admin ui, cron and web API separately).

design-documents/service-isolation/split-extensions.md
```diff
@@ -0,0 +1,80 @@
...
+
+This strategy proposes to expose all new modules via one (currently existing) `composer.json` and more described in [this proposal](https://github.com/magento/architecture/issues/88).
```
buskamuza: Add deprecation for old packages, so extensions move to the new structure and allow distributed deployment.

design-documents/service-isolation/split-extensions.md
```diff
@@ -0,0 +1,80 @@
...
+The main idea that extensions will use the same dependency in `composer.json` which will be distributed as metapackage. Magento core modules will depend on new modules like `catalog-ui` but not to `catalog` metapackage, such metapackage will be distributed separately.
```
buskamuza: Preserve old namespaces for existing code and introduce new namespaces for totally new modules.

buskamuza:
All agree to the 3rd approach.

buskamuza:
@rgoncharuk : maybe we should change namespace now? We should come to agreement before accepting the proposal.

design-documents/service-isolation/split-extensions.md
joni-jones: Added more explanation

design-documents/service-isolation/split-extensions.md
```diff
@@ -61,20 +61,41 @@ Cons:
...
- - Such approach won't get benefits of modules decoupling as an extension still depends on the whole component
\ No newline at end of file
+ - All extensions still have coupling to all dependant code even if they do not use it
+ - Such approach won't get benefits of modules decoupling as an extension still depends on the whole component
+ - To get benefits of distributed deployment extensions should be split as well
+ 
+## Summary
+
+The third approach looks like more convenient for now as allow combining two previous strategies: old modules will have old namespaces, new modules - new namespaces. Also, this approach allow splitting extensions according to new modules decomposition iteratively.
+The new modules won't be introduced in metapackage and extension developers should require them in `composer.json` manually.
```
melnikovi: Do you mean modules with new code or modules that we will get after module decomposition? I assume that you mean modules with new code, but please clarify in the proposal.

design-documents/service-isolation/split-extensions.md
joni-jones: Added to summary section

design-documents/service-isolation/split-extensions.md
```diff
@@ -61,20 +61,41 @@ Cons:
...
  - Allows to specify multiple package versions
   
 Cons:
+ - Introduce a metapackage with all needed dependencies
  - `composer.json` anyway should be changed to specify needed version of module
- - All modules and extensions still have coupling to all dependant code even if they do not use it
- - Such approach won't get benefits of modules decoupling as an extension still depends on the whole component
\ No newline at end of file
+ - All extensions still have coupling to all dependant code even if they do not use it
+ - Such approach won't get benefits of modules decoupling as an extension still depends on the whole component
```
melnikovi: To avoid confusion, could you clarify that this is only for extensions that are not decomposed? Maybe combine this and last one?

design-documents/service-isolation/split-extensions.md
```diff
@@ -61,20 +61,41 @@ Cons:
...
 "require": {
     "magento/module-catalog": "^103.0.0|^104.0.0"
 }
 ```
 
+Unlike the [original proposal]((https://github.com/magento/architecture/issues/88)), the `composer.json` probably will have another structure. As an example for `catalog` metapackage:
+
+```json
+"require": {
+    "magento/module-catalog-api": "^104.0.0",
```
melnikovi: Why don't we depend on specific patch version here?

joni-jones: The version here just for an example.

design-documents/service-isolation/split-extensions.md
joni-jones: Added clarification

design-documents/service-isolation/split-extensions.md
joni-jones: Updated.

melnikovi:
Most people voted for preserving old namespaces.

PR #131, opened 2019-03-22
AsyncImport - Update endpoints for base uploads

nuzil:
## Problem
After multiple discussion and real use cased we came to conclusions change workload a little bit for be able to be more flexible and same time make work for developers easier.

## Solution
I update endpoints structure.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=131) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=131) before we can accept your contribution.<br/>

design-documents/asynchronous-import/base-extension.md
```diff
@@ -50,15 +90,37 @@ Path is relative from Magento Root folder
...
+              }
+          ]
+      }
   }
 }
 ```
 
-Import of big file also can divided in several parts
-In this case input request will looks like:
+Import of big file also can divided in several parts.
```
paliarush:
```suggestion
Import of big file also can be divided in several parts.
```

design-documents/asynchronous-import/base-extension.md
```diff
@@ -68,30 +130,16 @@ In this case input request will looks like:
...
-      "uuid": "UUID" // Any UUID transferred by request creator. Not Required field
+      "uuid": "UUID"
   }
 }
 ```
 where *import_data* is a 1/N part of the whole content, and *data_hash* contains sha256 hash of full import_data body.
 
 `pieces_count` - its an amount of pieces that will be transferred for 1 file. We need it to be sure that import is completed and then we could detect if it was successfully finished or failed
-
 `piece_number` - its a number that detects which part of file currently transferred. This is required to have to support Asynchronous File import when we dont need to send parts in correct sequence
```
paliarush: Formatting is broken after removal of the line above..

design-documents/asynchronous-import/base-extension.md
```diff
@@ -68,30 +130,16 @@ In this case input request will looks like:
       "pieces_count": "5"
       "piece_number": "1",
       "import_type": "base64_encoded_data",
-      "source_type": "csv",
-      "uuid": "UUID" // Any UUID transferred by request creator. Not Required field
+      "uuid": "UUID"
```
paliarush: Why is `format` field not required here?

design-documents/asynchronous-import/base-extension.md
```diff
@@ -113,6 +161,44 @@ Example:
 }
 ```
 
+### Update Imported Source Format
+
+Its possible also to Update Format 
+
+PUT  `/V1/import/source/:uuid`
```
paliarush:
```suggestion
PUT  `/V1/import/source/csv/:uuid`
```

It still makes sense to have dedicated endpoint for validation and strict interface purposes.

design-documents/asynchronous-import/base-extension.md
nuzil: Yes you are right, forgot it.
But here is a reasonable question, if we are transferring multiple parts,  basically we need to transfer format only 1 time, which is ok.
But what IF customer transfer it in each body and some paramether will be different in some part? 
I would propose always use values from first request. If you want to change them, you have to use PUT

joni-jones:
@nuzil will it be a part of Magento core functionality or separate extension?

design-documents/asynchronous-import/base-extension.md
vkublytskyi: I would prefer to avoid any intelligence in such a scenario. Definition what is "first" request not clear as a client may initiate requests with all parts simultaneously. If requests have conflicting or missed data - throw an error and ask a client to resolve it.

As we don't have any instrument to declare "property required among several requests" I would prefer to make it required for requests with all parts. As this API created to process big amount of data, duplication of fields in request will not add significant overhead.

paliarush: Agree with @vkublytskyi there is no difference for a client app wether to include all data with each request or not. However it simplifies server side implementation by eliminating unnecessary state.

nuzil:
@joni-jones will be as separate service similar to MSI, so could be installed separately if required

PR #144, opened 2019-04-12
Mail library improvements

melnikovi:

### DIFF
--- a/design-documents/mail-library-improvements.md
+++ b/design-documents/mail-library-improvements.md
@@ -1,0 +1,338 @@
+# Mail library improvements
+
+## About
+
+Existing `\Magento\Framework\Mail\MailMessageInterface` missing many essential methods. There is no way to get data set on the object and library doesn't support sending attachments. This document describes improvements to Mail library.
+
+## Design
+
+Improvements to the library can be done in the following stages
+1. Add missing methods to the `\Magento\Framework\Mail\MailMessageInterface` interface
+2. Introduce new library with more clear interface that would allow to send attachments (wrappers around Zend\Mail)
+3. Refactor Magento modules to use new library
+
+1. Improving `\Magento\Framework\Mail\MailMessageInterface`
+    Below is the comparison with `Zend\Mail\Message`.
+    
+    - `isValid` - does’t need to be part of interface
+    + `setEncoding`/`getEncoding` - can pass in constructor
+    - `setHeaders`/`getHeaders`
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=144) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=144) before we can accept your contribution.<br/>

navarr:
@woutersamaey If I'm reading the RFC correctly, I think the From can have multiple addresses.

```
from            =       "From:" mailbox-list CRLF
mailbox-list    =       (mailbox *("," mailbox)) / obs-mbox-list
```

[Definition of mailbox-list](https://tools.ietf.org/html/rfc2822#section-3.4)
[Definition of from](https://tools.ietf.org/html/rfc2822#section-3.6.2)

woutersamaey:
@navarr OK. Not sure how what would translate in practise for Magento, but I don't object :-)

design-documents/mail-library-improvements.md
```diff
@@ -0,0 +1,338 @@
...
+
+1. Improving `\Magento\Framework\Mail\MailMessageInterface`
+    Below is the comparison with `Zend\Mail\Message`.
+    
+    - `isValid` - does’t need to be part of interface
+    + `setEncoding`/`getEncoding` - can pass in constructor
+    - `setHeaders`/`getHeaders`
+    + `setFrom`
+    - `getFrom`
+    - `setTo` - there is addCc, but not convenient
```
joni-jones: As `\Magento\Framework\Mail\MailMessageInterface` extends `\Magento\Framework\Mail\MessageInterface`, which already has `addCc`, does it make sense to introduce a new one method?

joni-jones:
@melnikovi could you describe how a composition of interfaces will look like in the code if logic should send HTML email with attachments?

design-documents/mail-library-improvements.md
melnikovi: I would introduce it, since `addCc` doesn't allow to replace.

melnikovi:
@joni-jones as interfaces are wrappers around `Zend\Mail` client code will look something like this https://github.com/zendframework/zend-mail/blob/master/docs/book/message/attachments.md.

PR #145, opened 2019-04-12
Module separation

melnikovi:
## Problem

Proposal describes different options for module names and namespaces when splitting modules.

### DIFF
--- a/design-documents/module-split.md
+++ b/design-documents/module-split.md
@@ -83,7 +83,7 @@ It makes sense to deploy admin and storefront API to different instances.
         * No overhead from preserving keeping backwards compatibility
         
         Cons
-        * We going to have incorrect namespaces
+        * We going to have namespaces that don't follow the pattern
         * Will need to fix existing tests (not a big con really)
         * Will have to decide what to do with new classes we introduce in the modules that have classes with old namespaces. Introduce with old namespace for consistency or with new namespace that matches module name to have correct namespace. Either way this might be a little confusing, and different from the case when we create brand new module with correct namespace
 
--- a/design-documents/module-split.md
+++ b/design-documents/module-split.md
@@ -8,15 +8,8 @@ This document clarifies the following two questions of the original [service iso
 
 ## Design
 
-### CLI
-It makes sense to move CLI in separate modules, because some commands are meant to me run only in admin area. Some commands are meant to run in both admin and storefront and will be shared.
-
-CLI is not meant to be deployed on separate instance. Because some commands need presence of all modules (static content deployment).
-
 ### Web API
 It makes sense to deploy admin and storefront API to different instances.
-
-Developer may want to deploy separately SOAP, REST and GraphQL. It should be possible, but we not necessarily need to distribute projects for SOAP, REST and GraphQL.
 
 ### Module naming
 
@@ -40,9 +33,6 @@ Developer may want to deploy separately SOAP, REST and GraphQL. It should be pos
     Cron (models and resources potentially)
     * CatalogCron _recommended_
     
-    CLI (models and resources potentially)
-    * CatalogCli _recommended_
-    
     WebApi
     * CatalogWebApi _recommended_
     
@@ -58,8 +48,8 @@ Developer may want to deploy separately SOAP, REST and GraphQL. It should be pos
     * CatalogImpl
     
     Common static resources
-    * CatalogCommonUi _recommended_
-    * CatalogUi
+    * CatalogCommonUi
+    * CatalogUi _recommended_
 
 2. Keep old or change namespaces
 
--- a/design-documents/module-split.md
+++ b/design-documents/module-split.md
@@ -37,6 +37,7 @@ This document clarifies the following two questions of the original [service iso
     Common code (business logic used by multiple areas, models, resource modelss)
     * CatalogCommon
     * CatalogShared
+    * CatalogDomain
     
     Common static resources
     * CatalogCommonUi
--- a/design-documents/module-split.md
+++ b/design-documents/module-split.md
@@ -78,5 +78,6 @@ This document clarifies the following two questions of the original [service iso
         Cons
         * We going to have incorrect namespaces
         * Will need to fix existing tests (not a big con really)
+        * Will have to decide what to do with new classes we introduce in the modules that have classes with old namespaces. Introduce with old namespace for consistency or with new namespace that matches module name to have correct namespace. Either way this might be a little confusing, and different from the case when we create brand new module with correct namespace
         
     Recommended option is to keep old namespaces.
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=145) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=145) before we can accept your contribution.<br/>

nuzil:
I guess rewriting all old modules in a new manner its a too big change in terms of backward incompatibility. Developers should not just align their modules, they have mostly rewrite all of them.
Also making magento update for all merchant will cost huge amount of time and money.
Maybe a solution here will be to leave old Namespaces and from release to release make vere small movements in a direction of new approach. I know that will take much more time, but will produce less pain.

design-documents/module-split.md
```diff
@@ -0,0 +1,81 @@
...
+    
+    CLI (models and resources potentially)
+    * CatalogCli
+    
+    WebApi (models and resources potentially)
+    * CatalogWebApi
+    
+    Common code (business logic used by multiple areas, models, resource modelss)
+    * CatalogCommon
+    * CatalogShared
```
joni-jones: Not sure about `Common` and `Shared`, it assumes that the code under these namespaces knows about its usage in the different areas. I propose to combine these namespaces into something like `Domain`.

design-documents/module-split.md
```diff
@@ -0,0 +1,81 @@
...
+    
+    2. Introduce new namespaces and don't preserve backwards compatibility
+        Pros
+        * We going to have correct namespaces after module split
+        * No overhead from preserving keeping backwards compatibility
+        
+        Cons
+        * Backwards incompatible, all of the extensions will be affected
+    
+    3. Keep old namespaces. Classes in CatalogAdminUi, CatalogStorefrontUi, CatalogCron, etc will have Catalog namespace
```
joni-jones: New classes probably will have new namespaces as we don't need to preserve backward compatibility for a new code.

design-documents/module-split.md
melnikovi: Added to the list.

design-documents/module-split.md
melnikovi: Yes. Added this for clarity.

design-documents/module-split.md
```diff
@@ -0,0 +1,83 @@
...
+        
+        Cons
+        * Backwards incompatible, all of the extensions will be affected
+    
+    3. Preserve old namespaces. Classes in CatalogAdminUi, CatalogStorefrontUi, CatalogCron, etc will have Catalog namespace
+        Pros
+        * No overhead from preserving keeping backwards compatibility
+        
+        Cons
+        * We going to have incorrect namespaces
```
Vinai: The namespaces are not incorrect. They simply don't follow the same pattern. The composer autoloader thinks they are perfectly valid.

melnikovi: You are right, I will rephrase.

design-documents/module-split.md
```diff
@@ -0,0 +1,83 @@
...
+        * May need to add new tests to make sure everything is consistent (edited) 
+        * One minor release probably is not enough time for developers to fix their extensions
+    
+    2. Change, introduce new namespaces and don't preserve backwards compatibility
+        Pros
+        * We going to have correct namespaces after module split
+        * No overhead from preserving keeping backwards compatibility
+        
+        Cons
+        * Backwards incompatible, all of the extensions will be affected
```
TravisBernard: This should not even be an option. You're talking about completely breaking every extension in the ecosystem overnight; causing every single extn developer to go fork their extn into two versions, one for older magento and one for this; and causing every develop to have to learn a new paradigm.  You know the costs of all that across the ecosystem is easily hundreds of thousands of dollars and right?

design-documents/module-split.md
rgoncharuk: CatalogBL (business logic)
Catalog DL (domain logic)

design-documents/module-split.md
```diff
@@ -0,0 +1,83 @@
...
+    Storefront Ui and storefront related business logic (blocks, controllers, static resources, models and resources potentially)
+    * CatalogStorefrontUi _recommended_
+    * CatalogStorefront
+    * CatalogStore
+    
+    Cron (models and resources potentially)
+    * CatalogCron _recommended_
+    
+    CLI (models and resources potentially)
+    * CatalogCli _recommended_
```
melnikovi: Need to clarify how the CLI will be used.

design-documents/module-split.md
```diff
@@ -0,0 +1,83 @@
...
+    
+    WebApi (models and resources potentially)
+    * CatalogWebApi _recommended_
```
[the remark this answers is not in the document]
paliarush: REST and SOAP have unified implementation, but represent different areas.
GraphQL is completely separate implementation and area.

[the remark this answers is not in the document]
joni-jones: We can't support some things for REST and SOAP at the same time, like associative arrays (SOAP doesn't allow to use them). Probably it makes sense to separate REST and SOAP.

design-documents/module-split.md
```diff
@@ -0,0 +1,83 @@
...
+    
+    Common code (business logic used by multiple areas, models, resource modelss)
+    * CatalogCommon
```
paliarush: Please consider having this module named Catalog and have metapackage named CatalogMetapackage (or something similar).

The issue is that third-party modules will not be able to get seamless upgrade if they depend on current Catalog. We need to evaluate how realistic is that we will not bump major version of the Catalog package with this module split.

design-documents/module-split.md
```diff
@@ -0,0 +1,83 @@
...
+1. Module naming
+
+    The following is proposed naming for modules:
+    API (service contracts)
+    * CatalogApi _recommended_
+    
+    The following is proposed naming for modules:
+    Admin Ui and admin related business logic (blocks, controllers, static resources, models and resources potentially)
+    * CatalogAdmin
+    * CatalogAdminUi _recommended_
```
melnikovi: Should we drop Ui postfix? We can have a case when resource model is only relevant only in storefront, as it's not UI it can be confusing.

design-documents/module-split.md
joni-jones: The idea was to don't introduce a new name and 3rd party developers will have a possibility to specify multiple versions in composer (option 3 from https://github.com/magento/architecture/blob/master/design-documents/service-isolation/split-extensions.md#new-module---old-composer)

design-documents/module-split.md
melnikovi: Decided to leave Ui postfix.

PR #146, opened 2019-04-16
GraphQL asynchronous mutations

akaplya:


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=146) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=146) before we can accept your contribution.<br/>

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+
+type OperationInProgress implements OperationStateInterface {
+     identifier: String!
+}
+```
+
+#### Request
+```graphql
+query {
+    queue($operataionIdentifier: [String!]!) {
```
real34:
```suggestion
    queue($operationIdentifier: [String!]!) {
```

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+}
+type OperationAccepted implements OperationStateInterface {
+     identifier: String!
+}
+```
+
+#### Request
+```graphql
+mutation {
+    reserveCart($cartMask: String!) {
```
paliarush:
```suggestion
    createEmptyCart($cartMask: String!) {
```

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+        __typename__
+    }
+}
+```
+
+#### Response
+```json
+{
+   "data":{
+      "operation":{
```
paliarush: Looks like there is no `operation` field in the schema currently, it should be added there. Otherwise "CreateEmptyCart(): identifier" sounds like it returns Cart Identifier.

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+
+Let's consider an example of an asynchronous workflow by using with shopping cart mutations.
+
+### Setting a command 
+Actually, because we do not want to process command immediately mutations can not return results.
+Instead of the result, such mutation will return a code which will identify a command in execution queue.
+
+```graphql schema
+type Mutation {
+    createEmptyCart(cartMask: String!): OperationStateInterface!
```
paliarush: How about
```suggestion
    createEmptyCart(cartUuid: String!): OperationStateInterface!
```

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+
+Generic way provides a possibility to return with success list of types that were changes during mutation execution.
+Basically, this is a list of updated entities which should be reloaded for application work. The application can use this result as tags for invalidation local cached entities etc.
+Filling of this list is a developer responsibility and does not guarantee by the framework.
+To return information about such entities operation state must implement 
+
+```graphql schema
+type ChangedEntity {
+    type: String!
+    identifiers: [String!]
```
paliarush: To me personally current option seems more complicated than flat structure of changed entities.
```suggestion
    identifier: String!
```

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+#### Request
+```graphql 
+query {
+    queue($operataionIdentifier: [String!]) {
+        identifier
+        __typename__
+        ... on PersistOperationCompleted {
+            changedEntities {
+                type
+                identifier
```
paliarush: I like current option, but it does not match current schema

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+#### Response
+```json
+{
+   "queue":[
+      {
+         "identifier":"3d8d487165a5653733e3a3e8bd46783e46b9a58d",
+         "__typename__":"PersistOperationCompleted",
+         "changedEntities":[
+            {
+               "Cart":[
```
paliarush: This response does not seem to match the schema

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+For instance, `CartSaveOperationCompleted` state may, return a corresponding cart as with query.
+```graphql schema
+type CartSaveOperationCompleted implements OperationStateInterface {
+     identifier: String!
+     cart: Cart!
+}
+```
+For such cases, a client should be aware of the operation and result that it expects to receive.
+This is kind of trade-off between generalized approach and numbers of request to the server.
+Such an approach may help to achieve better performance but definitely will cost more in development and testing.
```
paliarush: I would suggest to add recommendation to avoid this option unless needed for specific performance optimization.

design-documents/graph-ql/framework/async-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+### Mimicking async mutation
+
+Actually, with this proposal, we will have to change application workflows to support asynchronous communication. With the great benefit for the future system will face some challenges during the transition period.
+Async operations cause the domino effect. There are not enough to change the single mutation, all operation should be rewritten at some moment.
+Taking to account backward compatibility policies the migration logic should be defined in advance.
+So, the system should be proactive and provides a possibility to write a client application which will be resilient to mutations mode switch.
+
+To achieve this goal we can wrap the existing synchronous operations with `OperationCompleted` state.
```
paliarush: This is non-standard approach, which main benefit is ability to simplify client refactoring when they need async GraphQL for performance improvement. The main drawback is that the interface becomes complicated for everyone now, even for those who do not need async GraphQL.

nuzil:
@akaplya @paliarush 
Why is here used some abstract "identifier", why not to stick to return operation "uuid", which is unique and also already implemented in REST API Asynchronous operations?

jissereitsma:
I like the proposal so far and love to see the way it is going. However, there is one thing that is scaring me: With the new `createCustomer` mutation, it is dead-easy to create a script to create 1000s of customers in a loop: Theoretically, there is nothing wrong with this, but it could potentially lead to issues on the application side if not stopped. Additional tooling would be needed to apply rate limiting and other similar security measures. The faster and easier the GraphQL API becomes for things like these, the easier it is to abuse this and flood a shops database with fake entries. This is already an issue in the current version of Magento, but still theoretical (also because the same is possible with the regular frontend).

However, this problem becomes worse if asynchronous mutations lead to less loading time on the client side and more resources required on the server-side to process things. Imagine that I'm able to send in a cart-request for 100 products with complex discount rules in 1 second, while the server takes 10 seconds to correctly process this. This could potentially be a new DoS attack.

I, therefore, think we need to take some kind of rate limiting into account when dealing with cart-specific mutations. For instance, limitations of how many products could be added to cart, how many customers can be created from the same IP, etc.

cpartica:
aren't mutation supposed to be synchronous in graphql? Imagine the whole cartflow, you could send those mutations in the same request, but some of that flow is depending of previous executing before, like setPaymentMethod vs placeOrder

PR #156, opened 2019-05-08
 Update MFTF Versioning and BIC page

KevinBKozan:
## Problem
* MFTF BIC policy needed a comprehensive list of `change => version requirement`
## Solution
* I added a version increase matrix to the BIC policy page, containing all the permutations of MFTF test changes
## Requested Reviewers
@okolesnyk

### DIFF
--- a/design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
+++ b/design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
@@ -47,63 +47,52 @@ An approach of defining what each release should include was taken from [Semanti
 3. Introduce publication functionality for publishing `magento2-test-module` package type.
 4. Create metapackage with test packages only for each Magento edition.
 
-## Entity's attributes
-Changing any of this attribute will cause Backward Incompatible change.
-
-**Test entity:**
+## Version Increase Matrix
   
-  |xPath|attribute|
-  |---|---|
-  |`/tests/test`|name|
-  |`/tests/test/<ACTION> ⃰`|stepKey|
-  |`/tests/test/before/<ACTION> ⃰`|stepKey|
-  |`/tests/test/after/<ACTION> ⃰`|stepKey| 
-
-**Action Group entity:**
-
-  |xPath|attribute|
-  |---|---|
-  |`/actionGroups/actionGroup`|name|
-  |`/actionGroups/actionGroup/arguments/argument`|name|
-  |`/actionGroups/actionGroup/arguments/argument`|type|
-  |`/actionGroups/actionGroup/<ACTION> ⃰`|stepKey|
-
-**Data entity:**
-  
-  |xPath|attribute|
-  |---|---|
-  |`/entities/entitie`|name|
-  |`/entities/entitie/data`|key|
-  |`/entities/entitie/array`|key|
-  |`/entities/entitie/var`|key|
-  |`/entities/entitie/requiredEntity`|type|
-
-**Metadata entity:**
-  
-  |xPath|attribute|
-  |---|---|
-  |`/operations/operation`|name|
-  |`/operations/operation/field`|key|
-  |`/operations/operation(/object)+`|key|
-  |`/operations/operation(/object)+/field`|key|
-  |`/operations/operation(/object)+/array`|key|
-  |`/operations/operation/array`|key|
-  |`/operations/operation/array/object`|key|
-  
-**Page entity:**
-
-  |xPath|attribute|
-  |---|---|
-  |`/pages/page`|name|
-  |`/pages/page/section`|name|
-  
-**Section entity:**
-
-  |xPath|attribute|
-  |---|---|
-  |`/sections/section`|name|
-  |`/sections/section/element`|name|
+  |Entity Type|Change|Version Increase|
+  |---|---|---|
+  |ActionGroup|`<actionGroup>` added|MINOR
+  | |`<actionGroup>` removed|MAJOR
+  | |`<actionGroup>` `<action>` added|MINOR
+  | |`<actionGroup>` `<action>` removed|MAJOR
+  | |`<actionGroup>` `<action>` type changed|PATCH
+  | |`<actionGroup>` `<action>` attribute changed|PATCH
+  | |`<actionGroup>` `<argument>` added|MAJOR
+  | |`<actionGroup>` `<argument>` removed|MAJOR
+  | |`<actionGroup>` `<argument>` changed|MAJOR
+  |Data|`<entity>` added|MINOR
+  | |`<entity>` removed|MAJOR
+  | |`<entity>` `<array>` added|MINOR
+  | |`<entity>` `<array>` removed|MAJOR
+  | |`<entity>` `<array>` `<item>` removed|PATCH
+  | |`<entity>` `<field>` added|MINOR
+  | |`<entity>` `<field>` removed|MAJOR
+  | |`<entity>` `<required-entity>` added|MAJOR
+  | |`<entity>` `<required-entity>` removed|MAJOR
+  | |`<entity>` `<var>` added|MAJOR
+  | |`<entity>` `<var>` removed|MAJOR
+  |Metadata|`<operation>` added|MINOR
+  | |`<operation>` removed|MAJOR
+  | |`<operation>` changed|PATCH
+  |Page|`<page>` added|MINOR
+  | |`<page>` removed|MAJOR
+  | |`<page>` `<section>` added|MINOR
+  | |`<page>` `<section>` removed|MAJOR
+  |Section|`<section>` added|MINOR
+  | |`<section>` removed|MAJOR
+  | |`<section>` `<element>` added|MINOR
+  | |`<section>` `<element>` removed|MAJOR
+  | |`<section>` `<element>` changed|PATCH
+  |Test|`<test>` added|MINOR
+  | |`<test>` removed|MAJOR
+  | |`<test>` (before/after) `<action>` added|MINOR
+  | |`<test>` (before/after) `<action>` removed|MAJOR
+  | |`<test>` (before/after) `<action>` changed|PATCH
+  | |`<test>` (before/after) `<action>` type changed|PATCH
+  | |`<test>` `<annotations>` `<annotation>` added|PATCH
+  | |`<test>` `<annotations>` `<annotation>` changed|PATCH
+  | |`<test>` `<annotations>` `<annotation>` GROUP removed|MAJOR
 
 ---------------------------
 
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=156) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=156) before we can accept your contribution.<br/>

orlangur:
@KevinBKozan @okolesnyk can we simply reuse as much as possible from versioning for [other Magento components](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/dependencies.html)? And just add some minimum of MFTF specific on top of it.

orlangur:
@okolesnyk no, I mean minimize amount of MFTF-specific rules by reusing existing versioning policy to teh max.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
-  |`/sections/section/element`|name|
+|Entity Type|Change|Version Increase|
+|---|---|---|
+|ActionGroup|`<actionGroup>` added|MINOR
+| |`<actionGroup>` removed|MAJOR
+| |`<actionGroup>` `<action>` added|MINOR
+| |`<actionGroup>` `<action>` removed|MAJOR
+| |`<actionGroup>` `<action>` type changed|PATCH
+| |`<actionGroup>` `<action>` attribute changed|PATCH
+| |`<actionGroup>` `<argument>` added|MAJOR
```
paliarush: Please give example of arguments, why is adding an argument considered a MAJOR change?

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
+| |`<actionGroup>` `<argument>` removed|MAJOR
+| |`<actionGroup>` `<argument>` changed|MAJOR
+|Data|`<entity>` added|MINOR
+| |`<entity>` removed|MAJOR
+| |`<entity>` `<array>` added|MINOR
+| |`<entity>` `<array>` removed|MAJOR
+| |`<entity>` `<array>` `<item>` removed|PATCH
+| |`<entity>` `<field>` added|MINOR
```
paliarush: Please give an example

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
paliarush: if argument has default value it will be MINOR

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
paliarush: Should be `data` instead of `field`

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
+| |`<entity>` `<field>` removed|MAJOR
+| |`<entity>` `<required-entity>` added|MAJOR
+| |`<entity>` `<required-entity>` removed|MAJOR
```
paliarush: removal seems to be MINOR. Same may be true for `var`

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
+| |`<entity>` `<var>` added|MAJOR
+| |`<entity>` `<var>` removed|MAJOR
+|Metadata|`<operation>` added|MINOR
+| |`<operation>` removed|MAJOR
+| |`<operation>` changed|PATCH
```
paliarush: This may be MINOR, since compatibility with Magento is changing.
Same relates to `section=>element change`.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
+|Section|`<section>` added|MINOR
+| |`<section>` removed|MAJOR
+| |`<section>` `<element>` added|MINOR
+| |`<section>` `<element>` removed|MAJOR
+| |`<section>` `<element>` changed|PATCH
+|Test|`<test>` added|MINOR
+| |`<test>` removed|MAJOR
+| |`<test>` (before/after) `<action>` added|MINOR
+| |`<test>` (before/after) `<action>` removed|MAJOR
+| |`<test>` (before/after) `<action>` changed|PATCH
```
paliarush: Please clarify that steps and their sequence are expected to be preserved

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
+| |`<test>` (before/after) `<action>` type changed|PATCH
```
paliarush: Please clarify what is `type`

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
..., seeInField, seeInFormFields, seeInPageSource, seeInPopup, seeInSource, seeInTitle, seeLink, seeNumberOfElements, seeOptionIsSelected, selectOption, setCookie, submitForm, switchToIFrame, switchToNextTab, switchToPreviousTab, switchToWindow, typeInPopup, uncheckOption, unselectOption, wait, waitForAjaxLoad, waitForElement, waitForElementChange, waitForElementNotVisible, waitForElementVisible, waitForJS, waitForLoadingMaskToDisappear, waitForPageLoad, waitForText, assertArrayHasKey, assertArrayNotHasKey, assertArraySubset, assertContains, assertCount, assertEmpty, assertEquals, assertFalse, assertFileExists, assertFileNotExists, assertGreaterOrEquals, assertGreaterThan, assertGreaterThanOrEqual, assertInstanceOf, assertInternalType, assertIsEmpty, assertLessOrEquals, assertLessThan, assertLessThanOrEqual, assertNotContains, assertNotEmpty, assertNotEquals, assertNotInstanceOf, assertNotNull, assertNotRegExp, assertNotSame, assertNull, assertRegExp, assertSame, assertStringStartsNotWith, assertStringStartsWith, assertTrue, expectException, fail, dontSeeFullUrlEquals, dontSee, dontSeeFullUrlMatches, dontSeeInFullUrl, seeFullUrlEquals, seeFullUrlMatches, seeInFullUrl, grabFromFullUrl)
```
paliarush: Need to think about procedure of maintaining this list up-to-date. Or better give reference to externally defined list.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: If argument has default value I agree, it will be minor. I'll update this doc to reflect this.

As to the MAJOR change, if you add `<argument>` without the value, users need to update all existing references to the actionGroup otherwise MFTF will throw an exception about `Missing Argument`.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: Yes, it should be `<data>`
Adding a field via `<data>` corresponds to adding new functionality to an entity for use with MFTF, so it's more than a PATCH but backwards compatible (so it's not MAJOR)

Example I added `<entity name="ExistingEntity"><data key="newField">SomeValue</data>` that new data is available for use to any tests. Removing it means tests will break unless references are deleted.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: Originally it was, perhaps @okolesnyk may be able to provide the reason we bumped it to MAJOR?

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: Agree on this account.

On Section/element changed I realize need to break it into separate categories:
`selector` => `PATCH`
`parameterized` => `MAJOR` (making a selector parameterized will cause existing invocations to break)
`type` => `PATCH`

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: `type` = `click` for example, I'll change wording to make it more clear.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: Will do

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: I'll give reference to the list defined in MFTF documentation instead to remove need to keep it in sync.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
```diff
@@ -28,82 +31,75 @@ An approach of defining what each release should include was taken from [Semanti
...
-  |`/sections/section`|name|
-  |`/sections/section/element`|name|
+|Entity Type|Change|Version Increase|
+|---|---|---|
+|ActionGroup|`<actionGroup>` added|MINOR
+| |`<actionGroup>` removed|MAJOR
+| |`<actionGroup>` `<action>` added|MINOR
+| |`<actionGroup>` `<action>` removed|MAJOR
+| |`<actionGroup>` `<action>` type changed|PATCH
+| |`<actionGroup>` `<action>` attribute changed|PATCH
```
melnikovi: Which attribute? Might be minor if `ref` change or do we cover it separately.

KevinBKozan: @tomreece raised the same concern, `ref` means an actionGroup may change which can introduce new test actions. I think it can be changed to the following:
* `ref` => `MINOR` (since ext devs only care about merging before or after this set of actions, but the actions **do** change)
* `all others except for stepKey` => `PATCH` (since it will not affect merging).

melnikovi: Can it affect flow of the test and make test that depends on the flow fail (I don't know what are the other attributes)?

KevinBKozan: In the situation that:
```
ActionGroup1 contains AmOnPage, Click, FillField
ActionGroup2 contains AmOnPage, Click, FillField, Click(submit button)
```

Then yes, changing actionGroups can alter test flow. If my extension merges after this `actionGroup` call and relies on one more `fillField` before clicking submit then this would be BIC.

In light of the above, `ref` change should instead be MAJOR, no?

KevinBKozan: @melnikovi spoke with @tomreece and @jilu1, they both agree that changing `ref` would have to account for the most severe of test material changes, which would make it a `MAJOR`.

KevinBKozan: Also regarding other attributes, `actionGroup` element only has `ref`, `stepKey` (identifier attribute) and `before/after` which are not relevant to our own versioning.

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
KevinBKozan: @paliarush no response from @okolesnyk so I can try to answer it.

`required-entity` changes what is available for persisting an entity to Magento (for example, a required-entity can be a `custom_attribute` to a product). If you change the `required-entity`, you are potentially changing the definition of an entity outside a test's expected definition for the entity.

KevinBKozan:
@paliarush @melnikovi Can we merge this PR or resolve the last open question? There is only one unanswered question, and even if we go back and revise some of the BIC rules we still would benefit from this foundation being established.

melnikovi:
I would propose to close this PR and move content to versioning section of devdocs https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/. All backwards compatibility documentation is there.

KevinBKozan:
@okolesnyk what do you think about the above suggestion of moving this content to devdocs?

design-documents/testing/functional/versioning-and-backward-compatibility-policy.md
okolesnyk: If we will remove `required-entity` and some of extension tests are relying on having it, we will automatically brake those tests which made me think it is MAJOR.
It is possible that removing of it will not affect anyone but I wouldn't rely on it.

PR #157, opened 2019-05-08
Simplified composer project

buskamuza:
## Problem

Magento project `composer.json` contains a lot of definitions. The more definitions are present, the higher the necessity to "fix" some of them during upgrade.

## Solution

`composer.json` of the Magento project is revisited and definitions are either removed or moved to packages they belong to. As a result, `composer.json` is small and simple, chances that it needs to be fixed during upgrade are minimal.

### DIFF
--- a/design-documents/simplified-composer-project.md
+++ b/design-documents/simplified-composer-project.md
@@ -24,20 +24,21 @@ Performed as described on [DevDocs](https://devdocs.magento.com/guides/v2.3/inst
 composer create-project --repository=https://repo.magento.com/ magento/project-community-edition <install-directory-name>
 ```
 
-### Upgrade with the proposed project structure
+### Upgrade with the new project structure
 
 ```
 composer require magento/product-community-edition <new-product-version>
 ```
 
-### Upgrade with the old project structure and preserve it 
+### Upgrade with the old project structure and preserve it
 
-Use Magento's composer plugin.
+Discouraged.
+Update all the dependencies manually.
 
 ### Upgrade from the old project structure to the new one
 
-Modify `composer.json` to eliminate non-production dependencies, as described in the proposal.
-After that use approach described in "Upgrade with the proposed project structure"
+Modify `composer.json` to eliminate non-production dependencies, as described in the proposal or use plugin that will do it automatically.
+After that use approach described in "Upgrade with the proposed project structure".
 
 ## Low-level changes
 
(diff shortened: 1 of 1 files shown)


design-documents/simplified-composer-project.md
```diff
@@ -0,0 +1,172 @@
...
+
+Modify `composer.json` to eliminate non-production dependencies, as described in the proposal or use plugin that will do it automatically.
+After that use approach described in "Upgrade with the proposed project structure".
+
+## Low-level changes
+
+Based on current `composer.json` the following changes are proposed:
+1. Remove `require-dev` section
+1. Move `"conflict": { "gene/bluefoot": "*" }` to `magento/page-builder` module
+1. Remove `"Magento\\Framework\\": "lib/internal/Magento/Framework/"`, `"Magento\\": "app/code/Magento/"` and `"app/code/"` from `autoload` section
```
kandy: Can you clarify how it will affect development experience for developers who develop its own extension in app/code?

buskamuza:
1. Those changes are breaking for integrations
1. Cover extension/customization scenarios

design-documents/simplified-composer-project.md
```diff
@@ -0,0 +1,172 @@
...
+1. Move `"psr-4": { "Magento\\Setup\\": "setup/src/Magento/Setup/", "Zend\\Mvc\\Controller\\": "setup/src/Zend/Mvc/Controller/" }` to `magento/base` package
+1. Move `"autoload": { "exclude-from-classmap": [ "**/dev/**", "**/update/**", "**/Test/**" ] }` to `magento/base` package
+1. Move `"psr-0": { "": [ "generated/code/" ] }` to `magento/framework`
+1. Remove `"files": [ "app/etc/NonComposerComponentRegistration.php" ]`
```
buskamuza: Preserve this as it's used for customizations in production (@kandy )

design-documents/simplified-composer-project.md
```diff
@@ -0,0 +1,172 @@
...
+### Upgrade from the old project structure to the new one
+
+Modify `composer.json` to eliminate non-production dependencies, as described in the proposal or use plugin that will do it automatically.
+After that use approach described in "Upgrade with the proposed project structure".
+
+## Low-level changes
+
+Based on current `composer.json` the following changes are proposed:
+1. Remove `require-dev` section
+1. Move `"conflict": { "gene/bluefoot": "*" }` to `magento/page-builder` module
```
buskamuza: Discuss this one with PO.

pdohogne-magento:
Is there any merit to releasing a parallel `magento/dev-project-x-edition` project package that maintains the current structure for extension developers so we can keep the production project clean while still supporting those partners?

pdohogne-magento:
Another thought: Could we put all of our `require-dev` dependencies in a metapackage then just require that metapackage in the project so it's simpler to update existing dev projects for new versions (one requirement vs. many)?  Or are we concerned about making it more difficult for developers to change those dev dependencies?

buskamuza:
@pdohogne-magento , yes, one of the proposed options is a dev metapackage, so we can consider this.
A question about separate dev project I'd assume is that it would require having two different project. While with one project and `require-dev`, it's possible just to run Composer differently (w/ or w/o dev dependencies).

pdohogne-magento:
@buskamuza My updated proposal for the dev metapackage is to tell extension developers to run `composer require --dev magento/dev-requirements` instead of needing to install a separate project file.

Likewise, if we need to keep the NonComposerComponentRegistration autoload entry, is it possible to move that to the `magento/base` package?  The more things we can move out of the base project composer file, the more control we have for controlling upgrades.

I'm unsure if metapackages can handle `conflict` entries, but if they can then we may actually be able to have the dev requirements auto-update as part of the normal composer upgrade flow by having the `require-dev` constraint be `magento/dev-requirements: *` and then have each version of the metapackage have `conflict: { magento/product-community-edition: <2.3.2 >2.3.2 }`.

If we do the above, we might even be able to eliminate the need for the composer plugin altogether.

buskamuza:
Closing this proposal after a discussion.
Expected workflow for developers:
1. Base the project on Magento project (described in the proposal)
2. Perform development (extension/customization) using `dev` declarations
3. When pushing to staging/production run `composer` with `--no-dev` option to skip dev dependencies and autoload

Based on the above workflow, existing `composer.json` is a convenient template for a Magento project.

PR #158, opened 2019-05-09
Storefront APIs

akaplya:
This proposal describes the technical vision of possible storefront APIs


design-documents/storefront/storefront-api.md
```diff
@@ -0,0 +1,103 @@
...
+### Desired state
+
+![Storefront APIs](storefront-api/storefront-api-01.png)
+
+
+#### Storefront APIs are separated from admin APIs.
+
+We are using different object representation for storefront and admin scenarios.
+We are applying different rules and restrictions for such objects depending on where these objects appear.
+Currently, we do not have ACL resources specified for the storefront.
```
paliarush: ACL on storefront is possible in case of B2B

design-documents/storefront/storefront-api.md
```diff
@@ -0,0 +1,103 @@
...
+Blocks generates by instruction from assembled layout.
+GraphQL resolving data by traversing query tree.
+We can assume that all possible future storefront implementation
+will stick to this approach.
+This leads us to N+1 problem.
+Deferring data retrieving operation can resolve it.
+Technically, we do not want to have 
+a few similar repeating queries during the single page load.
+
+For the module perspective, this means that storefront APIs, even queries will be **asynchronous**.
```
paliarush: Please explain in more details and give examples for queries.

design-documents/storefront/storefront-api.md
```diff
@@ -0,0 +1,103 @@
...
+It does not mean that GraphQL or Blocks should be async either.
+They both as clients can execute APIs with synchronous adapter.  Technically we already have this implemented with GraphQL resolvers.
+
+[Promises/A+](https://promisesaplus.com/)
```
paliarush: The reference looks to be out of context.

design-documents/storefront/storefront-api.md
akaplya: Even for B2B, this is not an admin ACL, so statement still correct. Storefront operations use another authorization, ACL, etc.

design-documents/storefront/storefront-api.md
akaplya: Fixed

design-documents/storefront/storefront-api.md
kokoc: The statement is not correct. We have ACL on storefront and we need it(in all contexts, we don't want to expose orders and addresses to different people). However, this is not a reason to stop API separation. Why did you mention ACL in this context?

design-documents/storefront/storefront-api.md
```diff
@@ -0,0 +1,101 @@
...
+       |
+       ` getProductPriceTaxes       ----------> CatalogTaxStorefront
+
+```
+
+#### Storefront modules does not depends on each other
+
+Module purpose strictly restricted by the scenario it introduces.
+The are no particular reason to do this because
+all aggregation will be performed by the presentation layer.
```
kokoc: How to handle really big aggregations(like count of all products in the system, total revenue of the store, count of products with blue color, etc)?

design-documents/storefront/storefront-api.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Storefront APIs are replaceable
+
+Because we are going to manage storefront APIs as a set of 
+small granular modules which do not rely on each other
```
kokoc: This seems to be conflicting with `Involved APIs and modules` point. I see some dependencies between modules.

design-documents/storefront/storefront-api.md
akaplya: aggregation across service calls. regarding your example, graphql as a client may perform two API calls to resolve count and data retrieval operation.

design-documents/storefront/storefront-api.md
akaplya: dependencies will be resolved on a client level no need to mix services if clients can do it instead, as a result, this connection will be more natural and services will have more independent.

design-documents/storefront/storefront-api.md
akaplya: fixed

design-documents/storefront/storefront-api.md
```diff
@@ -0,0 +1,104 @@
+# Storefront API Technical Vision
+
+## Definition
+
+*Storefront APIs* - public operations and queries that
```
buskamuza: Please clarify through which interface.

PR #163, opened 2019-05-17
Batch query services

akaplya:
## Problem
Batch query services. This document specifies how input/output arguments should look like to implement batch query services

### DIFF
--- a/design-documents/batch-query-services.md
+++ b/design-documents/batch-query-services.md
@@ -51,17 +51,29 @@ But at least input arguments types are transparent for the developers.
      */
     public function getProductPrice(array $args) : array
 ```
-## Operations output
 
-Because we may receive multiple requests to one service it makes sense to return the results segregated per the request.
+
+
+
+
+## Results output
+
+Because we may receive multiple requests to one service 
+it makes sense to return the results segregated per the request.
+
+
+There are a few options to consider:
+
+### 1. Returns execution results with original request
+
 This decision allows us to reduce effort on the client to recognise part of the result that belongs to it.
 As a consequence, the client should not keep state (criteria) between requests. 
 
 ```php
-public class ResultContainer
+public class ResultContainer implements ContainerInterface
 {
     public function getCriteria() : CrtiteriaInterface
-    public function getResults() : array
+    public function getResult() : array
 }
 ```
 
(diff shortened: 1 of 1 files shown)


design-documents/batch-query-services.md
```diff
@@ -0,0 +1,77 @@
+#Batch query services
+
+## Purpose
+
+The main reason for such change is performance.
+The system triggers multiple queries during an operation (page load, service call). 
+Instead of making several requests for data retrieval,
+it would be more efficient to assemble all requests and send them as the single call to query service.
```
kokoc: I think it's worth to say that we are assembling multiple requests of the same type

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,77 @@
+#Batch query services
```
kokoc: I think we need to add error reporting section as it's very important part for batch APIs.

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,77 @@
...
+## Operations output
+
+Because we may receive multiple requests to one service it makes sense to return the results segregated per the request.
+This decision allows us to reduce effort on the client to recognise part of the result that belongs to it.
+As a consequence, the client should not keep state (criteria) between requests. 
+
+```php
+public class ResultContainer
+{
+    public function getCriteria() : CrtiteriaInterface
```
kokoc: I think that criteria is useless in api response. Clients may rely on query  position which is much simpler comparing to searchCriteria comparison

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,77 @@
...
+Even without internal code optimization such approach saves time that required for the service components bootstrap.
+* Such an approach saves time that required for the service components bootstrap.
+* The service implementation can be changed to guarantee better performance for batch operations.
+
+## Arguments of operation
+
+A service that manipulates with a single request could have a straightforward interface:
+```php
+    /**
+     * @param int $productId
```
kandy: should be int[] type

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,77 @@
...
+     * @return array
+     */
+    public function getProductPrice(array $args) : array
+```
+
+We can use object representation for input arguments to achieve better transparency and enforce strict types check.
+```php
+class ProductPriceSearchCriteria implements CriteriaInterface
+{
+    public function getProductId() : int;
```
kandy: should be int[] type

design-documents/batch-query-services.md
akaplya: done

design-documents/batch-query-services.md
akaplya: Not exactly correct. actually, this proposal about how to satisfy multiple clients of a single service. 
Arguments segmentation really matters in this case.

design-documents/batch-query-services.md
akaplya: see the previous comment.

design-documents/batch-query-services.md
akaplya: fixed

design-documents/batch-query-services.md
akaplya: added options to discuss

design-documents/batch-query-services.md
kandy: So, do you think that this service should not use batch? Looks like our Technical Vision say that all services should be batched

AlexMaxHorkun:
@akaplya Perhaps we don't have burden our (and 3rd party services) with these complex method definition requirements? Please check out https://github.com/magento/magento2ce/pull/4366 we could use deferred result proxies to keep the same method definitions but to postpone/make asynchronous operations

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,238 @@
...
+    public function getCriteria() : CrtiteriaInterface
+    public function getResult() : array
+}
+```
+
+Currently, the interface looks more clear.
+```php
+    /**
+     * @param ProductPriceSearchCriteria[] $args
+     * @return CrtiteriaInterface[]
```
buskamuza: Maybe `ResultContainer`?

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,238 @@
...
+     */
+    public function getProductPrice(array $args) : array
+```
+For some cases, requested information that returned with the response may help to build a stateless scenario.
+For instance, a widget which shows weather for the detected area and refreshes forecast every 5 minutes.
+
+Cons service always has to return request arguments increasing the size of the message that we are transfering through the network
```
buskamuza: "Cons:"

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,238 @@
...
+]
+```
+
+### 2. Return execution result in the order of received requests
+
+Results can be mapped to requests according to the order of incoming arguments.
+With this approach, we do not have to return request criteria with results.
+Client assumes that the order of received result containers is equal to the order of sent arguments.
+Service implementation has to guarantee this sort order.
+This approach does not work with services that returns
```
buskamuza: that returns what?

design-documents/batch-query-services.md
```diff
@@ -0,0 +1,238 @@
...
+         "minimalPrice":"8.56"
+      }
+   }
+]
+```
+
+## Error reporting
+
+An error may occur on overall service execution or during processing one of the received criteria.
+This is up to a developer how the service tolerates and handles runtime exceptions during.
```
buskamuza: during...

PR #168, opened 2019-05-24
Proposal to rewrite + modernize Static Content Deployment 😱

DrewML:
[Rendered Proposal](https://github.com/magento/architecture/blob/static-content/design-documents/node-static-content-deploy.md)

### DIFF
--- a/design-documents/node-static-content-deploy.md
+++ b/design-documents/node-static-content-deploy.md
@@ -1,0 +1,95 @@
+# Modernize Static Content Deployment
+
+**Note: This is a proposal for a possible improvement in the future. Even if we reach consensus, there are no guarantees yet that this would be implemented. Exploratory for the moment**
+
+- [Static Content Deployment](#static-content-deployment)
+  * [Docs](#docs)
+  * [Use Cases](#use-cases)
+  * [What does it do?](#what-does-it-do-)
+  * [Problems with Static Content Deployment](#problems-with-static-content-deployment)
+  * [Why rewrite Static Content Deployment?](#why-rewrite-static-content-deployment-)
+    + [Performance](#performance)
+    + [Access to Modern Tooling](#access-to-modern-tooling)
+    + [Accessible for Front-End Devs](#accessible-for-front-end-devs)
+    + [What opportunities open up with a re-write of SCD using node.js?](#what-opportunities-open-up-with-a-re-write-of-scd-using-nodejs-)
+  * [How do you re-write Static Content Deployment?](#how-do-you-re-write-static-content-deployment-)
+  
+## Docs
+- [Deploying Static View Files](https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-static-view.html)
+- [Strategies](https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-static-deploy-strategies.html)
+
+## Use Cases
+1. Pre-processing and flattening of static assets in local development (i.e. less)
+1. Pre-processing and flattening of static assets for production deployment
+
+## What does it do?
+
+- [Theme/module/i18n/"lib/web" file inheritance flattening](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/themes/theme-inherit.html)
+- Locale splitting
+    - Each locale gets its own static assets under a theme dir
+- Less compilation
+- HTML minification
+- JS minification
+- JS concatenation
+- `js-translation.json` generation
+- `requirejs-config.js` generation
+- Asset signing/versioning
+
+## Problems with Static Content Deployment
+
+There are a number of issues with the current implementation of static content deployment.
+
+1. It's prohibitively slow, sometimes taking 10+ minutes just to copy and process files
+1. There are 3 separate deployment strategies, all with different trade-offs that are not ideal. Having 3 separate ways to accomplish one task had led to [an explosion of open SCD bug reports (600+)](https://github.com/magento/magento2/search?q=static+content+deploy+is%3Aissue&type=Issues), where every strategy does not quite work as expected.
+    - **Quick**: Duplicates a ton of files, which makes cloud deployments slower
+    - **Standard**: Does way more work than it needs to, and duplicates tons of files
+    - **Compact**: Basically re-implements symlinks in the application layer, ships 150kb+ of extra JS to the front-end
+1. When something fails, it's very difficult to figure out why
+1. Lack of reliable cache + incremental builds (most people typically run it from scratch)
+1. Unable to introduce modern FE tooling (because it all expects a `node.js` runtime)
+1. Lacking granular reporting (unclear which parts of the process cause spikes in deployment times)
+
+## Why rewrite Static Content Deployment?
+
+There are 3 primary motivators for re-writing the Static Content Deployment tool
+
+### Performance
+
+The performance of `bin/magento setup:static-content:deploy` is _not great_, which leads to developers being unhappy about doing front-end work.
+
+It is also a very significant bottleneck for Magento Cloud deployments.
+
+### Access to Modern Tooling
+The majority of modern front-end tooling is written in JavaScript and executed using `node.js`. By moving our front-end compilation tool to JavaScript, we're able to leverage a large ecosystem of open-source tooling that cannot easily be interfaced with from PHP.
+
+A great example of this is our usage of LESS. The only [official compiler for the LESS language](http://lesscss.org/usage/) is written in JavaScript. Because of this, Magento uses a 3rd party (incomplete) implementation of the LESS compiler in PHP. However, because of the slowness associated with it, most developers prefer to use Magento's grunt setup _during local development_. The differences between these 2 compilers leads to [very unfortunate bugs](https://github.com/magento/magento2/issues/7231) when a production deployment is done.
+
+### Accessible for Front-End Devs
+When building something new, it's important to choose the right programming language for the job.
+
+Often, the "right" language is based on the skillset of those that will be maintaining the tool. As an example, the Magento CLI is written using PHP. PHP is not a language designed with the intent of being used to write CLIs, but it is capable of it. Magento chose to write `bin/magento` in PHP because it makes contribution accessible to those using the tool.
+
+A re-write in JavaScript means that the code for static content deployment can be maintained by front-end developers, who spend most of their time in Magento working on JavaScript/HTML/CSS.
+
+### What opportunities open up with a re-write of SCD using node.js?
+- Compilation of modern JS with [Babel](https://babeljs.io/)
+- Automatic live reloading of CSS during development
+- [Autoprefixer](https://github.com/postcss/autoprefixer) support for CSS
+- Linting/Unit-testing integrated in local dev and build pipelines
+- Automatic generation of a service worker via [Workbox](https://developers.google.com/web/tools/workbox/) (oh look, PWA functionality)
+- Incremental file watcher (build minimal changesets for fast dev mode)
+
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=168) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=168) before we can accept your contribution.<br/>

design-documents/node-static-content-deploy.md
```diff
@@ -0,0 +1,95 @@
...
+- JS concatenation
+- `js-translation.json` generation
+- `requirejs-config.js` generation
+- Asset signing/versioning
+
+## Problems with Static Content Deployment
+
+There are a number of issues with the current implementation of static content deployment.
+
+1. It's prohibitively slow, sometimes taking 10+ minutes just to copy and process files
```
joni-jones: It always depends on the number of files to deploy but for Magento vanilla installation (CE + EE) it takes 13sec on my laptop.

design-documents/node-static-content-deploy.md
```diff
@@ -0,0 +1,95 @@
...
+## Use Cases
+1. Pre-processing and flattening of static assets in local development (i.e. less)
+1. Pre-processing and flattening of static assets for production deployment
+
+## What does it do?
+
+- [Theme/module/i18n/"lib/web" file inheritance flattening](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/themes/theme-inherit.html)
+- Locale splitting
+    - Each locale gets its own static assets under a theme dir
+- Less compilation
```
joni-jones: I agree about Less compilation and JS minification but not sure about other tasks like moving files, concatenation, and locales generation. I suppose the problem that you need to interact with the file system frequently and as node.js is also single threaded will it be a big difference between node.js and PHP? From my experience python or Go are more suitable for these tasks.

design-documents/node-static-content-deploy.md
```diff
@@ -0,0 +1,95 @@
...
+### Accessible for Front-End Devs
+When building something new, it's important to choose the right programming language for the job.
+
+Often, the "right" language is based on the skillset of those that will be maintaining the tool. As an example, the Magento CLI is written using PHP. PHP is not a language designed with the intent of being used to write CLIs, but it is capable of it. Magento chose to write `bin/magento` in PHP because it makes contribution accessible to those using the tool.
+
+A re-write in JavaScript means that the code for static content deployment can be maintained by front-end developers, who spend most of their time in Magento working on JavaScript/HTML/CSS.
+
+### What opportunities open up with a re-write of SCD using node.js?
+- Compilation of modern JS with [Babel](https://babeljs.io/)
+- Automatic live reloading of CSS during development
```
joni-jones: Could you clarify what do you mean? Now, in the development mode, the css/js will be generated in runtime.

design-documents/node-static-content-deploy.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Why rewrite Static Content Deployment?
+
+There are 3 primary motivators for re-writing the Static Content Deployment tool
+
+### Performance
+
+The performance of `bin/magento setup:static-content:deploy` is _not great_, which leads to developers being unhappy about doing front-end work.
+
+It is also a very significant bottleneck for Magento Cloud deployments.
```
joni-jones: I think we need some benchmarks to compare different PL.

design-documents/node-static-content-deploy.md
DrewML: > I suppose the problem that you need to interact with the file system frequently and as node.js is also single threaded will it be a big difference between node.js and PHP

An extremely large different. PHP blocks when reading a file, node does not.

node's original selling point was async network and disk I/O - that's where it thrives. Since more of static content deploy is copying/resolving/moving, rather than pure CPU-bound tasks, it's a good fit

DrewML: > From my experience python or Go are more suitable for these tasks.

This leads us back to the same problem as PHP (outlined in this document): lack of a good interop story with node. All modern front-end tooling is written assuming node as a runtime.

design-documents/node-static-content-deploy.md
DrewML: Example:

You are on the checkout page. You have 3/4 of the checkout form filled out. You now want to tweak some values in CSS.

Currently, your options are:

1. Make the change in DevTools, then copy/paste into your store CSS
2. Make the change on the file system, then refresh the whole page and re-enter all form data

Live reloading injects the CSS changes into the live document - you get to keep your browser state, but use your own editor.

design-documents/node-static-content-deploy.md
DrewML: Right, you're on the happy path it was designed for, which is part of the problem: those speeds are only available on stores with a limited number of files and only 1 locale.

If it takes several minutes for multiple locales, and the majority of our user base in Europe is using multiple locales, then the speed of a basic store isn't super relevant.

design-documents/node-static-content-deploy.md
DrewML: Do we? The document states multiple reasons for using node - perf is 1 of several

DrewML: I believe you may have also missed this section in the document:
--------
### Accessible for Front-End Devs
When building something new, it's important to choose the right programming language for the job.

 Often, the "right" language is based on the skillset of those that will be maintaining the tool. As an example, the Magento CLI is written using PHP. PHP is not a language designed with the intent of being used to write CLIs, but it is capable of it. Magento chose to write `bin/magento` in PHP because it makes contribution accessible to those using the tool.

 A re-write in JavaScript means that the code for static content deployment can be maintained by front-end developers, who spend most of their time in Magento working on JavaScript/HTML/CSS.

design-documents/node-static-content-deploy.md
jedmao: Anyone who has worked on a dev server with hot reloading (pretty common among FE devs) would be offended at the non-existence of it. It's really painful to not have it nowadays. We need fast, iterative development cycles and if we don't have live reloading, it slows us down (a lot).

design-documents/node-static-content-deploy.md
jedmao: In other words, benchmarks or not, this one point should not be the dealbreaker for this proposal.

design-documents/node-static-content-deploy.md
joni-jones: Should be there some kind of file watcher? Will it be a separate tool or an existing one?

design-documents/node-static-content-deploy.md
joni-jones: Yes, I have seen this section. That's why I raised this question about benchmarks. I'm not forcing to use some specific language, my point is the usage applicable language especially if we are going to introduce a new stack dependency.

DrewML: > I'm not forcing to use some specific language, my point is the usage applicable language especially if we are going to introduce a new stack dependency.

Yep! Definitely agree, and glad you're bringing it up. If we were just shooting for raw performance here, I'd definitely agree we should just choose whichever language + runtime gives us the best running times.

But, one of the critical parts of the goal here is developer experience. Leveraging the ecosystem of common tools for front-end work will allow front-end devs in the Magento world to re-apply their skills they've learned elsewhere, and get productive quicker.

design-documents/node-static-content-deploy.md
DrewML: Yep! That would be the idea for development mode (when actively working on front-end tasks). You typically have a persistent process running that keeps a warm cache and does small incremental rebuilds when files change (which typically includes running linters/unit tests/etc).

The [typical file watcher](https://github.com/paulmillr/chokidar) lib used in node tooling supports the native watch mechanisms on most operating systems, but can fall back to polling for mounts (for our friends developing with vagrant).

design-documents/node-static-content-deploy.md
Igloczek: It would be great to thing about just styles compilation and be able to extend those tools with other compilers like SASS, Stylus or some PostCSS based process.
I thing people will be happy to migrate from Frontools to offical tools 😊

design-documents/node-static-content-deploy.md
```diff
@@ -0,0 +1,95 @@
...
+A high-level set of steps:
+
+1. Write code to scrape necessary configurations from Magento (mainly `config.php` to determine what modules to work with)
+1. Build a quick prototype that proves that file inheritance can effectively be re-implemented with _significantly_ better performance outside of Magento core, leveraging both symlinks and copy-on-write where available
+1. Create a base architecture for a CLI that can be made up of various tasks (minify, less, etc etc)
+1. Create a backlog of critical tasks needed for a fully functional Static Content Deployment tool, and open up for community contribution
+1. Iterate, iterate, iterate
+
+## Prior Art
+- https://github.com/SnowdogApps/magento2-frontools/issues/25 Static Asset Processing (used by many Magento FE developers and agencies)
```
Igloczek: 😍

design-documents/node-static-content-deploy.md
```diff
@@ -0,0 +1,95 @@
...
+When building something new, it's important to choose the right programming language for the job.
+
+Often, the "right" language is based on the skillset of those that will be maintaining the tool. As an example, the Magento CLI is written using PHP. PHP is not a language designed with the intent of being used to write CLIs, but it is capable of it. Magento chose to write `bin/magento` in PHP because it makes contribution accessible to those using the tool.
+
+A re-write in JavaScript means that the code for static content deployment can be maintained by front-end developers, who spend most of their time in Magento working on JavaScript/HTML/CSS.
+
+### What opportunities open up with a re-write of SCD using node.js?
+- Compilation of modern JS with [Babel](https://babeljs.io/)
+- Automatic live reloading of CSS during development
+- [Autoprefixer](https://github.com/postcss/autoprefixer) support for CSS
```
Igloczek: I think the whole PostCSS should be there and Autoprefixer as one of the options.

design-documents/node-static-content-deploy.md
Igloczek: Do you still plan to have some development mode auto processing of missing assets?

From PoV of a agency that is doing things own way - kill it with fire

From an imaginary possition of full stacks, backends and other people who might not know modern frontend tooling - "it doesnt't work after changes, just got a page without sytes and can't click anything" - at least that what I see from questions asked over years about SASS Blank which requires to manually run some tools to get styles.

design-documents/node-static-content-deploy.md
DrewML: Y'all did good work! Deserve some credit :)

design-documents/node-static-content-deploy.md
DrewML: Definitely have no disagreements there. BUT, right now I'm focused 100% on minimal feature parity. We need this thing to be correct + fast, then we can have fun!

design-documents/node-static-content-deploy.md
DrewML: I think we need some solution here. Just like I don't want to setup docker + a bunch of microservices locally, most backend folks won't want to be running a persistent node process.

We'll find a good middle-ground here.

DrewML:
I haven't seen any objections here, and I have a (so far promising) prototype going of this work locally, so I'm going to merge this and get it out of the queue.

Happy to iterate if folks have further feedback, but this is outside of core so it's low-risk 🎉

DrewML:
Good progress being made - this work now lives here while it's still a POC

https://github.com/DrewML/scd

PR #169, opened 2019-05-24
Webhooks

kokoc:
## Problem

As Magento is becoming more and more a composition of internal APIs and external services, Rest and GraphQL APIs are not able cover every real-life use case. Traditional Web APIs allow to retrieve or post content in a programmatical way, far from being optimal when you just want your application to be told when something changed. If that is the case, it would require polling of whole data at regular intervals, and that just doesn’t scale.

## Solution

A webhook (also called a web callback or HTTP push API) is a way for an app to provide other applications with real-time information.

## Requested Reviewers

@joni-jones , @paliarush  , @kandy


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=169) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=169) before we can accept your contribution.<br/><hr/>**Andrii Konosov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

DrewML:
Nice! One interesting use-case this opens up: being able to use 3rd party tools like IFTTT/Zapier based on Magento-triggered events. Really lowers the barrier for people that want to create cheap integrations between Magento and an external service without writing code.

jedmao:
FWIW, GraphQL actually does have a mechanism for persistent connections called a [subscription](https://facebook.github.io/relay/docs/en/subscriptions).

## References
- [Subscriptions in GraphQL and Relay](https://graphql.org/blog/subscriptions-in-graphql-and-relay/)
- [Subscriptions - Relay](https://facebook.github.io/relay/docs/en/subscriptions)
- [GraphQL | Operation names](https://graphql.github.io/learn/queries/#operation-name)

AlexMaxHorkun:
@kokoc Do you consider allowing using public web API for integrations to subscribe to these events? Maybe some merchants would like 3rd party systems to be able to react to certain events of their website. If so then it would make sense to assign ACL resources or find another way how to limit events to which integrations can subscribe

design-documents/webhooks.md
```diff
@@ -0,0 +1,160 @@
...
+```
+:warning: It's very important to use query API for data retreval purposes where applicable. Using query API allows to apply service business logic on top of changed data, so we can reduce copy-paste of business logic in consumers. Catalog example: `visibility` field + category permissions functionality + child product functionality affect the actual product visibility on store-front, so we need to send `category_assigment` event instead of providing the raw `visibility` field that should be processed on consumer's side in exactly same manner as Magento processes.:warning:
+
+Webhook messages are created using provided payload that must implement `JsonSerializable` interface. Example:
+```php
+/**
+ * The output of jsonSerialize is kind of api as external systems will rely on it's content
+ * @api
+ */
+class CustomerUpdatePayload implements \JsonSerializable
```
AlexMaxHorkun: in webhooks.xml you mention Magento\Customer\Api\Data\CustomerQueryInterface which looks like a regular DTO but here you say that payloads must return array of scalars to be serialized to JSON. What if we decide to use multiple gateway to send these messages and some of them don't employ JSON? I think it will be better to have DTOs as payloads and different gateway services serializing them as they see fit

scottsb:
> There should be at least three ways to subscribe for webhook message: ...
> - programmatically using web API - useful for isolated services which communicate with Magento by API only

Zapier provides a good description of how this can work:
https://zapier.com/engineering/webhook-design/
https://zapier.com/engineering/static-webhooks/

kokoc:
> Do you consider allowing using public web API for integrations to subscribe to these events? Maybe some merchants would like 3rd party systems to be able to react to certain events of their website. If so then it would make sense to assign ACL resources or find another way how to limit events to which integrations can subscribe

@AlexMaxHorkun Yes, there should be a separate web api for subscriptions and it should be covered by separate acl resource.  This is covered in `Other considerations`, let me know if it needs more explanation.

design-documents/webhooks.md
kokoc: DTO is not really important, I planned to utilize JsonSerializable contract instead. JsonSerialize doesn't produce json in output, most probably it's just a php array. Different gateways can use different formats for communications.

design-documents/webhooks.md
```diff
@@ -0,0 +1,160 @@
...
+
+![webhooks net flow](img/webhooks_flow.png)
+
+Simpler example:
+* Customer register on storefront (Command API)
+* Command API emits `after_transaction_commit` event providing customer id after committing customer's data
+* Webhook listener calls customer query API with provided customer id. The response may contain new data that was not provided in original request(e.g. totalSpent, customer_group, total_orders, etc)
+* Webhook framework notifies all subscribers with customer data
+
+:warning: Webhooks are supposed to push data updates only. Initial data should be extracted from Magento using standard APIs like GrapQL and REST. :warning:
```
nathanjosiah: I've done quite a bit of freelance work and this would render this feature nearly completely useless for the ways that most clients would want to use it. For example, having a Zap that adds a customer to a CRM workflow when they are created. Or having a customer unsubscribed from a mailing list when they are deleted.

design-documents/webhooks.md
```diff
@@ -0,0 +1,160 @@
+## Webhooks
```
paliarush: I am trying to understand the benefits of webhooks over message queue for us, considering the cost of implementation and support. 

Are there any use cases not covered with existing message queue framework?

nathanjosiah: @paliarush AFAIK there isn't a way to easily use message queue to call external http endpoints?

design-documents/webhooks.md
kokoc: This point is not about partial data, but more about the initial system state(e.g. send all existing products using webhooks at the moment of subscription). I'll remove this point in order to avoid any confusion.

PR #172, opened 2019-05-29
MFTF test packaging proposal

okolesnyk:
## Problem

Tests are dead code on production env.

## Solution

Exclude tests from `magento2-module` package and introduce new type of packages with Tests only
There is going to be new modules which will have Functional Tests only. 
Example:
`CatalogAdminUi` - will have Magento Admin UI logic inside
`CatalogAdminUiFunctionalTest` - will have Automated Test Coverage for Catalog Magento Admin UI module

## Requested Reviewers
@paliarush 
@maghamed 
@buskamuza 
@antonkril


design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+### Proposed solution
+During Magento package new releases automated tool will exclude tests from main `magento2-module` package and create other package with type `magento2-test-module`.
+
+## Packaging flows
+### General Flow
+1. Exclude `Test` folder from module and keep it for further packaging flow as it shown on image below.
+
+![Packaging Process](img/packaging.png)
```
ravmenon: This is different from the plan shown here where the Test/MFTF has only composer.json: https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/module-file-structure.html#module-file-structure

The original plan assumed that MFTF can run independently outside of an M2 instance, hence only this needs to be packaged separately.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+
+2. Create `magento2-module` package as usual, but without tests.
+
+3. Out of excluded tests create a separate package.
+   - For this package we will automatically generate `composer.json` based on existing module `composer.json` file as shown on image below.
+    
+![Composer file](img/test_composer_json.png)
```
ravmenon: I am assuming the 'name' field being the same is a typo?  It can't be the same name from the composer repo perspective, so there should be a convention to suffix '-tests'  here?

design-documents/testing/functional/test-package.md
okolesnyk: Hi @ravmenon 
Yes, this picture is a bit outdated. I'll update it to show that MFTF directory is going to be packaged only.

vkublytskyi:
For sure, code that may be executed is a subject of attack but in case of MFTF module contains only XML configuration so no executable that may be used somehow while MFTF itself is not installed.
 
Another question that deserves explicit discussion is what is the production environment for Magento. I totally agree that running instance of Magento that serve real merchant customers should not contain any executable test code. But are composer packages that we produce production code or not? With removing our tests from packages we increase the complexity of system integrators and vendors workflow and internal and external CI infrastructures. It also leads to controversial decisions in release management and packaging.
 
Alternatively, we may consider a possibility to provide instruments to remove test code (including from any possible 3rd party extensions and SI customizations) as part of operations during deploying Magento. This may be done as CLI command which may be part of enabling the production mode. Or it may be done by Composer plugin that will remove test code if `--no-dev` flag is used for `composer install` or `composer update` commands. In addition, we may detect available test code and show a warning message in the admin panel to clearly communicate potential security issue and need to perform actions on the server side.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Version setter flow #2
+1. Versions will be set independently from Magento module based on SVC tool results.
+2. Metapackage `magento/product-tests-community-edition` and `magento/project-tests-community-edition` created and have the same versions as Magento metapackage.
+
+#### Benefits:
+1. Possibility to release test package only
+
+#### Disadvantages: 
+1. Bigger changes to Publication build
```
paliarush: MFTF versioning proposal becomes obsolete because tests version bump must always be aligned with Magento. Also, it is possible that tests will have breaking changes while the module itself will have PATCH changes.
Does not seem like this is even an option.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+3. Difficult to determine what version of test package we should use against existing Magento Instance
+   - Open question about "What package requires what?" see in "[Open Questions](#Open-Questions)" block
+
+
+## Open Questions:
+1. What package requires what?
+   - Does Magento require-dev test packages?
+   - Does test packages require Magento module?
+   - Other potentials?
+2. Are we using a dynamically created `composer.json` or we create them in advance under `app/code/<vendor_name>/<module_name>/Test/Mftf/` directory?
```
paliarush: We can follow the same approach as for core modules: the list of dependencies is specified by developers, exact versions are calculated and set during publication.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+2. Not possible to make Backward incompatible changes without generating new Metapackage
+3. Difficult to determine what version of test package we should use against existing Magento Instance
+   - Open question about "What package requires what?" see in "[Open Questions](#Open-Questions)" block
+
+
+## Open Questions:
+1. What package requires what?
+   - Does Magento require-dev test packages?
+   - Does test packages require Magento module?
+   - Other potentials?
```
paliarush: I would say tests must not require magento modules because they can be installed remotely. 

Magento modules should also not depend on tests. Maybe tests can be added to `suggest` section.

It looks like we need some tool which will identify which test modules needed for specific Magento instance (via API calls).

design-documents/testing/functional/test-package.md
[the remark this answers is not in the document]
okolesnyk: I think best will be best to specify real package names:
`magento/module-catalog` and `magento/test-module-catalog` as it shown on `test_composer_json.png`

Will update diagram.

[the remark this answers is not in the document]
okolesnyk: Updated.

design-documents/testing/functional/test-package.md
melnikovi: Do we want to support installing tests separately? Might be hard to check version compatibility if you want to install only some test modules but not all.

melnikovi: Discussed in person and agreed that it might be better not to have dependency in test modules on Magento modules because of two reasons: 1) it would allow to install test modules separately from Magento, 2) we will not have to deal with issue of updating of Magento modules when installing test modules on Magento instance.

@okolesnyk could you investigate tooling that would allow to make sure that test modules compatible with Magento modules? One idea we had is to have dependency on Magento module specified in the extra section of composer.sjon of a test module. Tool that would collect list of installed modules on Magento instance and search for test packages in configured repositories and create composer.json with found test modules.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,95 @@
...
+1. Versions will be set independently from Magento module based on SVC tool results.
+2. Metapackage `magento/product-tests-community-edition` and `magento/project-tests-community-edition` created and have the same versions as Magento metapackage.
+
+#### Benefits:
+1. Possibility to release test package only
+
+#### Disadvantages: 
+1. Bigger changes to Publication build
+    - require running SVC against latest release and current changes to determine package version to be set
+2. Not possible to make Backward incompatible changes without generating new Metapackage
```
buskamuza: This point is not clear... It doesn't matter wether you do breaking or non-breaking changes, you always have to change metapackage if you want people could update automatically. But you can release new version of tests with breaking changes whenever you want, it just won't be included in the metapackage until next Magento release. Same as we have for MSI.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,95 @@
...
+3. Difficult to determine what version of test package we should use against existing Magento Instance
```
buskamuza: It seems unclear for extensions only. For vanilla Magento, just use corresponding metapackage.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,95 @@
...
+2. Metapackage `magento/product-tests-community-edition` and `magento/project-tests-community-edition` created and have own versions based on smaller package version bumps.
+
+#### Benefits:
+1. Possibility to release test package only.
+2. Full independence in releases from Magento release lines.
+
+#### Disadvantages: 
+1. Bigger changes to Publication build
+   - require running SVC against latest release and current changes to determine package version to be set
+2. Not possible to make Backward incompatible changes without generating new Metapackage
```
buskamuza: Same note as for version 2.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,104 @@
...
+    "license": [
+        "OSL-3.0",
+        "AFL-3.0"
+    ]
+}
+```
+
+#### Magento metapackage
+In Magento `magento/project-community-edition` we will be able to specify 2 metapackage under different blocks.
+Magento application will go to `require` section as it is right now and we will be able to put `magento/functional-test-project-community-edition` under `require-dev` section.
```
piotrekkaminski: functional-tests (instead of test?)

okolesnyk: Yes. This proposal is about packaging Functional tests only. And they shouldn't be distributed with other test types together.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+1. Bigger changes to Publication build
+   - require running SVC against latest release and current changes to determine package version to be set
+2. Not possible to make Backward incompatible changes without generating new Metapackage
+3. Difficult to determine what version of test package we should use against existing Magento Instance
+   - Open question about "What package requires what?" see in "[Open Questions](#Open-Questions)" block
+
+
+## Open Questions:
+1. What package requires what?
+   - Does Magento require-dev test packages?
```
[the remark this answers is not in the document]
okolesnyk: This question is outdated.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,96 @@
...
+   - Does test packages require Magento module?
```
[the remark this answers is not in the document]
okolesnyk: This question is outdated.

design-documents/testing/functional/test-package.md
buskamuza: `project` is something that is a project by itself. Tests metapackage is just a combination of packages and can't be installed standalone. So this metapackage should be something like `magento/functional-test-community-edition` (no `project`).

okolesnyk: Yes, that is typo. It meant to be `magento/functional-test-product-community-edition`

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,104 @@
...
+>Security:
+> - reduce possible attack surface (less code)
+> - the test code is not written with security in mind
+> - even if we protect the access via .htaccess, some sites run Nginx or skip .htaccess making this code accessible
+> - we had security issues in the past that resulted in remote code exec using code related to tests
+>
+>Future-proof:
+> - ability to release separately, on separate cadence
+> - matches the general plan for all other non-critical modules - we want to modularize and release separately things like payment extensions, shipping and more.
+
```
lbajsarowicz: Performance:
* MFTF coverage grows drastically, which side effect is growing deployment time for staging / production environment

lbajsarowicz:
I'm not sure if I'm missing something, but...
I'd take another approach. Each module having it's FunctionalTest dependency in it's `composer.json` as a `-dev` requirement. So that doint `composer install --no-dev` just does not install the tests.

In practice - each module needs tests only for `dev` environment, so this dependency would be reasonable.

design-documents/testing/functional/test-package.md
```diff
@@ -0,0 +1,109 @@
...
+    "license": [
+        "OSL-3.0",
+        "AFL-3.0"
+    ]
+}
+```
+
+#### Magento metapackage
+In Magento `magento/project-community-edition` we will be able to specify 2 metapackage under different blocks.
+Magento application will go to `require` section as it is right now and we will be able to put `magento/functional-test-project-community-edition` under `require-dev` section.
```
xmav: We might want to move "magento/magento2-functional-testing-framework" from 'require-dev' to new meta package with tests as well

buskamuza:
@lbajsarowicz, `dev` dependencies are supported in the root `composer.json` only.

buskamuza:
> the functional test metapackage should not live in the require-dev section of
magento/project-community-edition
but rather in the require-dev section of magento/product-community-edition
Same. It will not work. `require-dev` is supported on the project level only.

fooman:
@buskamuza sorry missed this discussion and your comment. Keeping the testing metapackage out of the root composer.json require-dev section was exactly my point. If someone chooses to use the same tools as Magento for their development - great. It's only a 
`composer require --dev magento/functional-metapackage` away.

buskamuza:
> `magento/project-community-edition` is merely a suggested starting template for a project 
Exactly. So we're adding the testing metapacakge to `reqiuire-dev` section of the suggested project template. If somebody doesn't want to use it, they can modify the `composer.json`. As soon as the project owner creates the project, `composer.json` is not under Magento's control anymore.
But it's a good place to show what exists natively.
In any case, I was just noticing that we can't use `require-dev` section in the "product" metapackage because it will not work. From there, we can either include or exclude tests from the project.

fooman:
@buskamuza I was hoping that this proposal here #157 was going ahead.

PR #178, opened 2019-06-04
Voting on Proposals

lenaorobei:
## Problem

Right now there is no mechanism to handle public proposals and reflect the actual community attendance. There is a way to submit proposals and participate in the discussion, but there is no way to participate in decision-making.
Many proposals produce conflict discussion and there is no established way to resolve it. Initiators are trying to resolve this with Twitter survey, but it doesn’t show who voted and allows any Twitter user to vote.

## Solution

This document represents the procedure to submit new proposals to the Magento community, vote and make decisions.

## Requested Reviewers

@maghamed 
@antonkril 
@okorshenko

### DIFF
--- a/decision-making/voting-on-proposals.md
+++ b/decision-making/voting-on-proposals.md
@@ -1,26 +1,26 @@
 # Voting on Proposals
-Architecture team leads proposal handling and is responsible for procedure compliance.
+The Architecture team handles the proposal workflow and is responsible for procedural compliance.
 
-### Proposal submission
-- Proposal is submitted by opening pull request to this repository.
-- Initiator must present the idea on the architecture meeting.
-- The announcement should include a problem the proposal is trying to resolve as well as solution(s).
+## Proposal submission
+- A proposal is submitted by opening a pull request to this repository.
+- The initiator must present the idea in an architecture meeting.
+- The announcement should include the problem the proposal is trying to resolve as well as any proposed solution(s).
 
-### Discussion and Voting
-- If there are controversial opinions in relation to the proposed idea architecture team may consider starting the voting.
-- The voting is considered open when `voting` label is added to the pull request.
+## Discussion and Voting
+- If there are controversial opinions in relation to the proposed idea, the architecture team may consider starting the voting.
+- The voting is considered open when the `voting` label is added to the pull request.
 - The voting can be represented as a single/multiple answer survey.
-- A valid voting period must be announced when the voting is started, must be greater that 1 week and must not be changed during the voting.
+- A valid voting period: 1) must be announced when the voting is started, 2) must be greater that 1 week and 3) must not be changed during the voting.
 - The Discussion may continue in the pull request thread.
 - Voters cannot change their vote during the voting period.
-- Intermediate results should be hidden until the voting period ends.
+- Results should be hidden until the voting period ends.
 
-### Results
+## Results
 - Voting results should be added to the corresponding pull request.
 - Participants should be mentioned by name.
 - When a majority vote exists for a proposal, it does not guarantee that the proposal will be adopted or implemented.
 
-# Eligible voters
+## Eligible voters
 All active GitHub users are treated as eligible voters.
 
 ## To Do
--- a/decision-making/voting-on-proposals.md
+++ b/decision-making/voting-on-proposals.md
@@ -1,0 +1,30 @@
+# Decision-making
+
+### Proposal submission
+Proposal is submitted by opening pull request to this repository.
+Initiator must present the idea on the architecture meeting.
+The announcement should include problem description the proposal is trying to resolve as well as solution(s).
+
+### Discussion
+There must be minimum 2 weeks of discussion in the pull request thread.
+After that initiator may decide when to start the vote.
+
+### Voting
+All eligible voters must get notification about the voting start.
+A valid voting period must be announced when voting is started, must be greater that 2 weeks and must not be changed during the vote.
+
+### Approval
+The number of `approve` votes must be greater than or equal to the number of `reject` votes multiplied by two in order to merge pull request.
+In case of multi-choice voting (implementation details) the decision may be taken by simple plurality. This means that the voting option with the most votes wins. If there are multiple options with the most number of votes, the initiator can choose one of them.
+Once pull request is merged the proposal is accepted.
+
+# Eligible voters
+The group of eligible voters should be established. The group should represent the Magento community and include internal developers and architects (40%), Magento maintainers (30%), extension developers and partner contributors (30%).
+
+## To Do
+- GitHub application for survey.
+- GitHub user group with eligible voters.
+
+_Inspired by:_
+- [Request for Comments: Voting on PHP features](https://wiki.php.net/rfc/voting)
+- [PHP-FIG Public Survey](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide-meta.md#44-public-survey)
--- a/decision-making/voting-on-proposals.md
+++ b/decision-making/voting-on-proposals.md
@@ -1,29 +1,30 @@
-# Decision-making
+# Voting on Proposals
+Architecture team leads proposal handling and is responsible for procedure compliance.
 
 ### Proposal submission
-Proposal is submitted by opening pull request to this repository.
-Initiator must present the idea on the architecture meeting.
-The announcement should include problem description the proposal is trying to resolve as well as solution(s).
+- Proposal is submitted by opening pull request to this repository.
+- Initiator must present the idea on the architecture meeting.
+- The announcement should include a problem the proposal is trying to resolve as well as solution(s).
 
-### Discussion
-There must be minimum 2 weeks of discussion in the pull request thread.
-After that initiator may decide when to start the vote.
+### Discussion and Voting
+- If there are controversial opinions in relation to the proposed idea architecture team may consider starting the voting.
+- The voting is considered open when `voting` label is added to the pull request.
+- The voting can be represented as a single/multiple answer survey.
+- A valid voting period must be announced when the voting is started, must be greater that 1 week and must not be changed during the voting.
+- The Discussion may continue in the pull request thread.
+- Voters cannot change their vote during the voting period.
+- Intermediate results should be hidden until the voting period ends.
 
-### Voting
-All eligible voters must get notification about the voting start.
-A valid voting period must be announced when voting is started, must be greater that 2 weeks and must not be changed during the vote.
-
-### Approval
-The number of `approve` votes must be greater than or equal to the number of `reject` votes multiplied by two in order to merge pull request.
-In case of multi-choice voting (implementation details) the decision may be taken by simple plurality. This means that the voting option with the most votes wins. If there are multiple options with the most number of votes, the initiator can choose one of them.
-Once pull request is merged the proposal is accepted.
+### Results
+- Voting results should be added to the corresponding pull request.
+- Participants should be mentioned by name.
+- When a majority vote exists for a proposal, it does not guarantee that the proposal will be adopted or implemented.
 
 # Eligible voters
-The group of eligible voters should be established. The group should represent the Magento community and include internal developers and architects (40%), Magento maintainers (30%), extension developers and partner contributors (30%).
+All active GitHub users are treated as eligible voters.
 
 ## To Do
-- GitHub application for survey.
-- GitHub user group with eligible voters.
+- Use [Community Portal](https://opensource.magento.com/) for surveys and vote tracking.
 
 _Inspired by:_
 - [Request for Comments: Voting on PHP features](https://wiki.php.net/rfc/voting)
--- a/decision-making/voting-on-proposals.md
+++ b/decision-making/voting-on-proposals.md
@@ -12,7 +12,7 @@ The Architecture team handles the proposal workflow and is responsible for proce
 - The voting can be represented as a single/multiple answer survey.
 - A valid voting period: 1) must be announced when the voting is started, 2) must be greater that 1 week and 3) must not be changed during the voting.
 - The Discussion may continue in the pull request thread.
-- Voters cannot change their vote during the voting period.
+- Voters can change their vote during the voting period.
 - Results should be hidden until the voting period ends.
 
 ## Results


joshuaadickerson:
Joshua Dickerson   [7 minutes ago]
An “aye” requires 2x “nays”? That’s extreme.


Joshua Dickerson   [6 minutes ago]
2/3rds votes are usually reserved for major changes.

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Discussion
+There must be minimum 2 weeks of discussion in the pull request thread.
+After that initiator may decide when to start the vote.
+
+### Voting
+All eligible voters must get notification about the voting start.
+A valid voting period must be announced when voting is started, must be greater that 2 weeks and must not be changed during the vote.
+
+### Approval
```
nathanjosiah: Are we ok with accepting proposals that have 100% opposition from Magento team but enough votes from community to qualify?

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
+# Decision-making
+
+### Proposal submission
+Proposal is submitted by opening pull request to this repository.
+Initiator must present the idea on the architecture meeting.
+The announcement should include problem description the proposal is trying to resolve as well as solution(s).
+
+### Discussion
+There must be minimum 2 weeks of discussion in the pull request thread.
```
nathanjosiah: Some proposals are created from internal Magento stories or bugs with deadlines. Two weeks blocking a whole team or initiative is a long time. Are there exceptions to this rule? Maybe a more important question, are the guidelines clear for when a proposal MUST be made versus when a simple code review from an architect is sufficient?

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
+# Decision-making
+
+### Proposal submission
+Proposal is submitted by opening pull request to this repository.
+Initiator must present the idea on the architecture meeting.
```
nathanjosiah: Does the initiator have to wait for the next meeting or is this requirement independent of the 2 week timeline given below?

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
...
+The announcement should include problem description the proposal is trying to resolve as well as solution(s).
+
+### Discussion
+There must be minimum 2 weeks of discussion in the pull request thread.
+After that initiator may decide when to start the vote.
+
+### Voting
+All eligible voters must get notification about the voting start.
+A valid voting period must be announced when voting is started, must be greater that 2 weeks and must not be changed during the vote.
```
nathanjosiah: Does this mean that every issue that is addressed from feedback resets the clock to 2 weeks?

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Approval
+The number of `approve` votes must be greater than or equal to the number of `reject` votes multiplied by two in order to merge pull request.
+In case of multi-choice voting (implementation details) the decision may be taken by simple plurality. This means that the voting option with the most votes wins. If there are multiple options with the most number of votes, the initiator can choose one of them.
+Once pull request is merged the proposal is accepted.
+
+# Eligible voters
+The group of eligible voters should be established. The group should represent the Magento community and include internal developers and architects (40%), Magento maintainers (30%), extension developers and partner contributors (30%).
```
nathanjosiah: Is there a requirement for how many votes are needed or is the number of votes required determined by the number of eligible people in the voter pool? Also, is there a requirement for percentage of votes from a particular category?

decision-making/voting-on-proposals.md
lenaorobei: I would say, initiator must present the idea and give 2 weeks discussion time. Need to work on wording.

decision-making/voting-on-proposals.md
lenaorobei: That's a really good point I would like to discuss. Maybe architecture team should initiate voting, not author.

decision-making/voting-on-proposals.md
lenaorobei: Hm, what I meant is when voting is started there should be no changes to the proposal.

decision-making/voting-on-proposals.md
lenaorobei: From my point of view - yes, but I need more thoughts and concerns here.

decision-making/voting-on-proposals.md
lenaorobei: Will add this info as well. This is the rough draft I wrote to understand if we want to move in this direction. It definitely requires defining all rules and use cases.

ishakhsuvarov: We might want to have some kind of rotation in "eligible voters" groups, otherwise there is a risk of forming a group biased to specific questions

decision-making/voting-on-proposals.md
ishakhsuvarov: More sophisticated model with someone having the power of veto may be implemented.

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
...
+### Discussion
+There must be minimum 2 weeks of discussion in the pull request thread.
+After that initiator may decide when to start the vote.
+
+### Voting
+All eligible voters must get notification about the voting start.
+A valid voting period must be announced when voting is started, must be greater that 2 weeks and must not be changed during the vote.
+
+### Approval
+The number of `approve` votes must be greater than or equal to the number of `reject` votes multiplied by two in order to merge pull request.
```
ishakhsuvarov: Maybe it's just me but wording seems complicated. Should we just stick to 2/3?

decision-making/voting-on-proposals.md
maghamed: Architecture team should initiate voting as for me, as some proposals may be misleading or propose the idea for which we already have resolution

decision-making/voting-on-proposals.md
maghamed: Don't think that we need to overcomplicate the voting process with veto right. I think every vote should have the same power.
For me, it's hard to imagine a situation when a proposal will get 100% opposition from Magento. It has never happened before, why should be foreseen in the future? 
Even internally for many questions, we don't have a single opinion.

decision-making/voting-on-proposals.md
maghamed: I would not introduce tight coupling to the maintainers group establishing some proportion, because sometimes we have contributors who are not maintainers but actively participating in Magento Development process, there are other influencers who do not maintain Magento either.

This supposed to be a new role - Community Architect.

Magento team should be also represented widely, for example, there should be people from Cloud and MOM.

The proportion may vary, thus I would like not to mention it here at all, and we may extend the group of eligible voters with both internal and external participants. We should be flexible enough here.

decision-making/voting-on-proposals.md
heiglandreas: Why does it need to be presented in a separate meeting? Or perhaps what types of changes need to go via this route and what kinf of changes can "just" go as PRs?

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
+# Decision-making
+
+### Proposal submission
+Proposal is submitted by opening pull request to this repository.
+Initiator must present the idea on the architecture meeting.
+The announcement should include problem description the proposal is trying to resolve as well as solution(s).
```
heiglandreas: So does the proposal actually need to contain code? Or descriptions of the solution? A PR template might be helpful to not forget things

decision-making/voting-on-proposals.md
heiglandreas: So the "shall we introduce this" vote needs 2/3 and possible "how shall we introduce this" need majority. so with 3 choices 1/3+1 might be sufficient..

decision-making/voting-on-proposals.md
heiglandreas: Is there a minimum threshold of cast votes from the voters pool? so at least 50% of people eligible to vote need to vote otherwise the vote will... fail? need to be done again? will a failed vote need to be rewritten? or can it be voted on again immediately?

decision-making/voting-on-proposals.md
benmarks: *I fully support the initiative of public voting on proposals and will provide further input on design, goals, etc.* I am commenting on this facet separately for sake of expediency.

**A veto mechanism is neither necessary nor appropriate.** Whereas Magento, an Adobe company (MAAC) _alone_ releases the product we all call "Magento", I propose to add a note which recognizes the following reality:

> When a majority vote exists for a proposal, it does not guarantee that the proposal will be adopted or implemented.

This approach allows for two important realties:

1. To commemorate the collective opinion of the balloters in the face of disagreement with Magento/Adobe strategic needs
2. To allow Magento/Adobe to make even-better-informed product decisions based on our business needs.

As an addendum to (2), I believe this will force MAAC to be aware of & to justify product decisions which are in contrast to the wishes of the our internal & external ecosystem members, as proxied by balloters.

decision-making/voting-on-proposals.md
joshuaadickerson: aka, what is quorum?

joni-jones: > The group should represent the Magento community and include internal developers and architects (40%)

We already have architects assigned to the components.

lenaorobei: Yes, it's a known fact. How do you propose to use those assignments?

lenaorobei:
**Open questions for me are:**

1. After the discussion with @okorshenko it was proposed to treat all GitHub @magento organization users as eligible voters and assign different vote weights depending on user group, e.g. 

| Group | Vote weight |
| ------------- | ------------- |
| Architects  | x5 ? |
| Internal devs & community maintainers  | x3  ?|
| Partners  | x2 ?  |
| Community contributors  | x1 ? |

2. Quorum. Do we need it?

nathanjosiah:
I think having a quorum is important or else this PR seems mostly irrelevant. Without a quorum a PR would basically be automatically accepted so long as it has a single vote (from anybody) after two weeks which is really not any different than we are already doing.

decision-making/voting-on-proposals.md
joni-jones: Actually, it contradicts with the proposal: `The group should represent the Magento community and include internal developers and architects (40%)` because if all architects are included to the group, then the group will be too big otherwise the assigned architect might miss proposals for related components.

nathanjosiah:
To provide another view on how these types of things are handled at a larger scale, the IETF published their process for approving standards [in an RFC](https://tools.ietf.org/rfc/rfc7282.txt). This document is directly relevant to the issue trying to be addressed in this PR. 

One thing that I think is very interesting about their process is they establish that they are against voting and a 100 to 1 vote can still potentially not be a correct answer. Very interesting read and something definitely worth thinking about before this PR is merged.

buskamuza:
1. I propose to update Readme file instead of creating a separate document. The process is already described there. If we want to improve it, I propose to modify existing document, otherwise it'll lead to confusion.
2. There is some process currently. We use reviews. So you can see who agrees and who disagrees in the reviewers list. But the list is usually limited to corresponding architects based on the components. As I understand the proposal includes expanding the list to developers and community members.

decision-making/voting-on-proposals.md
buskamuza: PR is a document or a set of documents. There is a template with suggested structure at https://github.com/magento/architecture/blob/master/design-documents/README.md

joshuaadickerson:
> 1. Quorum. Do we need it?
Yes, but it should be drawn from the Magento architects and not from the entire voting group, unless that voting group is small and then the quorum would be small to match it.

lenaorobei:
Summary after the discussion with @buskamuza @maghamed and @okorshenko:
- Architecture team should initiate and process the voting.
- The voting serves for informational purposes only (no need a quorum).
- Architecture team reserves the right to decline proposals with a majority vote.
- All GitHub users should be able to vote.
- Votes should be tracked by name.

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
+# Voting on Proposals
+Architecture team leads proposal handling and is responsible for procedure compliance.
```
dobooth:
```suggestion
The Architecture team handles the proposal workflow and is responsible for procedural compliance.
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Proposal submission
```
dobooth:
```suggestion
### Proposal submission workflow
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- Proposal is submitted by opening pull request to this repository.
```
dobooth:
```suggestion
- A proposal is submitted by opening a pull request to this repository.
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- Initiator must present the idea on the architecture meeting.
```
dobooth:
```suggestion
- The initiator must present the idea in an architecture meeting.
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- The announcement should include a problem the proposal is trying to resolve as well as solution(s).
```
dobooth:
```suggestion
- The announcement should include the problem the proposal is trying to resolve as well as any proposed solution(s).
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Discussion and Voting
+- If there are controversial opinions in relation to the proposed idea architecture team may consider starting the voting.
```
dobooth:
```suggestion
- If there are controversial opinions in relation to the proposed idea, the architecture team may consider starting the voting.
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- The voting is considered open when `voting` label is added to the pull request.
```
dobooth:
```suggestion
- The voting is considered open when the `voting` label is added to the pull request.
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- The voting can be represented as a single/multiple answer survey.
+- A valid voting period must be announced when the voting is started, must be greater that 1 week and must not be changed during the voting.
```
dobooth:
```suggestion
- A valid voting period: 1) must be announced when the voting is started, 2) must be greater that 1 week and 3) must not be changed during the voting.
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- The Discussion may continue in the pull request thread.
+- Voters cannot change their vote during the voting period.
+- Intermediate results should be hidden until the voting period ends.
```
dobooth:
```suggestion
- Results should be hidden until the voting period ends.
```

decision-making/voting-on-proposals.md
dobooth:
```suggestion
## Proposal submission
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
+# Voting on Proposals
+Architecture team leads proposal handling and is responsible for procedure compliance.
+
+### Proposal submission
+- Proposal is submitted by opening pull request to this repository.
+- Initiator must present the idea on the architecture meeting.
+- The announcement should include a problem the proposal is trying to resolve as well as solution(s).
+
+### Discussion and Voting
```
dobooth:
```suggestion
## Discussion and Voting
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- If there are controversial opinions in relation to the proposed idea architecture team may consider starting the voting.
+- The voting is considered open when `voting` label is added to the pull request.
+- The voting can be represented as a single/multiple answer survey.
+- A valid voting period must be announced when the voting is started, must be greater that 1 week and must not be changed during the voting.
+- The Discussion may continue in the pull request thread.
+- Voters cannot change their vote during the voting period.
+- Intermediate results should be hidden until the voting period ends.
+
+### Results
```
dobooth:
```suggestion
## Results
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- Voting results should be added to the corresponding pull request.
+- Participants should be mentioned by name.
+- When a majority vote exists for a proposal, it does not guarantee that the proposal will be adopted or implemented.
+
+# Eligible voters
```
dobooth:
```suggestion
## Eligible voters
```

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,31 @@
...
+- The initiator must present the idea in an architecture meeting.
+- The announcement should include the problem the proposal is trying to resolve as well as any proposed solution(s).
+
+## Discussion and Voting
+- If there are controversial opinions in relation to the proposed idea, the architecture team may consider starting the voting.
+- The voting is considered open when the `voting` label is added to the pull request.
+- The voting can be represented as a single/multiple answer survey.
+- A valid voting period: 1) must be announced when the voting is started, 2) must be greater that 1 week and 3) must not be changed during the voting.
+- The Discussion may continue in the pull request thread.
+- Voters cannot change their vote during the voting period.
```
buskamuza: Why not allow changing votes?

lenaorobei: My point was that it will lead to the carefully considered decision and more responsible attitude to the voting.

nathanjosiah: I've had my mind and vote changed many times after some good discussions. I think the idea you are trying to address is valid but I think there may be some validity to changing it as well

buskamuza: I'm not a psychologist, so maybe this is how people behave. So I'd just vote here to allow changing votes :) People may:
1. not notice that votes are not changeable until they really voted
1. click a wrong answer accidentally
1. vote and then see a discussion going in the PR (we still allow the discussion to go while voting is already in progress) and change their mind (thoughtfully)

lenaorobei: Makes sense. Updated this line.

decision-making/voting-on-proposals.md
```diff
@@ -0,0 +1,30 @@
+# Decision-making
+
+### Proposal submission
+Proposal is submitted by opening pull request to this repository.
```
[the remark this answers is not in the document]
lenaorobei: PR contains proposal document (`.md` file).
Keeping it on GitHub allows to track document changes during the discussion thread.

PR #179, opened 2019-06-04
Advanced JS bundling proposal

vzabaznov:
## Problem

We have 180 HTTP requests to fetch all needed JS code just for empty Home Page and overall JS size is 2.2 MB. According to google - mobile user with slow connection can have up to 8sec main-thread work time, bad pagespeed insight score, and over 21sec time to interactive.

## Solution

Better JS bundling approach can reduce package size by several times. This is one of the most valuable enhancement can be done from storefront side performance perspective. According to our measurements, advanced bundling can improve overall performance score, main-thread work time and TTI up to 2-3x time.

## Requested Reviewers
@kandy 
@DrewML 
@melnikovi


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=179) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=179) before we can accept your contribution.<br/>

DrewML:
These are my docs from the unreleased internal tool, right? Just a friendly reminder that attribution for work is typically appreciated :)

krzksz:
Since we pretty much run out of any further possibilities to improve the frontend speed of our shops apart from improving JavaScript delivery from Magento I would really much like to make this happen and leave some potentially valuable feedback.

What we already do in our stores is using webpack for all of the assets (TypeScript and Sass based) along with automatic code splitting that are then loaded depending on which type of the page is loaded. Below you can see a typical output of production compilation that we have: 
```
                             Asset      Size  Chunks                    Chunk Names
                    js/category.js  40.8 KiB       4  [emitted]         category
                    js/checkout.js  36.3 KiB       5  [emitted]         checkout
                         js/cms.js  32.9 KiB       6  [emitted]         cms
                     js/commons.js  77.4 KiB       0  [emitted]         commons
                    js/customer.js  16.1 KiB       8  [emitted]         customer
                         js/pdp.js  45.1 KiB      11  [emitted]         pdp
```
As you may have already noticed, the `js/commons.js` chunk is created automatically out of every module that exists in all of the other groups. Those groups cover following areas of the shop:
- Category:
  - Category, search results and brand overview pages.
- Checkout:
  - Cart and checkout pages.
- CMS:
  - CMS pages e.g. homepage.
- Customer:
  - Login, register, password forgotten and user area pages.
- PDP:
  - Product detail page.

For now it seems like those are sufficient, but still our configuration allows us to add any additional groups needed in a certain shop.

Let this be a preliminary to the questions I have for the proposal you submitted.

design-documents/frontend/advanced-js-bundling/advanced-js-bundling.md
```diff
@@ -0,0 +1,82 @@
...
+## Problems
+
+We have 180 HTTP requests to fetch all needed JS code just for empty Home Page and overall JS size is 2.2 MB. According to google - mobile user with slow connection can have up to 8sec main-thread work time, bad pagespeed insight score, and over 21sec time to interactive.
+
+Better JS bundling approach can reduce package size by several times. This is one of the most valuable enhancement can be done from storefront side performance perspective. According to our measurements, advanced bundling can improve overall performance score, main-thread work time and TTI up to 2-3x time.
+
+## Solution
+
+I propose to introduce 'Bundlegento' as separate tool for parsing Magento store pages and generate page specific bundles.
+Currently we have prototype written in NodeJS that parse HTML pages and can generate needed bundles for Magento store. Since that this is not version specific solution, it will support all Magento released versions
```
krzksz: The only pitfall I can see is the possibility that the development and production environments may be in totally different states. This may lead to the situation when module is requested on the former but is not on the latter or otherwise. This shouldn't be a big problem though as the potential overhead should not be to big and requirejs should be able to load any missing modules at runtime as usual.

design-documents/frontend/advanced-js-bundling/advanced-js-bundling.md
```diff
@@ -0,0 +1,82 @@
...
+
+Also we will need to add needed functionality inside Magento that will load needed JS bundle according to layout handlers was used. JS bundle files will be added under git track and stored in theme specific location
+
```
krzksz: I am not sure if there is any benefit in this approach at it is generally a bad practice to commit any automatically generated files as this may introduce a lot of noise in commits and merge requests.

Do I assume correctly that this is done to avoid the need of running this step during a build step of deploy on a CI? If so, then maybe storing only the list of modules in every group would be enough? This would even open up a potential possibility to fiddle with those settings like adding extra modules not caught by the tool or excluding any huge ones to separate files.

design-documents/frontend/advanced-js-bundling/advanced-js-bundling.md
```diff
@@ -0,0 +1,82 @@
...
+### Splitting
+
+With module usage data obtained during the fetching phase, modules can now be compared between groups to determine which are unique to a page, and which should be broken out to shared bundles.
+
+The algorithm for splitting modules is fairly basic:
+
+-   If a module exists in exactly 1 group, keep it in that group's bundle
+-   If a module exists in exactly 2 groups, move it to a shared file that is loaded for any page in those 2 groups
+-   If a module exists in >= 3 groups, move it to a shared bundle that will be loaded on all pages
```
krzksz: Taken that I can totally understand the reasoning behind common bundle, getting the grasp of shared ones is difficult.

Would it be true that if we have 5 groups we would end up with 10 shared bundles, 15 for 6, 21 for 7 etc.?

Is so, then I'm afraid it would negatively impact the performance across the pages as opposed to having one common bundle loaded from cache for every other group and one extra file loaded specific to each of them.

PR #182, opened 2019-06-05
Using scalars for E-mail template variables

AlexMaxHorkun:
## Problem

Unexpected behavior and code execution by regular users in E-mail templates

## Solution

Limit template variables to scalars and arrays of scalars, covert helpers like $this->getUrl() to directives

## Requested Reviewers

@paliarush @melnikovi @rganin


design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,154 @@
...
+and E-mails are presentation. To make matters worse - administrators can make their own templates
+through admin area and thus being able to execute any code on the server.
+
+#### The solution
+To prevent these problems from happening we must force using prepared data in
+E-mail templates - scalars received by calling certain model layer services.
+Thus templates will only receive data without the possibility to execute any code.
+
+#### What needs to be done
+###### Allow access to arrays' values
```
zetlen: How would a content author iterate through an indexed array rather than an associative array? For instance, to render a list of order items?

AlexMaxHorkun: we have 'for' directive for that purpose

AlexMaxHorkun:
> We need clear migration path.
> 
> Configurable whitelist of the allowed methods is an alternative approach. The whitelist can be managed by SIs or privileged admin users.

Added description of object/method whitelisting

design-documents/scalars-for-templates.md
```diff
@@ -143,41 +153,20 @@ interface FilterProcessorInterface
...
+###### Deprecate old methods
+Old public methods following _\<Directive Name\>Directive()_ pattern must be deprecated in favor of new directive
+processors.
 
 #### Backward compatibility
 This would not be a backward compatible change - we use objects in our own
 E-mail templates in core Magento and developers are using objects in their own templates
 as well.
+
+In order to preserve backward compatibility E-mail templates read from _.html_ files will keep workings
```
paliarush: What will be the deprecation strategy? How do we stop supporting legacy style?

AlexMaxHorkun: added a bit about notifying developers about new restrictions and keeping old variables accessible for existing templates. Also the documents states in the end that support for legacy templates will be removed next minor version after these restrictions are introduced (2.5?)

design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,177 @@
...
+###### Deprecate old methods
+Old public methods following _\<Directive Name\>Directive()_ pattern must be deprecated in favor of new directive
+processors.
+
+#### Backward compatibility
+This would not be a backward compatible change - we use objects in our own
+E-mail templates in core Magento and developers are using objects in their own templates
+as well.
+
+In order to preserve backward compatibility E-mail templates read from _.html_ files will keep working
```
melnikovi: What to do with templates saved to database?

AlexMaxHorkun: I mention this in the document below this statement - render them as they were. I was thinking about adding a flag to newly created/updated templates to render them with compliance to new restrictions

design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,177 @@
...
+     * @return string
+     */
+    public function process(?string $value, ?string $html, array $parameters): string;
+    
+    /**
+     * Subdirectories that can be used within this directory's HTML.
+     * 
+     * @return DirectiveProcessorInterface[]|null
+     */
+    public function getSubDirectives(): ?array;
```
melnikovi: Do we need to limit directives inside a directive? Can we just process recursively? Could you provide some examples, please?

AlexMaxHorkun: 'else' directive would only be valid inside an 'if' directive

melnikovi:
Per offline discussion please revise `DirectiveProcessorInterface` and add more info on backwards compatibility.

fooman:
This proposal is aimed at changing the Email behaviour only. However one of the suggested changes is affecting `Magento\Framework\Filter\Template` which is used beyond emails (a quick search yielded CMS and Catalog, plus we use it in our own extension). As it's currently proposed I believe this might break more than anticipated and if this is about improving emails the changes should be limited to that.

AlexMaxHorkun:
> This proposal is aimed at changing the Email behaviour only. However one of the suggested changes is affecting `Magento\Framework\Filter\Template` which is used beyond emails (a quick search yielded CMS and Catalog, plus we use it in our own extension). As it's currently proposed I believe this might break more than anticipated and if this is about improving emails the changes should be limited to that.

that is a good point and we'll make sure the proposed limitations will only be active for E-mail templates when implementing this design doc.

design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,194 @@
...
+ 
+When we update core E-mail templates to use scalars it is important to introduce new scalar variables replacing object
+variables with new names so that old templates using objects can remain functional. For instance the _account\_new.html_
+expects _\$customer_ variable to be a _Customer_ object so we would add _\$customer_data_ variables containing an _array_
+with a customer's data so if a merchant was using customized _account\_new_ template it will still render fine.
+ 
+We have to keep existing user-defined E-mail templates rendering as before to preserve backward compatibility. Newly-added
+templates will be marked appropriately in the DB and rendered with new limitation regarding template variables though.
+ 
+These measures aimed to improve backcompat are to be removed in a later minor version disallowing using objects
```
paliarush: Please use full phrase for clarity.

design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,194 @@
...
+This forces developers to create child classes of _Template_ in order to introduce new directives or modify existing ones.
+This way of working with directives has to be deprecated but preserved in order to keep backward compatibility.
+In the next minor version we will rewrite all existing directives using
+_DirectiveProcessorInterface_ and _SimpleDirectiveProcessorInterface_. In the meantime only new directives will be
+created using this SPI.
+ 
+We may create a dev-docs page explaining developers how to introduce directives and provide examples by creating new
+directives in order to replace _\$this.method()\_ usages inside E-mail templates (like using $this.getUrl() method).
+
+#### Backward compatibility
```
paliarush: What is the background behind this proposal? We need to have strong practical reasoning to introduce such changes. Have anyone been affected by described vulnerabilities in the past, please provide references.

design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,194 @@
...
+```
+ 
+_Template_ class would receive a list of such filters and use them accordingly.
+ 
+###### Old directive processors
+Currently directives are being processed by methods of _Template_ class following _\<Directive Name\>Directive()_ pattern.
+This forces developers to create child classes of _Template_ in order to introduce new directives or modify existing ones.
+This way of working with directives has to be deprecated but preserved in order to keep backward compatibility.
+In the next minor version we will rewrite all existing directives using
+_DirectiveProcessorInterface_ and _SimpleDirectiveProcessorInterface_. In the meantime only new directives will be
```
melnikovi: Need to rename SimpleDirectiveProcessorInterface to something else as it will be used by DirectiveProcessorInterface. Do you have any suggestions @AlexMaxHorkun? Also please describe it.

AlexMaxHorkun: @melnikovi why do we need to rename it?

design-documents/scalars-for-templates.md
```diff
@@ -0,0 +1,194 @@
...
+```
+_Magento\Framework\Filter\Template_ does not allow that right now so
+it's _getVariable()_ method must be updated to allow just that.
+ 
+To make these changes easier to adopt
+```
+{{var customer.getName() }}
+```
+ 
+Will be equal to accessing the 'customer' array by 'name' key.
```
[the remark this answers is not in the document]
AlexMaxHorkun: we don't have that implemented in Magento and for developers using other DTO implementations not extending DataObject this will not work anyways

design-documents/scalars-for-templates.md
AlexMaxHorkun: We may decide on a different directive processor interface name during implementation

PR #186, opened 2019-06-10
Binding variables to HTML in knockout/Magento templates

AlexMaxHorkun:
## Problem

Developers always using HTML binding instead of mixing text/HTML binding accordingly leaves XSS vulnerabilities

## Solution

Require variables/functions used for HTML binding to have 'html' syntax to indicate proper sanitization and conscious choice to use HTML binding

## Requested Reviewers

@melnikovi @joni-jones

### DIFF
--- a/design-documents/frontend-html-binding.md
+++ b/design-documents/frontend-html-binding.md
@@ -12,7 +12,7 @@ Developers may use this binding instead of text binding for all values leaving X
 vulnerabilities.
 
 #### Proposing
-Create static test to require developers to use variables/functions with _'html'_ suffix
-to bind HTML (like _variableHtml_ or _widget.generateButtonHtml()_). It's not a 100% guaranty that properly sanitized values will be used for dynamic HTML
+Create static test to require developers to use variables/functions with _'UnsanitizedHtml'_ suffix
+to bind HTML (like _variableUnsanitizedHtml_ or _widget.generateButtonUnsanitizedHtml()_). It's not a 100% guaranty that properly sanitized values will be used for dynamic HTML
 but it will notify developers to think whether they really need HTML binding for a value or
-maybe text binding will be enough. This rule would be consistent with the rule we have for _.phtml_ templates.
+maybe text binding will be enough. This rule would be more-or-less consistent with the rule we have for _.phtml_ templates.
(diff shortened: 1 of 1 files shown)


design-documents/frontend-html-binding.md
```diff
@@ -0,0 +1,18 @@
...
+<!-- Magento syntax -->
+<div html="variable"></div>
+```
+Developers may use this binding instead of text binding for all values leaving XSS
+vulnerabilities.
+
+#### Proposing
+Create static test to require developers to use variables/functions with _'html'_ suffix
+to bind HTML (like _variableHtml_ or _widget.generateButtonHtml()_). It's not a 100% guaranty that properly sanitized values will be used for dynamic HTML
+but it will notify developers to think whether they really need HTML binding for a value or
```
joni-jones: In most cases, 3rd party developers don't run static tests, so for them, it won't work.

design-documents/frontend-html-binding.md
```diff
@@ -0,0 +1,18 @@
...
+<!-- Knockout syntax -->
+<div data-bind="html: variable"></div>
+<!-- Magento syntax -->
+<div html="variable"></div>
+```
+Developers may use this binding instead of text binding for all values leaving XSS
+vulnerabilities.
+
+#### Proposing
+Create static test to require developers to use variables/functions with _'html'_ suffix
```
joni-jones: How the test will recognize if `_html_` should be present or not? Or will it always fail if `data-bind:html` is present and the variable/method doesn't contain `_html_` in the name?

design-documents/frontend-html-binding.md
AlexMaxHorkun: extension developers and community will run the tests - I think it's better than nothing

design-documents/frontend-html-binding.md
AlexMaxHorkun: always fail

AlexMaxHorkun:
> This does not solve any problem actually.
> 
> The only viable approach is "Escape All By Default" and make exceptions only in rare cases when reallt needed.

I don't think it would make sense to always escape values explicitly binded as html, would be counter-intuitive

orlangur:
@AlexMaxHorkun,

>  it would make sense to always escape values explicitly binded as html

I didn't say that. What I said is that those `text/html` bindings flawed and ideally we should replace KnockoutJS templates with something better, like Handlebars.

PR #188, opened 2019-06-12
Support JWT out-of-box

joni-jones:
## Problem

Magento provides a possibility to communicate with different 3rd party systems, some of them use JWT for data verification and, now, we don't have any standard mechanism to generate/parse/verify JWT.

## Solution

Use one of the PHP packages for JWT and create wrappers for the code unification.


AlexMaxHorkun:
@joni-jones I don't think we have to add another 3rd party dependency to Magento just because of these tokens. It's extremely easy to implement this feature ourselves or even our own kind of tokens. The main purpose of JWT is to allow server authenticate and authorize clients without checking the database, or at least partially trust some information coming from a client. So we use a string with serialized data and sign it on the server. It would be easier to unify this without a library I think - we can use our own EncryptorInterface::hash() to do the signing and serialize only data we actually want. Like we don't need token issued timestamp - we store it in the DB anyways and will load it on token validation. We could use these tokens to store some other user-specific data though and allow extensions to add some values of their own

joni-jones:
> It's extremely easy to implement this feature ourselves or even our own kind of tokens. 

Why do we need to implement all wrappers on hashing/encryption algorithms, claims checkers, parsing and signature validation as all this stuff already implemented?

> The main purpose of JWT is to allow server authenticate and authorize clients without checking the database, or at least partially trust some information coming from a client. So we use a string with serialized data and sign it on the server.

As I've described in the proposal, JWT it's not only about authentication/authorization, it also used for data verification. Different 3rd party integrations (like Cardinal Commerce for 3d secure) use JWT as a mechanism to guarantee that data wasn't changed. So, in general, it will be used not only to provide a possibility of token authentication/authorization usage inside the scope of Magento functionality but also for Magento extensions/customizations to integrate with 3rd party systems which use JWT. That's why we need to support a wide list of different hashing/encryption algorithms, claim checkers, etc.

>  we can use our own EncryptorInterface::hash() to do the signing and serialize only data we actually want. ... We could use these tokens to store some other user-specific data though and allow extensions to add some values of their own

This approach doesn't unify a solution but makes it specific to the Magento platform.

design-documents/jwt-support.md
```diff
@@ -0,0 +1,180 @@
...
+
+The following diagram represents needed interfaces to unify the workflow for JWT usage.
+
+![Class diagram](img/jwt-class-diagram.png)
+
+The `\Magento\Framework\Jwt\KeyGeneratorInterface` will provide a possibility to use different types of key generators like: Magento deployment secret key, X.509 certificates or RSA keys.
+```php
+interface KeyGeneratorInterface
+{
+    public function create(): JWK;
```
AlexMaxHorkun: Is JWK a Magento class? If it's from the library then we should not expose it

design-documents/jwt-support.md
```diff
@@ -0,0 +1,180 @@
...
+    ) {
+        $this->algorithmFactory = $algorithmFactory;
+        $this->keyGenerator = $keyGenerator;
+        $this->serializer = $serializer;
+    }
+
+    public function generate(array $claims = []): string
+    {
+        $timestamp = time();
+        $baseClaims = [
```
AlexMaxHorkun: are these claims mandatory by the RFC? What if someone doesn't want them in their tokens?

design-documents/jwt-support.md
```diff
@@ -0,0 +1,180 @@
...
+    {
+        $verifier = $this->getVerifier();
+        $jws = $this->serializer->unserialize($token);
+
+        if (!$verifier->verifyWithKey($jws, $this->keyGenerator->create(), 0)) {
+            return false;
+        };
+
+        $checkers = ['iat', 'nbf', 'exp'];
+        $claimChecker = $this->checkerManagerFactory->add('iat', new IssuedAtChecker())
```
AlexMaxHorkun: there seem to be no way for 3rd-party developers to use Magento SPI to modify claims

design-documents/jwt-support.md
```diff
@@ -0,0 +1,180 @@
...
+
+class CryptKeyGenerator implements KeyGeneratorInterface
+{
+    public function __construct(AlgorithmFactory $algorithmFactory, DeploymentConfig $deploymentConfig)
+    {
+        $this->algorithmFactory = $algorithmFactory;
+        $this->deploymentConfig = $deploymentConfig;
+    }
+
+    public function create(): JWK
```
AlexMaxHorkun: I think it would be better if default key generator would utilize EncryptorInterface - we always update it to follow the most recent security requirements

design-documents/jwt-support.md
joni-jones: Now, it's class from the library but it should be wrapped by Magento class.

design-documents/jwt-support.md
joni-jones: RFC doesn't describe mandatory claims, it describes only reserved claims. This is just an example from POC, the claims should be configurable.

design-documents/jwt-support.md
joni-jones: They are checkers, not claims. The checkers also should be configurable, it just an example from POC.

design-documents/jwt-support.md
joni-jones: EncryptorInterface doesn't allow to work with 3rd party systems because they can use the different algorithms to generate the key (which are not supported by EncryptorInterface).

AlexMaxHorkun: Like I've said, I would like to see it being used by default signature generator only. Developers should be able to configure other algorithms

design-documents/jwt-support.md
AlexMaxHorkun: I am just worried that we would create wrappers by just extending these library classes and duplicating the library's API like it's done with the graphQL library wrappers which would be wrong. DO Magento developers really need JWK? How about having sign(string $data): string method instead?

joni-jones: The idea is not to duplicate the functionality of a library but to provide a convenient approach for JWT usage.

I need to work a little bit more on JWK functionality because @kokoc mentioned that key sets are useful for webhooks.

design-documents/jwt-support.md
joni-jones: Updated

design-documents/jwt-support.md
joni-jones: Updated

design-documents/jwt-support.md
```diff
@@ -0,0 +1,246 @@
...
+        // as payload represented by url encode64 on json string,
+        // the same claims structure with different key's order will get different payload hash
+        ksort($claims);
+        $payload = $this->json->serialize($claims);
+
+        $jwsBuilder = new JWSBuilder($this->algorithmFactory->getAlgorithmManager());
+        $jws = $jwsBuilder->create()
+            ->withPayload($payload)
+            ->addSignature(
+                $this->keyGenerator->create()->getKey(),
```
kokoc: `$this->keyGenerator->create()->getKey()` looks strange. I thought keys should be relatively stable, why do we need factory-like class here?

design-documents/jwt-support.md
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    public function encode(array $claims): string
+    {
+        // as payload represented by url encode64 on json string,
+        // the same claims structure with different key's order will get different payload hash
+        ksort($claims);
+        $payload = $this->json->serialize($claims);
+
+        $jwsBuilder = new JWSBuilder($this->algorithmFactory->getAlgorithmManager());
```
kokoc: The creational pattern looks strange here. Is it possible to use shared objects which are injected using DI?

design-documents/jwt-support.md
```diff
@@ -0,0 +1,246 @@
...
+
+The main use cases of JWT usage:
+
+#### User authorization
+
+![User authorization](img/jwt-user-authorization.png)
+
+JWT contains user's "session" representation. After successful login, all user session data (like ID, permissions, etc.) is stored in JWT token and user's permissions and allowed operations can be verified just based on information from JWT.
+
+#### API authorization
```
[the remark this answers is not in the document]
joni-jones: The Magento secret is just a base implementation. We can use different approaches to generate keys like RSA pair, X.509 Certificate, etc. For example, CardinalCommerce integration uses API Key as secret https://github.com/magento/magento2/pull/23874/files#diff-008edfb30d2f4d84f29ff06499f5f393R42-R46.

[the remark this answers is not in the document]
joni-jones: The Magento secret is just a base implementation. We can use different approaches to generate keys like RSA pair, X.509 Certificate, etc. For example, CardinalCommerce integration uses API Key as secret https://github.com/magento/magento2/pull/23874/files#diff-008edfb30d2f4d84f29ff06499f5f393R42-R46.

design-documents/jwt-support.md
joni-jones: The `\Magento\Framework\Jwt\KeyGeneratorInterface::create` returns the `\Magento\Framework\Jwt\Data\Jwk` which is a wrapper for key management from the 3rd party library. This wrapper allows avoiding the direct dependency from modules which might have own key creation mechanism, like CardinalCommerce, to the 3rd party library.

design-documents/jwt-support.md
```diff
@@ -0,0 +1,246 @@
...
+                    'alg' => $this->algorithmFactory->getAlgorithmName(),
+                    'typ' => 'JWT'
+                ]
+            )
+            ->build();
+
+        return $this->serializer->serialize(new Jwt($jws));
+    }
+
+    public function decode(string $token): array
```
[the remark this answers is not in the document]
joni-jones: Yes, probably makes sense.

design-documents/jwt-support.md
joni-jones: The `JWSBuilder` it's actually the class provided by 3rd party library which requires algorithm as a dependency during object initialization:
```php
__construct(AlgorithmManager $signatureAlgorithmManager)
```
In POC implementation, the default algorithm is provided via di.xml:
```xml
<type name="Magento\Framework\Jwt\AlgorithmFactory">
    <arguments>
        <argument name="algorithm" xsi:type="object">Jose\Component\Signature\Algorithm\HS256</argument>
    </arguments>
</type>
```
https://github.com/magento/magento2/pull/23874/files#diff-77f16aba9febcc1731c41d65797402e1R1777-R1781

We also can configure the default algorithm via di.xml for `JWSBuilder` and inject it using DI. But in a case, if some module/custom implementation needs another algorithm like sha512 it should specify it not for our core implementation but directly for `JWSBuilder` - it will introduce a dependency from module/custom implementation to 3rd party library what we don't recommend to do.

design-documents/jwt-support.md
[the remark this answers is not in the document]
joni-jones: UPD: distributed applications should not use shared secret key.

PR #189, opened 2019-06-12
Create RefcatoringForStateCopling.md

larsroettig:

### DIFF
--- a/design-documents/RefcatoringForStateCopling.md
+++ b/design-documents/RefcatoringForStateCopling.md
@@ -1,0 +1,101 @@
+# Independent entities for order state
+
+The order state is currently saved in `sales_order` database but a state should be have and event history also it must be not cupeled to an order. 
+
+#### Common Problems
+
+This section describes common problems which affect the current flow of order entity manipulations.
+
+1. No possibility to update only order state or status without updating the whole entity.
+2. An order's life circle depends on its state which can be updated in an inconsistent way.
+3. As the state|status can be updated with the whole order and all related data it has more performance impact.
+4. Payment integrations use asynchronous operations, they can't update order state|status.
+5. Overloaded tables: *sales_order* - 206 columns (EE), *sales_invoice* - 66 (EE), *sales_creditmemo* - 73 (EE).
+6. No possibility to create own states for order.
+
+
+
+### Database Tables for Order State
+
+There will no forrend key between `sales_order_state` and `sales_order_state_link`
+
+### DB Design  for `sales_order_state`
+
+| id   | state   | created_at          |
+| ---- | ------- | ------------------- |
+| 1    | new     | 2019-06-01 19:55:56 |
+| 2    | new     | 2019-06-01 19:57:56 |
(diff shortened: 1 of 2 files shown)


design-documents/RefcatoringForStateCopling.md
```diff
@@ -0,0 +1,101 @@
...
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Sales\Model;
+
+interface GetOrderStateByIdInterface
```
joni-jones: I would propose to make interfaces more agnostic, it's better to have separate interfaces for state and status as other entities like an invoice, shipment, the payment might use it.

design-documents/RefcatoringForStateCopling.md
```diff
@@ -0,0 +1,101 @@
...
+{
+    /**
+     *
+     * @param int $orderId
+     * @return OrderStateInterface
+     */
+    public function execute(int $orderId);
```
joni-jones: We need to have more generic methods to get state not only by related entity id, but also by search criteria.

design-documents/RefcatoringForStateCopling.md
```diff
@@ -0,0 +1,101 @@
...
+
+namespace Magento\Sales\Model;
+
+class OrderStateInterface extends \Magento\Framework\Api\ExtensibleDataInterface
+{ 
+    public function getState() :string;
+
+    public function getCreatedAt(): string;
+  
+    public function getEventInfomation(): string;
```
joni-jones: I would propose to create a separate entity for the history log purposes.

design-documents/RefcatoringForStateCopling.md
```diff
@@ -0,0 +1,101 @@
...
+
+| id   | link_id | order_id |
+| ---- | ------- | -------- |
+| 1    | 1       | 1        |
+| 2    | 2       | 2        |
+| 3    | 3       | 1        |
+
+
+
+### DB Design  for `sales_order_state_event_log`
```
joni-jones: I would propose to add additional fields like `prev_state`, `cur_state` because just `log` will have the same drawbacks as the current comments history.

design-documents/RefcatoringForStateCopling.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+### Database Tables for Order State
+
+There will no forrend key between `sales_order_state` and `sales_order_state_link`
+
+### DB Design  for `sales_order_state`
+
+| id   | state   | created_at          |
+| ---- | ------- | ------------------- |
```
joni-jones: Looks like this table doesn't make sense with such structure because the states are not unique and it more looks like history table. I would propose to create a separate table for states and statuses which will store only different types with states/statuses.

joni-jones:
I would prefer to have state and status as separate order-agnostic entities. Something like this:
Order Entity Composition

PR #190, opened 2019-06-14
MFTF: Friendly labels to elements, actions and action groups

lbajsarowicz:
## Problem

Increase business value from Magento Functional Testing Framework output

## Solution

* User-defined labels
* Auto-generate labels

## Requested Reviewers

@okolesnyk

### DIFF
--- a/design-documents/testing/functional/bdd-like-mftf-output.md
+++ b/design-documents/testing/functional/bdd-like-mftf-output.md
@@ -34,7 +34,7 @@ Magento Functional Testing Framework (MFTF) aims to generate valuable output for
 
 ## PoC
 
-https://github.com/magento/magento2-functional-testing-framework/issues/364 
+https://github.com/magento/magento2-functional-testing-framework/issues/364
 
 ## Usage
 
@@ -117,3 +117,25 @@ Adding extra attribute to XML nodes that allows to define user-friendly label fo
 #### Benefits
 
 * Automatic label generation enforces MFTF user to provide descriptive `name`s / `stepKey`s in Test Scenarios
+
+# Implementation
+
+1. **MFTF Naming Rules**
+
+  There's no standard regarding naming in MFTF Tests. One should be formed and published in Magento DevDocs, as well as in the MFTF repository. Tests migrated / created after forming such rules should follow the standards.
+
+2. **Naming interpretation strategy**
+
+  Having consistent Naming for Tests, we are able to get benefits of it. We need to prepare strategy that describes how we convert names to human-readable labels, for example:
+
+  *  `StorefrontFillCustomerAccountCreationFormActionGroup` -> `Storefront Fill Customer Account Creation Form`
+  * `fillPasswordField` -> `Fill Password Field`
+
+  Having strategies established, Issues should be created to cover implementation in `magento/magento2-functional-testing-framework` repository.
+
+3. **Codeception CLI and Allure reports**
+
+  Having all previous done, the final step is to implement adapters both for CLI output and Allure Reports. As a result, we should receive the output with desired verbosity:
+  * Debug
+  * Non-verbose
+  * Behavioral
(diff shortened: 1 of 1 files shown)


design-documents/testing/functional/bdd-like-mftf-output.md
```diff
@@ -0,0 +1,141 @@
...
+  ```xml
+  <click selector=".niceButton" stepKey="clickButton" label="Click Nice Button"/>
+  ```
+* Action Group
+  ```xml
+  <actionGroup name="resetProductGridToDefaultView" label="Reset Product Grid">
+  ```
+* Action Group Execution
+  ```xml
+  <actionGroup ref="someActionGroup" label="Running Action Group">
```
KevinBKozan: Not sure I agree with `label` addition to actionGroup, can we not accomplish the same thing by using Codeception's `humanize()` functions on the name provided?
Example:
`<actionGroup name="loginAsAdmin"` => `I Login As Admin`
`<actionGroup name="resetProductGridToDefaultView"` => `I Reset Product Grid To Default View`

design-documents/testing/functional/bdd-like-mftf-output.md
```diff
@@ -0,0 +1,141 @@
...
+
+Example of such is Sylius Framework: https://github.com/Sylius/Sylius/blob/48729b86334ba78d8f2011b9743389024c334d56/features/checkout/paying_for_order/paying_with_paypal_during_checkout.feature#L7-L24
+
+### User-defined Label Attribute
+
+Adding extra attribute to XML nodes that allows to define user-friendly label for element / action / action group.
+
+* Element
+ ```xml
+ <element name="password" type="input" selector="#login" label="Password Input"/>
```
KevinBKozan: `<element>`s are (usually) named something intuitive for the test writer, so I'm not 100% sure the benefit of adding a separate verbosity to elements.

In your examples it's the difference between:
`I fill password` vs `I fill Password Input`

I'm not the biggest believer in BDD/Gherkin style test output but I understand it's a matter of preference, so I'd prefer keeping it as non-instrusive for developers (meaning I'd prefer utilizing the existing `type` attribute).

design-documents/testing/functional/bdd-like-mftf-output.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Benefits
+
+* Automatic label generation enforces MFTF user to provide descriptive `name`s / `stepKey`s in Test Scenarios
+
+# Implementation
+
+1. **MFTF Naming Rules**
+
+  There's no standard regarding naming in MFTF Tests. One should be formed and published in Magento DevDocs, as well as in the MFTF repository. Tests migrated / created after forming such rules should follow the standards.
```
KevinBKozan: As far as I understand, this is currently in progress (@tomreece ?)

design-documents/testing/functional/bdd-like-mftf-output.md
lbajsarowicz: That issue was partially mentioned by @okolesnyk - that in most of the cases we can "humanize" Action Group name, only a few needs extra "explanation" with specific label.

design-documents/testing/functional/bdd-like-mftf-output.md
tomreece: Yup! We are currently incrementally adding content to the heading "Functional Testing" on this page:
https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html#113-functional-testing

These rules were carefully groomed by the entire Pangolins team and should be obeyed by all.

tomreece:
Hello, this proposal is very well thought out. Thank you for taking your time to detail everything.

I would summarize this as: Make tests more readable to a wider set of people.

If you agree with my interpretation, I would like to revisit the original problem:

> Increase business value from Magento Functional Testing Framework output

I agree with you that readability would be improved but I don't understand the intended audience or their needs. How does the business benefit from this? Why does everyone need to be able to read tests with such precision?

PR #192, opened 2019-06-18
Alternative checkout flow proposal

paliarush:
## Problem

In scope of work on GraphQL and storefront APIs we have an opportunity to improve design and features of storefront checkout.

The purpose of this document is to discuss possible alternatives to current Magento checkout flow.

## Solution

This document is under discussion. Participants are welcome.


design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,50 @@
...
+       * Adjustments
+           * Taxes
+           * Cart rule discounts
+     * Billing addresses
+     * Shipping addresses
+     * Coupons
+     * Gift cards
+     * Store credit
+     * Customer (optional)
+  6. Place Order: Order
```
akaplya: we allow creating multiple quotas based on the single cart (quote segregation based on addresses), from my standpoint it makes sense to make order placement as an atomic operation which allows submitting multiple quotes. It runs us in another question - multiple payments handling together. but this also has a sense. Orders with different destination mean addresses may have different process flow which is fine. For example, the multi-tenant installation will require such functionality.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,50 @@
...
+     * Cart
+     * Addresses
+     * Date & Time (admin preview)
+  2. Quote factory arguments: Quote
+     * !Line Items
+     * Dimensions (weight & size) (based on LineItems)
+     * Shipping (requited for physical products)
+       * Addresses
+       * Selected Shipping Methods
+     * Billing address (required for virtual products). How to deal with multi-payment?
```
akaplya: let's reverse this question, can multiple quotes have the same payment method and billing address? In my opinion, this is a more relevant case for e-commerce. Definitely, for processing and querying, each order will have its own billing address but for creation argument, I would suggest to keep it as a single argument.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,50 @@
...
+  1. Cart properties:
+     * Line Items:
+       * SKU
+       * Selected options (custom/configurable)
+       * Quantity
+       * Regular price
+       * Price
+  4. Applicable cart rules calculator arguments
+     * Cart
+     * Addresses
```
akaplya: we need to pass some kind buyer representation here

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,54 @@
...
+     * Date & Time (admin preview)
+  2. Quote factory arguments: Quote
+     * !Line Items
+     * Dimensions (weight & size) (based on LineItems)
+     * Shipping (requited for physical products)
+       * Address
+       * Selected Shipping Method
+     * Billing address (required for virtual products). Need use cases
+     * Coupons
+     * Gift cards
```
akaplya: Gift cards, coupons and store credits may be moved to the payment step.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,54 @@
...
+     * PreviousTotals
+  5. Quote
+     * LineItems
+       * Totals
+       * Adjustments
+           * Taxes
+           * Cart rule discounts
+     * Billing addresses
+     * Shipping addresses
+     * Coupons
```
AlexMaxHorkun: Why are coupons, gift card codes, store credit are not part of (price) Adjustments? It would make sense for quotes to accept a list of abstract PriceAdjustmentInterface which would be used when calculating totals and to show users which adjustments were applied to their cart/order/quote. I don't think that quote should explicetly rely on taxes, coupons etc. And there must be a mechanism for 3rd-party developers to add their own adjustments. Having an abstract PriceAdjustmentInterface DTO seems that quotes would receive seems like a step to that direction

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,63 @@
...
+
+The cart is just a container for the items the user wants to purchase. In the proposed flow the cart is created as soon as user adds product to cart. The data from cart entity should be enough to render minicart and cart pages. Taxes and other adjustments are not calculated at these steps.
+
+The quote, on the other hand, contains a full break down of all adjustments calculated. It provides the user with the total he has to pay for the items in the cart.
+
+It is possible to split cart items into separate quotes. This can be done based on shipping addresses or shipping sources.
+
+Each quote is used to create a separate order. Multiple payment methods (with independent billing address each) can be selected during each order creation.
+
+Cart will depend on Catalog. Quote will depend on Cart and must not depend on Catalog. Order will depend on Quote and must not depend on Cart or Catalog.
```
Vinai: > Quote will depend on Cart and must not depend on Catalog

Why?
I don't see the benefit of this. I see the attraction of it, but I think it's wrong. It's an unnatural restriction. Signs of this are how awkward it is to pass the product dimensions. The same problems will occur also for example when other product data is required for totals calculation by third parties.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,63 @@
...
+     * For physical products on Review & Payments step
+     * For virtual products billing address has to be entered first
+  1. Cart properties:
+     * Line Items:
+       * SKU
+       * Selected options (custom/configurable)
+       * Quantity
+       * Regular price
+       * Price
+  4. Applicable cart rules calculator arguments
```
Vinai: What is the benefit of separating out the cart rules from the totals calculation?

design-documents/checkout/alternative-checkout-flow.md
Vinai: That would be the customer.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,63 @@
...
+     * Cart
+     * Addresses
+     * Date & Time (admin preview)
+  2. Quote factory arguments: Quote
```
Vinai: The Quote/QuoteFactory still seems like an aggregate for all entities required by checkout to me. If I understand correctly one of the suggestions of this proposal is to not build the quote over time, but all required data has to be provided at creation.
If this data is aggregated over time, where would it be stored? In the API client (e.g. store front)?

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,63 @@
...
+     * Shipping (requited for physical products)
+       * Address
+       * Selected Shipping Method
+     * Billing address (required for virtual products). Need use cases
+     * Coupons
+     * Gift cards
+     * Store credit
+     * Cart rules (calculated by Applicable catalog rules calculator)
+     * Customer
+  2. Totals calculator: Totals
```
Vinai: in the video of the architecture meeting where this proposal was discussed it was mentioned that totals are entities. Was that meant literally?
From my point of view the totals are a function of the quote.
The results of the calculation may be stored for performance reasons, but a total does not have any identity over time.
Also, I didn't understand the point about how this new proposal allows for sorting of total calculation.
It sounded as if that is not possible in the current implementation. However, it certainly is possible.

The downside of the current implementation is that each total aggregates it's results on the quote items and the quote. Shared mutable state is always tricky. However, it does allow for many customizations where one total calculation can change the value of a previous one. After listening to the discussion, I'm not sure if the total results where supposed to be immutable or not. If they are, how would such customizations be possible?

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,63 @@
...
+
+Each quote is used to create a separate order. Multiple payment methods (with independent billing address each) can be selected during each order creation.
+
+Cart will depend on Catalog. Quote will depend on Cart and must not depend on Catalog. Order will depend on Quote and must not depend on Cart or Catalog.
+
+  1. When Quote is created?
+     * For physical products on Review & Payments step
+     * For virtual products billing address has to be entered first
+  1. Cart properties:
+     * Line Items:
```
Vinai: Can more information on how different product types would be represented be added to the proposal?
Currently a product in the quote is represented by one or more quote items. Each quote item can have one or more quote item options to store relevant data.
These options are used for many different things, starting from enabling reorders to linking of related quote items over additional data to display on line items to many different types of customizations.
Because they are so heavily used I would like to read a little on how those problems would be solved with this proposal

design-documents/checkout/alternative-checkout-flow.md
joni-jones: > in the video of the architecture meeting where this proposal was discussed it was mentioned that totals are entities. Was that meant literally?

The idea is to have a separate immutable extensible entity for totals which can be used in other entities like order, invoice, credit memo because there are issues with the calculation of their total.

> I didn't understand the point about how this new proposal allows for sorting of total calculation.

There will be some kind of class which will check the configuration and create to a composite of totals in the order according to configuration or applied rules (like cart rules).

> It sounded as if that is not possible in the current implementation.

Actually no. The point was that the current mechanism is hard to understand because you can see the only configuration in Admin Panel and logic is smashed across the code. The main idea is making it more visible (even provide some kind of visualization in Admin Panel) and have only one place in the code.

> After listening to the discussion, I'm not sure if the total results where supposed to be immutable or not. If they are, how would such customizations be possible?

Totals should be immutable. Each totals calculator uses the result of previous totals calculator, any extension/customization can add own calculator to the list.

design-documents/checkout/alternative-checkout-flow.md
Vinai: I second this, quotes split by shipping address should still be payable with a single payment during checkout. This has been a major limitation of the current design and is one of the reasons so few merchants utilize the multi address checkout. Having to do separate payments for each order makes the process incompatible with most current payment processors.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,160 @@
...
+
+The proposed solution assumes that a quote will be an immutable object and will operate with totals list, each calculator will create add new totals object based on totals list and the order of each calculation can be changed in runtime. As one of the benefits of the proposed approached - the list of calculators and their order can be visualized for better calculation understanding.
+
+Let's consider the calculation of the totals might look like.
+
+![Totals Calculation](img/totals-calculation-pipeline.png)
+
+Each calculator receives Quote DTO and Totals List, calculates totals, creates new Totals DTO with the calculated amount and adds it to Totals List. This approach allows to do not change a quote object, have a defined interface for totals and change the order of calculation. Magento provides multiple configurations to change the order of calculation, for example, discount can be applied before shipping amount or after.
+
```
Vinai: One common customization for the current totals calculation is that custom discounts are implemented by adjusting prior totals, e.g. the subtotal, from a later total model. If the totals are immutable, there should be an option for total models to return a changed list of totals, e.g. remove or replace total models that where calculated by prior. Each total model then can be immutable, but can be replaced or removed as needed by customizations.

Vinai:
Sorry to harp on and on about backward compatibility, but I still think it would be good to include a paragraph into the design document about how the new API will be introduced.
From my point of view it should be introduced as a new module. Then existing stores can continue to use their current implementation with all customizations, while new stores and APIs can be built using the new checkout flow.

Without using this approach merchant developers will simply see "Magento broke the checkout flow in a major way without adding any benefit, so now we have to pay many hours of upgrading customizations, just to be able to use the latest version of Magento, without having any real return of investment, because the current checkout was working fine for years.". This would be really painful, l and is the reason why I hope some thoughts on the introduction process will be included in the PR.

joni-jones:
@Vinai thanks for your feedback.

The current version of the proposal contains just mention that it will an alternative API and we don't have intentions to replace everything in the current checkout. Currently, we are focusing on use cases coverage, missed parts in existing checkout, and most "painful things".

The next stage is defining extension points in the current API to see how to integrate alternative flow in the Magento ecosystem: Catalog, Sales, Payments.

If you have any thoughts like stuff which difficult to customize/extend, missed functionality, please share your opinion.

design-documents/checkout/alternative-checkout-flow.md
joni-jones: Not sure if I clearly understand your point but totals calculators will be created in runtime according to the configuration and priority and each calculator has `isApplicable` method (the customization can disable needed calculator, add a new one) which decides if calculator should be executed or not, for example, it does not make sense to execute a shipping calculator for the quote only with virtual products.

A custom discount calculator can be added in the same way as standard calculators with own priority and grand total calc will automatically collect all amounts and adjustments.

Vinai:
Over the last few years I've worked on a few customizations that involved dynamically making specific products free.

So far I've implemented that by adding a quote item option to the free product (so it is listed as a separate line item and not merged with the same regular priced product in the cart) and custom total models which set the free line item cart price to zero and adjusted the subtotal result.

How would that be implemented with the new checkout flow?

joni-jones:
Cart and Quote entities will be separated, the Cart will use own totals estimator (some calculators might be reused from quote) based on Catalog prices, also Cart will have own products representation, for example, groups can be used for products separation. The quote items will be calculated with prices from PIM.

Based on the provided case, I see there a few possible approaches:
1. The cart receives a free item with zero/non-zero price and displays the item with other items (not sure why it shouldn't be merged with other items but groups might be used), the quote gets item zero price from PIM (assumes, that is an item is free than the price in PIM also zero) and uses default calculation.
2. For the cart logic the same as in the first approach, but the quote will get a non-zero price from PIM and will have additional sales rule calculator to adjust items prices and custom subtotal calculator. Priority would allow specifying an order of calculations.

Vinai:
@joni-jones 
> not sure why it shouldn't be merged with other items but groups might be used

Because the one item is free, and the others are not.
It is a business requirement that the free item should be listed as a separate line item.

The quote can not get a zero price from the PIM because the product is not generally free. It's just the one item for free.
Imagine you want to order pizzas. You order two and get a bottle of coke for free. Other bottles of coke you also add to the cart are not free. There are many similar scenarios.
The PIM only knows about regular prices bottles of coke.

design-documents/checkout/alternative-checkout-flow.md
```diff
@@ -0,0 +1,230 @@
...
+## Uni-directional checkout flow
+
+The flow assumes that each component like Cart, Quote, Shipping, etc. will receive all needed details to perform operations. This approach allows resolving and reducing a list of dependencies between modules.
+
+![Checkout data flow](img/checkout-data-flow.png)
+
+The Cart will depend on Catalog. Quote will have a knowledge about PIM, Shipping Rates estimator, Coupons, Store Credit and other services for totals calculation. Cart will provide data for Quote. Quote will provide data for Order. Order will not have any knowledge about Catalog, Quote, Cart.
+
+![One-directional checkout flow](img/alternative-checkout-flow-2.png)
```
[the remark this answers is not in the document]
joni-jones: Totals won't be the same. In most cases, cart totals contain items summary and some adjustments like free item, the quote totals - shipping rate, discounts, taxes, customer balance, gift cards. But, in general, I don't see restrictions to don't use it for cart re-calculation, as calculation should be used for different scenarios like B2B quotes.

PR #193, opened 2019-06-19
Service Isolation: Integration Tests

buskamuza:
## Problem

What should we do with integration tests when application is deployed on separate instances?
Related issue - https://github.com/magento/service-isolation-backlog-wiki/issues/30

## Solution

Preserve integration tests framework, but limit it to a single instance execution. Existing tests should be revisited and updated/removed.

## Requested Reviewers

@melnikovi @maghamed @paliarush

### DIFF
--- a/design-documents/service-isolation/integration-tests.md
+++ b/design-documents/service-isolation/integration-tests.md
@@ -14,15 +14,16 @@ Take into account future support of microservices architecture.
 
 Before deciding whether we should drop support of integration tests for distributed systems, let's revisit purpose of integration tests in Magento:
 1. validate extension points of a component (potential integration with other components)
-2. validate integration with other existing components. Integration tests don't use mocks (usually)
-3. validate integration with the environment. Integration tests run on installed application, with real DB, etc
+1. validate integration with other existing components. Integration tests don't use mocks (usually)
+1. validate integration with the environment. Integration tests run on installed application, with real DB, etc
+1. replacement for unit tests **inside a module**, w/o mocking of module dependencies
 
 
 ## Solution
 
 From the purpose of integration tests described above, 
 - 2 and 3 can and should be covered by functional tests as they validate real scenarios and so bring more value
-- 1 still makes sense for integration tests, but only in scope of one instance (single-area application or service) because this is where in-process extension points are supported  
+- 1 and 4 still makes sense for integration tests, but only in scope of one instance (single-area application or service) because this is where in-process extension points are supported  
 
 Based on this, it makes sense to run integration tests for a single-area application instance (or for a service in the future).
 
@@ -33,14 +34,17 @@ Changes to integration tests:
   1. Testing package declares which Magento package it depends on
   1. Testing packages are deployed on the same instances their dependencies are deployed. Deployment can be managed in the same way as Magento feature packages
 1. Each instance has a dedicated DB (cache, other resources) for tests, so tests can run in isolation
-1. Integration tests must not make network calls to other instances (though any network calls should not be necessary in integration tests)
-
-Changes to integration testing framework:
-1. Application area is configured per instance and does not change per test. `@magentoAppArea` annotation is eliminated as unnecessary.
-1. `\Magento\TestFramework\TestCase\AbstractController` is eliminated. Consider converting tests relying on it (400+) to functional tests.
+1. Integration tests must not make network calls to other instances
 
 ### Scope of Work
 
-Proposed changes require revisiting all or almost all of existing integration tests.
-It is a lot of work, but would help clean up and clarify purpose of integration tests.
-By deprecating undesired framework functionality and using static tests validation, new tests can be implemented in a way that can be used with distributed deployment, while old tests will be considered compatible with monolithic deployment only.  
+Proposed changes require revisiting all or almost all of existing integration tests which is a lot of work.
+
+For existing tests:
+- no changes
+- clearly document that old integration tests can be run only on a monolithic deployment
+- some tests may be converted to the new tests if required during Service Isolation work
+
+For new tests:
+- located in a Magento module (separate or the one being tested) 
+- MUST NOT execute network calls to controllers that may belong to a different instance (either because of different area or different service)
--- a/design-documents/service-isolation/integration-tests.md
+++ b/design-documents/service-isolation/integration-tests.md
@@ -1,0 +1,46 @@
+# Integration Tests
+
+## Problem Statement
+
+Integration tests run in the same PHP process as Magento application.
+A test from one module may depend on a data fixture or a model from another module.
+In case of **single-area setup** or setup with **different application areas served by different instances**, integration tests and their dependencies may be deployed on **different instances** and, as a result, such tests won't be able to resolve their dependencies.
+
+The question: should integration testing framework support distributed deployment or keep it compatible to monolithic deployment only?
+
+Take into account future support of microservices architecture.
+
+## Purpose of Integration Tests in Magento
+
+Before deciding whether we should drop support of integration tests for distributed systems, let's revisit purpose of integration tests in Magento:
+1. validate extension points of a component (potential integration with other components)
+2. validate integration with other existing components. Integration tests don't use mocks (usually)
+3. validate integration with the environment. Integration tests run on installed application, with real DB, etc
+
+
+## Solution
+
+From the purpose of integration tests described above, 
+- 2 and 3 can and should be covered by functional tests as they validate real scenarios and so bring more value
+- 1 still makes sense for integration tests, but only in scope of one instance (single-area application or service) because this is where in-process extension points are supported  
+
+Based on this, it makes sense to run integration tests for a single-area application instance (or for a service in the future).
+
+It should be taken into account that in case of multiple single-area application, DB is shared between all of them.
+
+Changes to integration tests:
+1. Break down tests by module (according to new module structure).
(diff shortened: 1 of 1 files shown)


design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+
+Changes to integration tests:
+1. Break down tests by module (according to new module structure).
+  1. Testing package declares which Magento package it depends on
+  1. Testing packages are deployed on the same instances their dependencies are deployed. Deployment can be managed in the same way as Magento feature packages
+1. Each instance has a dedicated DB (cache, other resources) for tests, so tests can run in isolation
+1. Integration tests must not make network calls to other instances (though any network calls should not be necessary in integration tests)
+
+Changes to integration testing framework:
+1. Application area is configured per instance and does not change per test. `@magentoAppArea` annotation is eliminated as unnecessary.
```
melnikovi: I think we need to keep it as we still want to run integration tests on monolith.

design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+- 2 and 3 can and should be covered by functional tests as they validate real scenarios and so bring more value
+- 1 still makes sense for integration tests, but only in scope of one instance (single-area application or service) because this is where in-process extension points are supported  
+
+Based on this, it makes sense to run integration tests for a single-area application instance (or for a service in the future).
+
+It should be taken into account that in case of multiple single-area application, DB is shared between all of them.
+
+Changes to integration tests:
+1. Break down tests by module (according to new module structure).
+  1. Testing package declares which Magento package it depends on
```
melnikovi: Do you propose to also move tests into separate package?

design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+From the purpose of integration tests described above, 
+- 2 and 3 can and should be covered by functional tests as they validate real scenarios and so bring more value
+- 1 still makes sense for integration tests, but only in scope of one instance (single-area application or service) because this is where in-process extension points are supported  
+
+Based on this, it makes sense to run integration tests for a single-area application instance (or for a service in the future).
+
+It should be taken into account that in case of multiple single-area application, DB is shared between all of them.
+
+Changes to integration tests:
+1. Break down tests by module (according to new module structure).
```
melnikovi: Should fixtures go to domain module?

design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+Before deciding whether we should drop support of integration tests for distributed systems, let's revisit purpose of integration tests in Magento:
+1. validate extension points of a component (potential integration with other components)
+2. validate integration with other existing components. Integration tests don't use mocks (usually)
+3. validate integration with the environment. Integration tests run on installed application, with real DB, etc
+
+
+## Solution
+
+From the purpose of integration tests described above, 
+- 2 and 3 can and should be covered by functional tests as they validate real scenarios and so bring more value
```
melnikovi: Functional tests take a lot of time to run. Integration tests allow developer to get quick feedback loop and I think it's important to keep it. Especially considering that we currently suggest to create integration tests instead of unit tests.

buskamuza: Also, include WebAPI tests

design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+In case of **single-area setup** or setup with **different application areas served by different instances**, integration tests and their dependencies may be deployed on **different instances** and, as a result, such tests won't be able to resolve their dependencies.
+
+The question: should integration testing framework support distributed deployment or keep it compatible to monolithic deployment only?
+
+Take into account future support of microservices architecture.
+
+## Purpose of Integration Tests in Magento
+
+Before deciding whether we should drop support of integration tests for distributed systems, let's revisit purpose of integration tests in Magento:
+1. validate extension points of a component (potential integration with other components)
```
buskamuza: WebAPI tests can be used for this too

design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+2. validate integration with other existing components. Integration tests don't use mocks (usually)
+3. validate integration with the environment. Integration tests run on installed application, with real DB, etc
```
buskamuza: Replacement for unit tests inside a module, but w/o mocking.

buskamuza: @maghamed : we should use integration tests for infra (e.g., service mesh) testing.
* @melnikovi : we probably need different type of tests

design-documents/service-isolation/integration-tests.md
buskamuza: * Not feasible for existing tests.
* For new tests, integration testing framework needs improvements to support modularity. Limitations:
  * Tests can be run on a defined set of enabled modules
  * Tests can be run per service (no inter-service communication)

design-documents/service-isolation/integration-tests.md
buskamuza: For inter-service testing, we need to think about another type of tests.

buskamuza:
For multi-area setup, check which tests can be run on separate instances.

design-documents/service-isolation/integration-tests.md
```diff
@@ -0,0 +1,46 @@
...
+Changes to integration tests:
+1. Break down tests by module (according to new module structure).
+  1. Testing package declares which Magento package it depends on
+  1. Testing packages are deployed on the same instances their dependencies are deployed. Deployment can be managed in the same way as Magento feature packages
+1. Each instance has a dedicated DB (cache, other resources) for tests, so tests can run in isolation
+1. Integration tests must not make network calls to other instances (though any network calls should not be necessary in integration tests)
+
+Changes to integration testing framework:
+1. Application area is configured per instance and does not change per test. `@magentoAppArea` annotation is eliminated as unnecessary.
+1. `\Magento\TestFramework\TestCase\AbstractController` is eliminated. Consider converting tests relying on it (400+) to functional tests.
```
joni-jones: I'm not sure if replacing such integration tests by functional tests is a good idea. In some cases, it will be overcomplicated to cover some functionality of a single controller by a functional test because controllers can be entry points for independent business flow. For instance, we have a webhook for Signifyd integration to update case details. To cover such functionality by the functional test we need:
* Go through all checkout flow and place the order
* Go to Signifyd account and make an update for a case
* Go to Admin Panel and verify if the case is updated

Integration test allows doing it much easier because we need only to use fixtures to create an order and stub case and just call a webhook controller with all needed data.

buskamuza: As long as an integration test doesn't make network calls to other Magento instances, will it be sufficient?
I still think that Magento instances should be isolated, so running of integration tests doesn't spoil another instance.

design-documents/service-isolation/integration-tests.md
buskamuza: https://github.com/magento/service-isolation-backlog-wiki/issues/32

buskamuza:
Had a discussion with core architects about the role of integration (and other tests).

Meeting minutes from July 13th:
- How to do TDD with functional tests? In Magento projects, func tests are usually written after the functionality is done
  - Olga to talk to Vinai to collect his experience
- Necessary features for integration tests
  - Customization. Ability to write a test that can be customized from another module if the module changes tested behavior
  - Ability to disable a test from another test. It would allow to replace original test if behavior changes
  - Parametrized fixtures to improve reusability of fixtures
  - Also better granularity
  - In-memory DB – look at this once again for integration tests. Talk to Kasian
  - Distinguish tests that require full installation (with DB) and the ones that require only configured DI

buskamuza:
Meeting notes from July, 16th:
* Participants: @buskamuza , @Vinai
* Integration tests
  * The most useful kind of tests
  * Much easier to write than MFTF tests
  * No need to distinguish between installed app vs non-installed app tests
  * Look at https://github.com/ho-nl/magento2-ReachDigital_TestFramework
  * Int tests are useful for SIs
  * Core int tests fail outside of default Travis setup
  * Disabling selected tests would be a useful feature
  * The framework is good, the only complain is speed
  * Fixture would be good to improve. Object Mother (builders) pattern. See also https://github.com/tddwizard/magento2-fixtures
* MFTF
  * More brittle than other kinds of tests. More work to maintain because more parts are involved
  * Slow. Devs don’t run them and rely on CI only
  * With time devs stopped fixing MFTF tests, because they thought it’s the test is broken, not functionality, and just deploy w/ broken func tests. There is no time to fix the tests afterwards
  * MFTF don’t fit into TDD
  * Cover 2-3 important user scenarios with func tests, but not more
* Unit tests
  * Writing unit tests when doing TDD only. Re-running unit tests afterwards is not very useful
* Manual vs auto testing
  * During feature dev, some manual testing is necessary

Vinai:
Another note regarding this design document: from my point of view it would be fine to only support monolithic deployments for integration tests. 
Supporting distributed deployments would cause a lot of (80%?) more complexity. Distributed deployments will be rare and only used by top tier merchants, so all of that effort would only help a small percentage of users.
There might be specific tests targeting distributed deployments, for those it might make sense to be able to emulate those kind of deployments, rather than having an actual distributed deployment on kubernetes or such.

PR #200, opened 2019-07-08
Added the proposal for Abstract Method removal

joni-jones:
The`AbsractMethod` removal strategy proposal.

### DIFF
--- a/design-documents/abstract-method-removal.md
+++ b/design-documents/abstract-method-removal.md
@@ -1,0 +1,37 @@
+# Abstract Method removal
+
+## Overview
+
+Currently Magento supports two API for new payment integration creation `\Magento\Payment\Model\Method\AbstractMethod` and `\Magento\Payment\Model\Method\Adapter` (the part of [Magento Payment Provider Gateway](https://devdocs.magento.com/guides/v2.3/payments-integrations/payment-gateway/payment-gateway-intro.html)). The `AbstractMethod` was deprecated in Magento 2.1 release and `Adapter` is a recommended approach for new payment integrations.
+
+Let's consider the reasons why the `AbstractMethod`'s usage should be eliminated:
+
+- Allows storing credit card details in database
+- The credit card storage requires additional protection for PCI compliance certification
+- A new integration should override most of the parent's methods and properties to provide a new functionality
+- The integration's support and customization requires to change existing code instead of adding/changes only some parts (like Payment Gateway allows adding new commands, request builders, response handlers, etc.)
+- Does not support Vault storage
+- An introduction of new features is complicated to provide in BC way
+- Complicated to cover an integration or separate parts by integration/unit tests
+- Tightly coupled with Sales Order Management
+- Inheritance leads to the code duplication
+- The usage of `AbstractMethod` requires to support both API `AbstractMethod` and `Adapter`
+- Does not have a developer documentation
+
+## Removing strategy
+
+There are still a few core integrations (Offline payments and PayPal), also some extensions (~30 on Marketplace), which based on or use `AbstractMethod`. The integrations should be re-implemented based on Magento Payment Provider Gateway. The core integrations can be re-implemented during patch releases (the old integrations should be marked as deprecated).
+
+The desired state for 2.4 Magento release:
+ 
+- Core integrations do not use `AbstractMethod` (old integrations marked as deprecated).
+- All related to `AbsractMethod` infrastructure code is marked as deprecated (which not marked yet)
+- All appropriate methods/functions trigger `E_DEPRECATED` warning (https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#deprecation).
+
+
+The desired state for 2.5 Magento release:
+ - Remove all `AbstractMethod` infrastructure related code/database columns
+ 
+## Summary
+ 
+The 3rd party developers will have a full minor-release cycle to rework their integrations. As each payment integration has a lot of specific implementation details, own flow and data set for payment operations, there is no unified approach to provide tools for code migration from `AbstractMethod` to Payment Provider Gateway.


Vinai:
As stated before, I think the right approach is to keep the current `AbstractMethod` based implementation code around and not remove it in the short or medium term.

They are used as the base for many customized and tweaked payment integrations in existing integrations. Only analyzing the extensions in the marketplace misses the larger part of the picture.

To be able to move forward while still not breaking backward compatibility, mark them only as deprecated and introduce new implementations with a new name for offline payments and PayPal using `\Magento\Payment\Model\Method\Adapter`.

In any new installations the old `AbstractMethod` payment packages should be disabled by default, while the new implementations are enabled. For upgrades, the new packages are either enabled (if they don't conflict with the existing modules) or disabled (if there is a technical reason not to enable both at the same time).
For reference, MSI was introduced with the same process.

Keep the deprecated packages around for at least another two or three years, while encouraging and educating all to move to the new API.

This approach would avoid breaking backward compatibility, thus keeping merchants and developers happy.

speedupmate:
I agree with keeping old stuff around and just bringing in new things in general. 

Related to payment methods I think faster and stricter deprecation model is justified , this will force merchants and extension providers to move faster on critical areas and outdated and unmaintained stuff will just fail to work. Failing hard here is a good thing. 

* This area is often targeted by malware and attacks , keeping old and deprecated but still working ways around will hurt this area of Magento . 
*  the complexity of payment methods and devs trying to just hack things together by copy/pasting things from different payment method as examples not understanding the flow will soon show us "mutants" trying to use both ways at the same time

Vinai:
@speedupmate  
Breaking the existing user base will result in stores not being upgrades for a longer time.
From my point of view that's an even more dangerous strategy than keeping the old code around.

Regarding your second point, there will always be developers with no clue what they are doing building insecure payment integrations. The only way to improve the situation is through education.

joni-jones:
> there will always be developers with no clue what they are doing building insecure payment integrations.

That's why a prefer an approach to have only one way to build the integration.

> The only way to improve the situation is through education.

We wrote the documentation, presented a new framework on multiple conferences but I still see questions on magento.stackexchange.com that developers don't know which approach should be chosen and still start building a new payment integration based on `AbstactMethod`. Unfortunately, there is not enough feedback from the community to understand which parts are missed in the documentation and what is not clear enough.

Vinai:
> > there will always be developers with no clue what they are doing building insecure payment integrations.
> 
> That's why a prefer an approach to have only one way to build the integration.

That is not a solution to the problem.

> 
> > The only way to improve the situation is through education.
> 
> We wrote the documentation, presented a new framework on multiple conferences but I still see questions on magento.stackexchange.com that developers don't know which approach should be chosen and still start building a new payment integration based on `AbstactMethod`. Unfortunately, there is not enough feedback from the community to understand which parts are missed in the documentation and what is not clear enough.

I believe the problem is developers look at the core code and use it as an example and also there still is Magento 1 knowledge being applied in the context of Magento 2.

Since implementations using `AbstractMethod` where not marked as deprecated in the core, new implementations by third party developers are still done that way.
However, only once alternative implementations that can be used as a reference are available in the core, will there be proper adoption.

But even then, existing code in live stores won't be changed. That's why it's important to preserve backward compatibility for as long as possible (from my point of view forever would be the right target time frame)

joni-jones:
> I believe the problem is developers look at the core code and use it as an example and also there still is Magento 1 knowledge being applied in the context of Magento 2.
Since implementations using AbstractMethod where not marked as deprecated in the core, new implementations by third party developers are still done that way.

From a developer perspective, for me, it's not clear why I should look at some private implementation which is based on the deprecated approach if sample modules and the documentation propose to use another approach.

speedupmate:
@Vinai 
> Breaking the existing user base will result in stores not being upgrades for a longer time.
> From my point of view that's an even more dangerous strategy than keeping the old code around.

If you look how payment gateways deprecate out API versions it's pretty harsh and strict since standards are involved, laws are involved etc. API -s are shut down , notifications are sent , deadlines are set, ignorant merchants disconnected. 

Payment method supporting code could be handled the same. If something in this area is outdated it is outdated. If new approach is back-ported it's not as big issue as new ways will be rewarded with also "working on lower versions". Devs can replace components without upgrading your whole mess. 

Sites not being able to upgrade is the same: "educate, educate, educate" . In all I agree with you but there are (mostly security related) areas that can be approached more strictly. Magento team should have clear opinion on this as I see different conversations in github offering different strategies and going over and over on what should be the deprecation strategy (and also you fighting for this constantly).

@joni-jones 
> From a developer perspective, for me, it's not clear why I should look at some private implementation  which is based on the deprecated approach if sample modules and the documentation propose to  use another approach. 

Different devs have different skills it's that simple. They might have same amount of money to spend on a solution, time to burn but results will differ.  

Teaching and writing good devdocs is art on its own . As is skill to understand content (good or bad). Magento2 overhead in abstraction just limits true understanding of its moving parts to (really)small superset of developers.

Being involved in checkout area related development for a long time I can say by looking at the core components (payment methods in this case) that even core team struggles to get their magic and theres no clear pattern to follow by just looking at the code. Same with 3rd party approaches vary with long margins.

The point to pick up from this for me is that complex parts and parts that need security baked in could be handled by different rules. This of course assumes proper communication and really clear timelines, back-porting options if possible.

Vinai:
> > Breaking the existing user base will result in stores not being upgrades for a longer time.
> > From my point of view that's an even more dangerous strategy than keeping the old code around.
> 
> If you look how payment gateways deprecate out API versions it's pretty harsh and strict since standards are involved, laws are involved etc. API -s are shut down , notifications are sent , deadlines are set, ignorant merchants disconnected.

In this case payment extensions that are secure and use current APIs would still be broken.

> Payment method supporting code could be handled the same. If something in this area is outdated it is outdated. If new approach is back-ported it's not as big issue as new ways will be rewarded with also "working on lower versions". Devs can replace components without upgrading your whole mess.

If an API is released as stable, removing it will cause breakage, regardless if it has been marked as deprecated or not. For that reason I believe it's okay to mark APIs as deprecated once an alternative has been implemented, but ideally the deprecated code should not be removed.

This implies that code should only be released as stable once an API has had the chance to mature, as an initial design will probably never work ideally when it meets with reality.

Payment APIs are no exception to this rule. In fact, it's even more important to maintain backward compatibility because payment is a crucial part of almost all Magento instances.

Vinai:
> > I believe the problem is developers look at the core code and use it as an example and also there still is Magento 1 knowledge being applied in the context of Magento 2.
> > Since implementations using AbstractMethod where not marked as deprecated in the core, new implementations by third party developers are still done that way.
> 
> From a developer perspective, for me, it's not clear why I should look at some private implementation which is based on the deprecated approach if sample modules and the documentation propose to use another approach.

I believe that in this case it's not appropriate to take your point of view and apply it to the developer ecosystem at large.
Magento developers have a long history of studying core code to understand how to do things.

This is because for the longest time there was no, or only very limited or hard to understand documentation.
Released sample code was insufficient, quickly outdated and not maintained.
This culture within the Magento ecosystem to study "how it's done in the core" is nothing that will change quickly. In fact, I think it's one of the strengths of the ecosystem.

"Private implementation detail" for Magento developers does NOT mean "don't use directly", instead it means "look at this code and try to copy and adjust it".

If you want developers to do things "the right way" release code that does things "the right way".
But please do it without breaking backward compatibility. Instead, release new APIs as new modules that can live side by side with existing deprecated code (like described above, like MSI, like many projects do).

joni-jones:
> I believe that in this case it's not appropriate to take your point of view and apply it to the developer ecosystem at large.

I'm not trying to apply my point of view, I'm trying to understand why developers use one approach instead of another. 

> Released sample code was insufficient, quickly outdated and not maintained.

Could you explain why? `SamplePaymentGateway` still works.

> If you want developers to do things "the right way" release code that does things "the right way".

Due to time/resource limitations for payment integrations migration from Magento 1, `AbstractMethod` and `Payment\Adapter` were released both in Magento 2.0 (with integrations based on `Adapter`). I believe this is true for most of the payment extensions based on `AbsractMethod` and adapted from M1 to M2.

> But please do it without breaking backward compatibility. Instead, release new APIs as new modules that can live side by side with existing deprecated code (like described above, like MSI, like many projects do).


I agree that in most cases it should be done in this way but, as described in the proposal, there are reasons why `AbsractMethod` should be removed in the future.

PR #201, opened 2019-07-08
Added a proposal with place order flow improvements

joni-jones:


speedupmate:
@joni-jones 

There will still be need for both pre-order placement (pro forma is actual accounting term for this https://en.wikipedia.org/wiki/Pro_forma )  and post-order placement payment flows until "pay in advance"  payment types exist (means forever). 

Dependant of regions authorisation based payment methods might just be not available at all as people don't have credit cards and some regions (in EU) live solely off from debit cards (that do not allow authorisation at all) . This is often tied to local legislation as well.  

> - If payment transaction fails the order entity won't be saved into the database 

this can be expected behaviour to avoid spamming out order placements or never issue  order if theres no advance payment. DDOS for orders is also a scenario to count for. This is not the only place in flow something can fail . Also every thing that can fail in this flow does not need to be converted automatically to refund or order cancellation  etc

> - A successful transaction does not guarantee that an order will be saved into the database

successful transaction does not have to guarantee that its not related. in pre order placement type of payment this should be validated/estimated before the flow starts eq if all quote details are valid to promise a transaction or to be converted to order later the flow can be executed (As promise if needed) if something fails after successful transaction then this process needs to allow executing different parts in the same flow to be repeated after flaws or dataset  is fixed (to save the transaction from getting refunded) . 

A successful transaction means something is promised, customer has paid  based on quote/offer, money is yours to fill the promise. Failing to create order at this point and refunding based on that technical fact is  just lost money for the merchant. Money can be saved easily  if this order placement can be repeated (with fixed issues technical or in data) after transaction is made even if manually in database or in multiple systems. Same way merchant needs a way to refund this (be it inside magento or outside does not really matter) if its a truly failed process.

> - If order place action fails, a customer won't get any details like reserved order increment ID, so we can't easily link a customer's failed order with the failed payment transaction

customer is tied to quote or not?  You started the transaction with full customer and transaction details 
 those do not have to come from order in all cases, redirection needs to be stateless process in ideal world passed over with signed data in all directions. Legal fact that data is sent can be stored separately or within quote/order/payment entities. This also reveals where the chain of events has flawed out  , request can be repeated or reconstructed to finish a certain workflow

> - If order fails after payment transaction, the payment transaction won't be void or refunded

This is not related to the flow, this is a matter of solving cross systems issues (accounting, ERP, gateways, magento etc) its not always possible to refund failed transactions, orders, payments, invoices within single process , approvals might be needed , accountant help might be needed etc parallel processes started. Also merchant might want to make sure this transaction/sale gets closed for their benefit instead of refunding it automatically.

>- Sales module contains interfaces to work with payments infrastructure (like adapter classes) but payment infrastructure anyway depends on the order entity

again this is not true it might or might not depend on order entity but does not have to , just a bold claim 

> To reduce timeouts we need to split up place order and processing payment transaction operations. The order entity should be saved as soon as it converted from quote and submitted to place operation. The payment transaction execution should be asynchronous background process.

Sure thing that order placement and payment flow should possibly be separated. But at the same time you need to allow multiple flows .  If processes are designed to be async  or as you illustrated as backend processes with queues etc or as parallel processes then this does not have to happen "as soon as possible" . Chain of events decides if entities needs to be created/saved and in what order 

eq: 
* quote -> per order placement flow > transaction promise based on quote> true/false/failure(timeout, error etc)  > next actions 
* quote -> order entity creation/save > after order placement flow  > transaction promise > true/false/failure(timeout, error etc)  > next actions 
* quote -> order entity creation promise | transaction promise | backend processes for e-mail notifications | etc events > true/false/failure(timeout, error etc)  > next actions

joni-jones:
@speedupmate thanks for your feedback

> this can be expected behaviour to avoid spamming out order placements or never issue order if theres no advance payment.

It will be configurable flow but the main issue right now that orders are lost by different reasons and it's hard to find out the original reason. But definitely, corner cases like unavailable message queue, simultaneous transactions, etc should be taken into account.

> DDOS for orders is also a scenario to count for. 

In the case of DDOS attacks, the difference between current and proposed flow will be only that merchant will have a possibility to see such orders with fraud status. The other mechanism should be used to prevent spamming and DDOS attacks.

> Also every thing that can fail in this flow does not need to be converted automatically to refund or order cancellation etc

This proposal doesn't offer an idea to make auto cancellations or refunds. As mentioned in the proposal If the order fails after payment transaction, the payment transaction won't be void or refund automatically.

> successful transaction does not have to guarantee that its not related

In the current implementation, a payment transaction is coupled with Sales Order Management infrastructure and they are related. The idea to separate them. A payment should operate only with data needed to make a transaction (like amount, items, currency, previous transaction, etc.) and should not depend on order/invoice/creditmemo entity.

>> Sales module contains interfaces to work with payments infrastructure (like adapter classes) but payment infrastructure anyway depends on the order entity

> again this is not true it might or might not depend on order entity but does not have to , just a bold claim

Probably I defined my point of view as not clear enough. In the current implementation, the payment infrastructure depends on Sales Order Management (payment infrastructure extracts needed data from an order entity) but it should not. Again, the payment infrastructure just should receive data, make a transaction and return the result, it should not worry about how the order details should be updated.

> But at the same time you need to allow multiple flows .

This proposal assumes bulk operations: "... a lot of payment gateways support batch operations, so some transactions can be combined to bulks and executed in the one scope...".

> If processes are designed to be async or as you illustrated as backend processes with queues etc or as parallel processes then this does not have to happen "as soon as possible" . Chain of events decides if entities needs to be created/saved and in what order

As Quote, Sales Order Management, Payment are complicated and have a lot of relationships, this is the first step to separate payment operations, do not lose orders for negative cases and reduce dependency from payment to order. The next things, order entity simplification like extracting states, statuses, totals, defining transaction API. Your examples look like CQRS approach, I like that idea but it looks more like alternative API but the idea to provide a refactoring for current implementation in BC way.

speedupmate:
> Probably I defined my point of view as not clear enough. 

It indeed is , all described in proposal leaves impression that old ways/methods will be removed and new ones honoured . If internal mechanisms is what you will separate then I completely misunderstood the proposal. 

I was confused by backend processes on payment area as to process a "form based redirection method" in background sounds like a job for a service-worker in frontend tech , many payment methods in EU won't even do authorisation or backend processing  (jet) and solely relay on message exchange via web requests indicating "don't kill the browser, session or internet connection etc" messages on each step of the process. Even if they exchange tokens or preflight requests to generate their forms theres not a lot you can do "on background" in these cases beside choosing if you want to do it before or after order processing. 

But I do understand you now when you have explained what you are wanting to separate 

> our examples look like CQRS approach,

it does not assume CQRS but just describing clear flows and giving out promises based on those actions based on existing methods already present in core (in frontend and in backend) could be handled better.

joni-jones:
> I was confused by backend processes on payment area as to process a "form based redirection method" in background sounds like a job for a service-worker in frontend tech , many payment methods in EU won't even do authorisation or backend processing (jet) and solely relay on message exchange via web requests indicating "don't kill the browser, session or internet connection etc" messages on each step of the process. Even if they exchange tokens or preflight requests to generate their forms theres not a lot you can do "on background" in these cases beside choosing if you want to do it before or after order processing.

The separation won't affect the UI flow because redirection can't be done in the background as in most cases requires customer interaction. For example, PayPal Express place order flow assumes that customer will be redirected to PayPal and when a customer chooses a payment method and redirected back the transaction should be finalized by `DoExpressCheckout` API call. According to the proposed changes, only transaction finalization will be moved to the background process, the other steps of the customer flow will be the same.

PR #203, opened 2019-07-08
Web api controllers

AlexMaxHorkun:
## Problem

Service contracts used directly as presentation layer for web API

## Solution

Introduce presentation layer controllers and DTOs for web APIs

## Requested Reviewers

@paliarush


AlexMaxHorkun:
@buskamuza @akaplya @melnikovi @joni-jones please review the latest changes

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
+# Controllers for web APIs, user input processing on top of service contracts execution
```
joni-jones: I would propose to remove "Controllers" and add something like middleware.

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+                    <!-- Validators for the whole collection -->
+                    <property name="addresses" entity="customer_addresses" group="web_api_create" walk="false" />
+                </properties>
+            </parameter>
+        </parameters>
+    </validation>
+</route>
+```
+  
+REST/SOAP framework then would create validator for entity _"customer"_ and group _"web_api_create"_, perform validation
```
paliarush: How web API framework would create validator? Will it be generic framework logic based on declaration above or each validator will have to be implemented separately and just referenced from the declaration?

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+```
+right before passing arguments to service contracts. This way validation rules could be reused and extended by 3rd-party
+developers via configs.
+ 
+This rules lists are extensible by 3rd party developers allowing them to easily add validation for new properties
+they introduce or change defaults via config files.
+ 
+ 
+#### Introduce reusable user input processors
+Sometimes before passing arguments to a service contracts we would have to perform user input processing that cannot
```
paliarush: This section should also be moved to dedicated Validation proposal. That proposal should describe validation approach holistically. This topic was discussed multiple times in the past but I don't think we have solid technical vision for validation approach.

design-documents/web-api-controllers.md
AlexMaxHorkun: Agreed

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
+# Controllers for web APIs, user input processing on top of service contracts execution
+## RESTful and SOAP API
+#### Current situation
+Service contracts, their arguments and return types are supposed to be used directly
+as web API endpoints. The problem is service contracts are also used as API
+and are supposed to be a part of business layer. By using them for presentation developers may encounter various
+problems like having to put additional logic related to specifically processing web API requests (fields validation
```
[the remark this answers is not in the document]
AlexMaxHorkun: not sure what do you mean?

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+for instance) inside models, DTOs used as arguments and return types may have sensitive data that is fine to use programmatically but
+should not be exposed to web APIs or simply having properties needed for implicit business logic that would confuse web
```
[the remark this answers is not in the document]
AlexMaxHorkun: * We can choose what service contracts to exposed but there is no standard way to solve situations when we need to expose certain functionality but service contracts responsible for it are not fit to be exposed as web API
* Developers can assign an ACL resource to an endpoints but that's it. In some cases more complex authorization is required like checking ownership over a specific entity or access only to certain properties of an entity
* there is a way to ignore certain properties coming in requests but no way to hide them in responses

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+ 
+The service contract responsible for it has this method signature, right off the bat we can see an argument that should
+not be accessible to web API - _redirectUrl_ which is used to set redirect URL inside confirmation E-mail.
+And what about password? From a web API client's perspective it should be the $customer's property but it's a separate
+account because _CustomerInterface_ is meant to represent business layer entity which doesn't store original password
+and only has the password's hash. Also _CustomerInterface_ has _defaultShipping/Billing_ properties, but then
+_AddressInterface_ also has _isDefaultShipping/Billing_ properties - which one should a web API client use to set a
+customer's default billing/shipping address? Another problem - customer's addresses have _customerId_ property which
+doesn't make sense in context of a customer managing itself since a customer can manage only their own addresses.
+All of these problems stem from having business layer contracts/entities exposed as presentation layer operations/data.
```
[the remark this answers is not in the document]
AlexMaxHorkun: I don't think that when designing a service contract for a module a developer must always consider it's being used DIRECTLY as web API

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+ 
+## HTML Presentation
+#### Current situation
+Controllers implementing _ActionInterface_ do complex validation/authorization themselves which is fine as they are
+supposed to be responsible for user input processing, still there's a problem with other areas not being able to reuse
+this logic.
+ 
+ 
+## What needs to be done to solve this?
+#### Using operation-specific DTOs
```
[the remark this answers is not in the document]
AlexMaxHorkun: I don't think that we should always use operation specific DTOs but we have to propose such idea to developers with our devedocs articles regarding code design and try to employ them when needing in the future to provide examples.

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+    //No getPassword()
+    ....
+}
+```
+ 
+By having these operation-specific DTOs we would only have the data we need as arguments and responses.
+ 
+ 
+#### Use declarative validation
+Web API clients displaying forms should be able to get a list of invalid properties sent to display to end users or
```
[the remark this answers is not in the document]
AlexMaxHorkun: agreed. Will make merging this PR easier

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+even have validation rules provided to them. Right now entities in Magento are being validated manually and errors
```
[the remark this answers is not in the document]
AlexMaxHorkun: I don't touch that topic here and I'm not sure core Magento needs it but with declarative validation rules it will be possible for merchants to implement

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+are being displayed one at the time. Having validation rules provided with the endpoints configuration would solve these
```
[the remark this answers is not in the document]
AlexMaxHorkun: even so these messages do not point to a particular chain of properties and it's impossible to gather them programmatically and display errors for corresponding form fields

design-documents/web-api-controllers.md
AlexMaxHorkun: it would use Magento\Framework\Validator\Factory to create validators based on the configuration (groups, entity, constaints defined)

design-documents/web-api-controllers.md
AlexMaxHorkun: These user input processors are the topic of this design doc. They are not meant to be validators but rather classes containing user-input processing/presentation-layer logic to be reused among other presentation-layer classes like controllers/graphQL resolvers

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,289 @@
...
+ 
+webapi.xml
+```xml
+<route url="/V1/cart/mine/shippingAddress" method="POST">
+    <service class="Magento\Sales\Api\OrderManagerInterface" method="updateShippingAddress"/>
+    <resources>
+        <resource ref="anonymous"/>
+    </resources>
+    <processors>
+        <before class="Magento\Sales\Presentation\WebApi\CartProcessor" method="beforePostAddress" />
```
nuzil: ❤️

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,289 @@
...
+    private $userContext;
+    
+    .....
+
+    /**
+     * @throws SecurityViolationException
+     */
+    public function validateAddressForOrder(CustomerAddressInterface $shippingAddress): void
+    {
+        $customerId = $this->userContext->getUserId();
```
joni-jones: How the validator will receive the `$userContext` if it injects via constructor? For example, GraphQl resolver receives it via `resolve` method and the validator instance already should be initialized.

AlexMaxHorkun: the default implementation of UserContextInterface is a composite of all possible context classes which cycles them until one of them can return a user, I assume for GraphQL context it will use CustomerSessionUserContext to get the user on runtime when validation is invoked

joni-jones: My question is how the context dependency will be initialized because when GraphQl resolver calls `$this->addressValidator->validateAddressForOrder()` in `resolve` method the `OrderShippingAddressValidator` object should be already initialized with the `Context` dependency. So, probably, `OrderShippingAddressValidator` should receive context via method argument instead of constructor?

design-documents/web-api-controllers.md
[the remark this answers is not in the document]
paliarush: `fields validation and are supposed to be part of business layer` - this is controversial from my point of view since there is no technical vision for validation

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,289 @@
...
+    //No getPassword()
+    ....
+}
+```
+ 
+By having these operation-specific DTOs we would only have the data we need as arguments and responses.
+ 
+###### How to introduce it to Magento?
+There are couple of ways we can introduce this concept to Magento:
+* use it for new storefront API
```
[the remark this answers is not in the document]
AlexMaxHorkun: that's just one suggestion meaning that we should employ this concept when designing new storefront API since they will be new and it will be easier to design them with this concept in mind

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,287 @@
...
+Example is checking whether is a customer allowed to use a saved shipping address for an order by checking whether the
+address was created by the same customer - we'd have to do it programmatically since validation rules wouldn't work
+because of not having access to the full context and our ACL system is not fit to process those cases. HTML controllers
+and GraphQL resolvers would be able to include such logic but it would not be reusable by other areas. This validation
+could be performed by the service contract responsible for posting order but service contracts are not supposed to know
+about authenticated users.
+ 
+The solution to these problems would be to introduce reusable area-agnostic user input processors that would perform
+such actions. These would be arbitrary classes with no common interface but they would be reusable across different areas
+and reside inside a special namespace `Presentation` to inform developers that this classes belong to presentation
```
buskamuza: I propose `InputValidator` or `InputValidation` namespace for validators. It's more precise.
Processors are area-specific, so I don't see a reason to group them under one namespace. In the example below, the processor is specifically for `WebApi`.

AlexMaxHorkun: @buskamuza don't want to limit these classes containing presentation area logic to be seen as validators only. Some of them may contain logic related to extracting data from requests, reading headers for instance. Right now such classes just reside inside Model namespace but they are not part of service layer. What about Model\ContextAware? For classes like session managers, classes for generating URLs for users, input validators, access validators etc?

buskamuza:
Use cases:
1. ACL for specific fields in Admin Panel (e.g., layout updates).
2. Brute force check.

design-documents/web-api-controllers.md
```diff
@@ -31,6 +31,28 @@ doesn't make sense in context of a customer managing itself since a customer can
...
+authorization related logic.
+ 
+Examples of cases when we need additional context aware/presentation logic:
+* When we need to validate entity ownership before performing an operation using authenticated user entity
+* Limit frequency of operation execution (for anti-bruteforce measures)
+* Prepare service layer DTOs for presentation, use additional service contracts for reading data to add to an
+  operation's result (most used in HTML area)
+* Plugin that would execute an additional operation based on session data/user context
+* Request throttling for a specific operation
+* Save additional data to session after performing an operation for presentation purposes
```
paliarush: We should not rely on session for web API requests.

design-documents/web-api-controllers.md
```diff
@@ -31,6 +31,28 @@ doesn't make sense in context of a customer managing itself since a customer can
...
+user input. Since service contracts are not supposed to know about context they are also not feet to contain additional
+authorization related logic.
+ 
+Examples of cases when we need additional context aware/presentation logic:
+* When we need to validate entity ownership before performing an operation using authenticated user entity
+* Limit frequency of operation execution (for anti-bruteforce measures)
+* Prepare service layer DTOs for presentation, use additional service contracts for reading data to add to an
+  operation's result (most used in HTML area)
+* Plugin that would execute an additional operation based on session data/user context
+* Request throttling for a specific operation
```
paliarush: Why is this use case separate from anti brute force one?

design-documents/web-api-controllers.md
```diff
@@ -31,6 +31,28 @@ doesn't make sense in context of a customer managing itself since a customer can
...
  
+###### Examples of classes that contain presentation layer logic
+While service contracts implementations usually can contain all the logic needed to execute an operation various
+presentation areas would require additional logic to convey results to users or to extract operations' arguments from
+user input. Since service contracts are not supposed to know about context they are also not feet to contain additional
+authorization related logic.
+ 
+Examples of cases when we need additional context aware/presentation logic:
+* When we need to validate entity ownership before performing an operation using authenticated user entity
+* Limit frequency of operation execution (for anti-bruteforce measures)
```
paliarush: Why do we care about this on the application level instead of infrastructure level?

design-documents/web-api-controllers.md
```diff
@@ -31,6 +31,28 @@ doesn't make sense in context of a customer managing itself since a customer can
...
  
  
+###### Examples of classes that contain presentation layer logic
+While service contracts implementations usually can contain all the logic needed to execute an operation various
+presentation areas would require additional logic to convey results to users or to extract operations' arguments from
+user input. Since service contracts are not supposed to know about context they are also not feet to contain additional
+authorization related logic.
+ 
+Examples of cases when we need additional context aware/presentation logic:
+* When we need to validate entity ownership before performing an operation using authenticated user entity
```
paliarush: In General, this can be achieved using built-in param override mechanism.

design-documents/web-api-controllers.md
```diff
@@ -31,6 +31,28 @@ doesn't make sense in context of a customer managing itself since a customer can
...
+* Limit frequency of operation execution (for anti-bruteforce measures)
+* Prepare service layer DTOs for presentation, use additional service contracts for reading data to add to an
```
paliarush: Please provide an example

design-documents/web-api-controllers.md
```diff
@@ -117,17 +139,25 @@ and GraphQL resolvers would be able to include such logic but it would not be re
 could be performed by the service contract responsible for posting order but service contracts are not supposed to know
 about authenticated users.
  
-The solution to these problems would be to introduce reusable area-agnostic user input processors that would perform
+The solution to these problems would be to employ reusable area-agnostic context aware classes that would perform
 such actions. These would be arbitrary classes with no common interface but they would be reusable across different areas
-and reside inside a special namespace `Presentation` to inform developers that this classes belong to presentation
+and reside inside a special namespace `ContextAware` to inform developers that this classes belong to presentation
```
paliarush: This name is confusing too. Similar classes are already located in `Controller` namespace, why don't we use existing namespace for these purposes? It also prevents potential confusion caused by putting such logic into `Model` namespace.

design-documents/web-api-controllers.md
AlexMaxHorkun: true but I think classes that work with it should be separated

design-documents/web-api-controllers.md
AlexMaxHorkun: brute force involves captcha, the throttling involves managing response times

design-documents/web-api-controllers.md
AlexMaxHorkun: we do have CAPTCHA in our application level since we want to protect our functionality and can't rely on merchants' infrastructure

design-documents/web-api-controllers.md
AlexMaxHorkun: for REST/SOAP, what about HTML, GraphQL and CLI? I think we should reuse such logic

design-documents/web-api-controllers.md
AlexMaxHorkun: adding company attributes to /customer/me endpoint but not to every CustomerRepository::getById() result since when it's used inside a chain there is no need to load them because other module are not aware of such attributes and will not use them

design-documents/web-api-controllers.md
AlexMaxHorkun: Do you propose to put these classes that will be usable across different areas like REST, GraphQL, CLI into Controller namespace? A sub-namespace like Controller/Helper?

paliarush: Yes, not `Helper` however. E.g.`Controller/Validator`

AlexMaxHorkun: I describe multiple cases when we would create classes working with contexts like request, session, authenticated user reading HTTP request object, providing data for templates/response objects, doing authorization and validation. Those are not just validators

AlexMaxHorkun: furthermore Controller namespace reserved for HTML area controllers which are being picked up by name by our framework depending on URI requested. That might create some problems for us

design-documents/web-api-controllers.md
```diff
@@ -0,0 +1,320 @@
...
+All of these problems stem from having business layer contracts/entities exposed as presentation layer operations/data.
+ 
+ 
+###### Examples of classes that contain presentation layer logic
+While service contracts implementations usually can contain all the logic needed to execute an operation various
+presentation areas would require additional logic to convey results to users or to extract operations' arguments from
+user input. Since service contracts are not supposed to know about context they are also not feet to contain additional
+authorization related logic.
+ 
+Examples of cases when we need additional context aware/presentation logic:
```
buskamuza: These are not use cases, but assumptions that some use cases need this logic.
Can you please provide use cases (scenarios)? For example, "When updating customer address from SF, we should validate that the customer owns it". For example, this use case should not need any additional framework implementation. W/o real use cases we might be adding features that will move us in the direction opposite to the desired state.

So far, classes that are needed for presentation (e.g., data filtering/conversion) should be just located somewhere close to that presentation (controller, block, etc). And other cases (e.g., ownership validation) can either be solved differently or should not be considered basis for generalization (it's just a feature that should be implemented as is on the framework level).

PR #204, opened 2019-07-15
Domain Whitelist for Configurable 3rd Party Redirects

danielrenaud:
## Problem

Given an implementation where Magento acts as a headless server, there are cases where Magento should accept redirect URLs from the client. These URLs should be validated in some way to prevent an open redirect vulnerability.

## Solution

Introduce a new configuration where a list of trusted redirect domains can be specified. Implement a class top validate redirect URLs against this whitelist.

## Requested Reviewers
@paliarush 
@joni-jones 
@AlexMaxHorkun

### DIFF
--- a/design-documents/redirect_domain_whitelist.md
+++ b/design-documents/redirect_domain_whitelist.md
@@ -25,10 +25,10 @@ We can use this as a whitelist to validate redirect urls, significantly reducing
 The domain whitelist configuration should live in `app/etc/env.php` for the following reasons:
 
 - This configuration file is already in use and well-known by implementors
-- env.php is a secure location that is only editable via file edit or cli command
+- env.php is a secure location
 - The configuration could be set by setup scripts using `setup:config:set` cli command
 
-We will add a new option (`--domain-whitelist`) to the `setup:config:set` that accepts a comma separated list of domains.
+We will add a new option (`--domain-whitelist`) to the `setup:config:set` command that accepts a comma separated list of domains.
  
 ```
 bin/magento setup:config:set --domain-whitelist="foo.store.com, bar.store.com, otherstore.com"
@@ -49,10 +49,10 @@ Disallowing wildcards (e.g. `*.store.com`) will prevent unintentionally (or inte
 
 #### Redirect Validation Implementation
 
-A new class will be introduced in the WebapiSecurity module to encapsulate the logic of validating redirect URLs against this whitelist (e.g. `\Magento\WebapiSecurity\Model\Validator\Redirect`).
+A new class will be introduced in the Backend module to encapsulate the logic of validating redirect URLs against this whitelist (e.g. `\Magento\Backend\Model\Url\RedirectValidator`).
 It will perform similar validation as already done by `\Magento\Framework\Validator\Url` as well as compare the URL's domain against the domain whitelist in `env.php`.
 
-`\Magento\WebapiSecurity\Model\Validator\Redirect::isValid()` will validate:
+`\Magento\Backend\Model\Url\RedirectValidator::isValid()` will validate:
 1. Valid URL format (checked by `\Magento\Framework\Validator\Url`)
 2. Scheme (checked by `\Magento\Framework\Validator\Url`)
 3. Host (domain checked against whitelist configuration and Magento configuration*)
--- a/design-documents/redirect_domain_whitelist.md
+++ b/design-documents/redirect_domain_whitelist.md
@@ -1,0 +1,56 @@
+# Domain Whitelist for Frontend Redirects
+
+With the introduction of PWA as well as other server-client Magento solutions, the Magento host is acting as a headless server.
+This headless Magento server can serve multiple frontend applications, potentially on different domains.
+Currently there is no defined way that Magento knows which client domains it serves and are safe to redirect the user.
+This presents a problem for some PayPal payment solutions when trying to implement them through GraphQl.
+
+## Issue
+Several Paypal payment methods require some URL to redirect to once the payment has completed (e.g. [PayFlow](https://developer.paypal.com/docs/classic/payflow/integration-guide/#introducing-the-gateway-checkout-solutions)).
+Currently Magento uses the current host and some predefined controller that handles the redirect from PayPal.
+
+For the GraphQl implementation we want to support the scenario where a single "headless" Magento server will act as the API server for multiple frontend clients (possibly on different domains).
+In order to support this case, we need the ability for the frontend client to specify the URLs that PayPal will redirect the frontend to after the PayPal transaction.
+However, introducing the ability for the client to define the redirect URL opens a security concern.
+With an open redirect it could be possible for an attacker to manipulate this URL to send the user to a malicious website after payment.
+
+## Proposed Solution
+Currently, the Magento server has no way of knowing which domains all it's frontend clients are using.
+Therefore we are not able to screen the URLs provided by the client, to ensure they are safe.
+
+We propose introducing a new configuration that will contain a list of domains that are allowed in these redirect URLs.
+We can use this as a whitelist to validate redirect urls, significantly reducing the security concern.
+
+#### Configuration
+The domain whitelist configuration should live in `app/etc/env.php` for the following reasons:
+
+- This configuration file is already in use and well-known by implementors
+- env.php is a secure location that is only editable via file edit or cli command
+- The configuration could be set by setup scripts using `setup:config:set` cli command
+
+We will add a new option (`--domain_whitelist`) to the `setup:config:set` that accepts a comma separated list of domains.
+ 
+```
+bin/magento setup:config:set --domain-whitelist="pwa.store.com, othersite.store.com"
+```
+Would result in an `env.php` entry:
+```
+...
+'domain_whitelist' => [
+    'pwa.store.com',
+    'othersite.store.com'
+]
+...
+```
+
+#### Redirect Validation Implementation
+
+A new class will be introduced in the WebapiSecurity module to encapsulate the logic of validating redirect URLs against this whitelist (e.g. `\Magento\WebapiSecurity\Model\Validator\Redirect`).
+It will perform similar validation as already done by `\Magento\Framework\Validator\Url` as well as compare the URL's domain against the domain whitelist in `env.php`.
+
+`\Magento\WebapiSecurity\Model\Validator\Redirect::isValid()` will validate:
+1. Valid URL format (checked by `\Magento\Framework\Validator\Url`)
+2. Scheme (checked by `\Magento\Framework\Validator\Url`)
+3. Host (domain checked against whitelist configuration)
+
+The URL route will not be validated/sanitized by this class. This can be handled by the specific business logic class as it will be use case dependent. 
--- a/design-documents/redirect_domain_whitelist.md
+++ b/design-documents/redirect_domain_whitelist.md
@@ -1,6 +1,6 @@
-# Domain Whitelist for Frontend Redirects
+# Domain Whitelist for Configurable 3rd Party Redirects
 
-With the introduction of PWA as well as other server-client Magento solutions, the Magento host is acting as a headless server.
+With the introduction of PWA as well as other solutions, there are use cases where Magento is treated as a headless server, serving a frontend client.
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=204) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=204) before we can accept your contribution.<br/>

AlexMaxHorkun:
I suggest placing this configuration beside `base URL` configuration and calling it something like `Front-end domains` since in future it may be used for other purposes beside payment systems like for links inside E-mails we send to customers. Therefor we should not have the setting inside app/etc/env.php but rather as a regular config path just like `base URL`.

joni-jones:
@AlexMaxHorkun could you explain your objections about placing the configuration in `app/etc/env.php`?

AlexMaxHorkun:
> @AlexMaxHorkun could you explain your objections about placing the configuration in `app/etc/env.php`?

@joni-jones
* It's harder to update by merchants
* some setups might have it as read-only
* this setting doesn't require more protection than `base URL` does

joni-jones:
> some setups might have it as read-only
this setting doesn't require more protection than base URL does

Not sure if these options are applicable because:
 - the storefront application (like PWA, etc) can't be changed in runtime, so anyway the storefront should be configured before deployment
 - `env.php` can be changed only with a person who has access to the deployment infrastructure or environment, the Admin configuration only restricted by credentials and an appropriate role which is less strict to access rather than access to deployment infrastructure.

AlexMaxHorkun:
@joni-jones 
> * the storefront application (like PWA, etc) can't be changed in runtime, so anyway the storefront should be configured before deployment
 
design doc mentions a situation when Magento is used only as a headless server having multiple storefronts so there is a possibility of storefronts being added on runtime when a new integration is created for example
 
> * `env.php` can be changed only with a person who has access to the deployment infrastructure or environment, the Admin configuration only restricted by credentials and an appropriate role which is less strict to access rather than access to deployment infrastructure.
 
Again, you can change base URL with the regular Configuration admin page, how is this more sensitive?

joni-jones:
It mentioned as theoretically Magento can be headless but practically it doesn't work, even for WEB API we have a state as PHP session.

Your statement was `frontend domain` requires less protection than `base url` or I didn't understand it. My point, the protection level should be the same and `env.php` modification requires more strict privileges than Admin configuration.

DrewML:
> For the GraphQl implementation we want to support the scenario where a single "headless" Magento server will act as the API server for multiple frontend clients (possibly on different domains).
In order to support this case, we need the ability for the frontend client to specify the URLs that PayPal will redirect the frontend to after the PayPal transaction.
However, introducing the ability for the client to define the redirect URL opens a security concern.
With an open redirect it could be possible for an attacker to manipulate this URL to send the user to a malicious website after payment.

Does this actually need to be user-defined? Curious why GraphQL can't use the hostname + port + protocol used to query the API.

joni-jones:
@DrewML such links should not be user-defined. They should be accessible/changeable via configuration/deployment/SI etc. The general case, Magento backend sends storefront's redirect URLs to a payment gateway and a customer should not have a possibility to define/re-define such URLs.

DrewML:
> @DrewML such links should not be user-defined. They should be accessible/changeable via configuration/deployment/SI etc. The general case, Magento backend sends storefront's redirect URLs to a payment gateway and a customer should not have a possibility to define/re-define such URLs.

Should have clarified. "user-defined" here meaning defined by a developer in some configuration. I'm not sure I quite understand why we'd let someone customize the host/port/protocol. Is there a situation where you'd want a return URL that didn't use the hostname/port/protocol used to invoke the GraphQL API?

DrewML:
> This headless Magento server can serve multiple frontend applications, potentially on different domains.

This is probably the part I don't understand. There shouldn't be a case where the GraphQL API is called with a different hostname than what the front-end is being served with.

I _believe_ in the pwa-studio project they're handling this either through a proxy when using upward-js (https://github.com/magento-research/pwa-studio/blob/7172a3f04174ae1d8bfa96033bdc66ee5ff85837/packages/venia-concept/upward.yml#L30-L35), or doing nothing (no need to proxy) when using upward-php. Someone from @magento/pwa-studio can correct me if I'm wrong, though.

cpartica:
@DrewML not all frontends will have upward, Venia is just one implementation. we will have to solve this on Magento backend side

joni-jones:
@DrewML I thought about "user-defined" like as a customer.

> Is there a situation where you'd want a return URL that didn't use the hostname/port/protocol used to invoke the GraphQL API?

I think, in general, we are talking about the case when storefront and backend are located on different domains.

> There shouldn't be a case where the GraphQL API is called with a different hostname than what the front-end is being served with.

Now, GraphlQl resolvers are part of backend application and GraphQl queries are part of frontend, they could be deployed on different instances with different domains.

AlexMaxHorkun:
> It mentioned as theoretically Magento can be headless but practically it doesn't work, even for WEB API we have a state as PHP session.
> 
> Your statement was `frontend domain` requires less protection than `base url` or I didn't understand it. My point, the protection level should be the same and `env.php` modification requires more strict privileges than Admin configuration.

@joni-jones I didn't say it requires less protection I said it's the same and users are allowed to change base URL via admin panel

DrewML:
> It mentioned as theoretically Magento can be headless but practically it doesn't work, even for WEB API we have a state as PHP session.

Are there any plans for the GraphQL API to support token auth, or is it always going to be PHP session based? Because, for any user with 3rd party cookies disabled (anyone using Safari or Firefox in 2019), this isn't going to work.

With Safari's "Intelligent Tracking Prevention" specifically, if the UI interaction doesn't happen on the same domain the API is living under, it's going to drop the cookies
https://webkit.org/blog/7675/intelligent-tracking-prevention/

joni-jones:
@DrewML we had such discussion in the scope of Services Isolation project where storefront will communicate with services via API gateway and token-based authentication will be used instead but I'm not sure if that is near future.

joni-jones:
I see a few still open questions: should it be an environment variable or Admin configurable option?

DrewML:
@danielrenaud and I spoke about this a bit more offline.

## Concerns
Domain allow-lists are generally not great for security, and tend to be forgotten about or incorrectly configured. If possible, we should opt to "do the right thing," instead of expecting a developer to setup proper config.

## Why can't we do the right thing?
At the moment, the Magento instance has no knowledge that there are external UIs for the store. In my opinion, the "redirect allow list" is a work-around for this missing data.

With an allow list, when an API request comes in to `foo.com/graphql`, the redirect would then be allowed to be using _any_ domain from the allow-list, but this is too permissive: only the UI currently being used should be able to receive the redirect.

## Possible Solution
In my opinion, it should be possible to set a configuration value for the hostname used by the UI for any given store ID. In other words, store ID 4 could be configured to be associated with the headless "foo.com" domain. This way, the only place we'd ever generate redirects to would be the correct UI.

DrewML:
After chatting with @zetlen a bit, one idea came up.

Is there a reason not to use secure base url? This would impose a requirement that any headless storefront needs to map the url for the UI in the Magento admin, but they'll already have to do this for order confirmation emails (as mentioned above).

If that happened, it seems like we could just limit the redirect url to be under the domain of the host associated with the current store ID.

Thoughts?

danielrenaud:
That makes sense. Furthermore, since the server knows the configured URL for the store, we could only have the client provide the redirect path; and from the store code the server can figure out the secure_base_url.

zetlen:
This only works if the admin area and frontend area can use different base URLs; if they both rely on `secure_base_url`, then it's a little awkward.

We want to preserve a one-way dependency, where the PWA can run with the Magento API as a backing service, without specifically having to configure Magento to serve it. This seems like a case for plain old authorization: the PWA's middle tier should have a shared secret with Magento.

If it has this shared secret, then the middle tier can place special calls that authorize a domain to receive a redirect URL, _one request at a time_. Let a GraphQL mutation exist to start a PayPal redirect payment workflow.

```graphql
mutation startPaypalRedirect($returnBase: String) {
  examplePayPalRedirectAuth(returnBase: $returnBase) {
    auth
    targetUrl
  }
}
```

In that example, `$returnBase` is a `String` and not a `String!` because it is optional. If you don't provide a `$returnBase`, then this mutation requires no authorization, to support simple workflows on the same domain. But if you pass a custom `$returnBase`, it would only work with an administrator bearer token in the Authorization header.

The middle tier for the PWA would have an environment variable with an admin auth token. It would intercept this type of mutation, and instead of proxying the raw call through to Magento as it does with other GraphQL calls, it would place an authorized call to create a one-time PayPal session that redirected back to the current domain.

This has the advantage of being stateless and not requiring a whitelist; it only requires that the Magento-consuming client app, the PWA, have an authorized token. That seems to expose the least surface area to security risks, while enabling the most alternative workflows.

DrewML:
> We want to preserve a one-way dependency, where the PWA can run with the Magento API as a backing service, without specifically having to configure Magento to serve it

Is there a solve for the other blockers relating to that? As mentioned above, it seems like we're going to need that mapping at some point for order confirmation emails.

zetlen:
@DrewML Yeah, let me amend that to "We want to preserve a one-way dependency _as much as possible_". Ideally we would simply pass the domain through as part of the object the email template renders.

nathanjosiah:
> At the moment, the Magento instance has no knowledge that there are external UIs for the store

There was a big PR that was just merged to security yesterday that is relevant to this discussion. https://github.com/magento/magento2ce/pull/4499

zetlen:
@nathanjosiah Is this new whitelist for downloadable domains the _only_ domain whitelist we maintain in store data, or is there more than one?

nathanjosiah:
@zetlen As far as I am aware.

zetlen:
@nathanjosiah It seems like an antipattern to maintain parallel, disconnected allow-domain lists for different privileges. Seems like it ought to be a shared entity.

```sql
CREATE TABLE `capabilities`
(
  `id` int PRIMARY KEY,
  `name` varchar(255) UNIQUE,
  `description` varchar(255)
);


CREATE TABLE `domains`
(
  `id` int PRIMARY KEY,
  `name` varchar(255) UNIQUE,
  `description` varchar(255)
);


CREATE TABLE `domain_capability`
(
  domain_id int,
  capability_id int,
  FOREIGN KEY (domain_id) REFERENCES domains(id) ON DELETE CASCADE,
  FOREIGN KEY (capability_id) REFERENCES capabilities(id) ON DELETE CASCADE
);

INSERT INTO `capabilities` (name, description) VALUES
(
  'allow_href',
  'Allow this domain to be the origin of link targets in emails, SMS, and notifications'
),
(
  'allow_payment_redirect',
  'Allow this domain to be the "return" and "cancel" target of payment workflows that require redirects to and from the payment provider'
),
(
  'allow_download',
  'Allow this domain to be the origin of Downloadable Product links'
),
(
  'allow_graphql',
  'Allow requests from this domain to the GraphQL API'
);

INSERT INTO `domains` (name, description) VALUES
(
  'store-pwa.app',
  'Progressive Web App storefront domain'
),
(
  'affiliate-tracking.tld',
  'Affiliate marketing redirector'
),
(
  'some.cdn',
  'CDN for static assets'
);

insert into domain_capability (
  select domains.id as domain_id, capabilities.id as capability_id
  from domains, capabilities
  where 
    (
      domains.name = 'store-pwa.app' and
      capabilities.name in
        ('allow_graphql', 'allow_payment_redirect', 'allow_href')
    )
      or
    (
      domains.name = 'some.cdn' and
      capabilities.name = 'allow_download'
    )
      or
    (
      domains.name = 'affiliate-tracking.tld' and
      capabilities.name = 'allow_href'
    )
);
```

Or something like that

nathanjosiah:
@zetlen I agree that a broader approach such as what you suggested appears to make more sense Magento-wide but I am not on the team that made the whitelist so I don't know what was considered. Our team (Borg) was pushing for this feature for some time and our proposal was repeatedly shut down. I'm not sure which architect or PO approved it but the Obsessive Owls team would know as they are the ones that delivered it.

danielrenaud:
After discussion it seems like a broader approach will be needed.

PR #208, opened 2019-07-17
Evaluation of Apollo schema federation for GraphQL

paliarush:
## Problem

Apollo [schema federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/) was recently released as a replacement for their [schema stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching/). The goal of this document is to evaluate the need of migration from existing Magento schema stitching mechanism to the Apollo schema federation

## Requested Reviewers

@akaplya

### DIFF
--- a/design-documents/graph-ql/framework/schema-federation.md
+++ b/design-documents/graph-ql/framework/schema-federation.md
@@ -1,0 +1,32 @@
+## Apollo schema federation as an alternative to Magento schema stitching
+
+Apollo [schema federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/) is a replacement for their [schema stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching/). Schema federation is intended to:
+- Solve modularity and extensibility challenges of schema stitching
+- Allow code separation by concern
+- Provide support for distributed GraphQL
+
+## Apollo federation vs Magento server-side schema stitching
+
+Feature parity:
+- Modularity
+- Types extensibility
+- Separation of concerns
+
+Benefits of Apollo Federation:
+- Works with microservice architecture
+- Mechanism for entity reference resolution between independent services
+
+Concerns:
+- No implementation of federation gateway in PHP
+- Requires refactoring of existing schema. Potentially in a backward incompatible manner
+
+## Migration path from Magento schema stitching to Apollo federation
+
+To start supporting federated schema, we need to:
+1. Rewrite modular GraphQL schemas to include entity `@key` directives and type extensions using `extend type` syntax
+1. Implement federation gateway in PHP according to the [federation spec](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/). Its responsibility is to generate final schema and execute federated queries (resolve references).
+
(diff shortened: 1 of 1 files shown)


Vinai:
#### Additional thoughts on what @paliarush wrote in the document:

Using a supported mechanism to merge schemas like federation might give better support from third party tools. For examples, linters would not complain about some types trying to redefine existing types.

In general federation seems more well thought out than schema stiching, which now in hindsight seems more like an ad-hoc solution.

I'm not sure in what way federated schema merging would be useful in the context of distributed deployments. Maybe @paliarush can share some more thoughts on this topic?

I agree that moving to schema federation is not a high priority at this point in time, unless some clear benefit for distributed deployments can be identified.

#### Thoughts on preserving backward compatibility while introducing federation:

Preserving backward compatibility would probably require introducing a new file for the schema declarations following federation syntax, e.g. `etc/federated-schema.graphqls`.

An alternative approach could be that both formats use the same file name `schema.graphqls` but then some kind of detection algorithm would need to me used to figure out the merge strategy, e.g. the presence of `@key` directives of the use of the `extends` keyword, or maybe a mandatory comment in the first line as a file header, like `# file-type: federated schema`.

All existing `schema.graphqls` files could be migrated to `federated-schema.graphqls` files (or, if type detection is implemented, be migrated to the federated syntax) with release of the new functionality.
Magento would then use the federation strategy to merge all `federated-schema.graphqls` files first, and then stitch any `schema.graphqls` files onto the result.

paliarush:
@Vinai thank you for sharing your thoughts.

The main benefit of federated schema for distributed deployment I see is the schema of each individual service is valid on its own. And the GraphQL API of each individual service is fully functional without other services, i.e. Reviews service will be functional even without Products service, the details of Product records will, however, not be resolvable until Product service is deployed. These two features are not available with current Magento schema stitching mechanism.

Vinai:
I think the independent individual services working would not only require federation, it would probably also require resolvers to be implemented keeping distributed deployments in mind accordingly.

If I understand https://blog.apollographql.com/apollo-federation-f260cf525d21 correctly, one of the services would have to extend the other one to add the products to the reviews and vice versa, so the same should also be possible with schema stitching, or not?

paliarush:
Magento schema stitching in its current form allows extension, e.g. of reviews with products and vice versa. 

What it does not allow is running reviews service without product service, because there will be unresolved references to products. With federated schema, instead of references to full product entities there will be references to product types which only contain some kind of ID (marked with `@external`) and this ID will be fetched by the reviews service.

Vinai:
Thanks @paliarush, I think I get it. In essence it would allow the reviews service to be deployed without the corresponding catalog `schema.graphqls` files. Those would only be deployed on the catalog service in your example, right?

Thank indeed sounds like a useful thing that is required by the direction Magento is headed, is it not?
So in the case that it is required to be supported by Magento at one point in time anyway, wouldn't it be better to implement federated schema sooner rather than later?
Or do you prefer to implement it later when a open source PHP implementation of federated schema becomes available?

PR #213, opened 2019-07-24
[MFTF] Additional improvement on test directory name

okolesnyk:
## Problem
Not everyone understand what test types are located under `dev/tests/acceptance`

## Solution
Rename `dev/tests/acceptance` to `dev/tests/ui-functional`

### DIFF
--- a/design-documents/testing/functional/test-package.md
+++ b/design-documents/testing/functional/test-package.md
@@ -104,7 +104,21 @@ This packages will be used only on CICD systems and Development environments whe
 ### Disadvantages:
 1. Dependencies between test modules must be maintained manually by Automation engineer.
 
+----
+## Additional improvement
+Since architecture team agreed to move Functional Tests from `app/code` to `dev/tests` there is a request to rename `acceptance` folder to `ui-functional`.
+
+### Benefits:
+1. More clear definition of what tests located under that directory.
+
+### Disadvantages:
+1. Require changes to CI Infrastructure.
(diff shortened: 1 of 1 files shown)


design-documents/testing/functional/test-package.md
```diff
@@ -104,7 +104,21 @@ This packages will be used only on CICD systems and Development environments whe
...
+----
+## Additional improvement
+Since architecture team agreed to move Functional Tests from `app/code` to `dev/tests` there is a request to rename `acceptance` folder to `ui-functional`.
+
+### Benefits:
+1. More clear definition of what tests located under that directory.
+
+### Disadvantages:
+1. Require changes to CI Infrastructure.
+2. Backward incompatible change because some SIs and Extension developers can have own tests located under `dev/tests/acceptance`.
```
paliarush: Can we mitigate this by implementing logic which would find tests in the old locations?

okolesnyk: Yes, we can. But I think we should avoid it.
Because we will support such places as:
`dev/tests/acceptance/...`
`dev/tests/ui-functional/...`
`app/code/...`
`vendor/...`

The problem is to merge XML files in proper order. Which path should have higher priority?
And even if we implement this and set priority to new dir name it can still cause failure.

fooman:
I do not believe the advantages stated for the rename outweigh the disadvantages. If you want people to adopt MFTF don't break their investments to make it work.

If there is confusion about what each folder does just add a Readme file under `dev/tests`.

PR #215, opened 2019-07-30
add proposal for Magento\Framework\Url\RouteValidator

cpartica:
## Problem

Validating a path in M2 is not currently possible. The problem ocured during
[Domain Whitelist for Configurable 3rd Party Redirects](https://github.com/magento/architecture/pull/204). We just don't have a service class to validate paths, for code re-usage, with that can be injected in all resolvers or classes that need this validator such as PayPal Express and others.

## Solution

As this validator doesn't have anything to do with GraphQl, the best place for it is to add a new validator in framework
similar to existing `\Magento\Framework\Url\Validator`. So this new class in framework will need approval.

Looking at the implementation of the `\Magento\Framework\Url\Validator` it's based on `\Zend_Validate_Abstract` so it
complies with our standard of not using native php functions such as parse `parse_url`. Nor we would want to use regex.

## Requested Reviewers

@paliarush

### DIFF
--- a/design-documents/graph-ql/path-validator.md
+++ b/design-documents/graph-ql/path-validator.md
@@ -1,0 +1,100 @@
+# Magento Framework validators
+The purpose of this document is to a validator for relative paths / routes not only urls.
+
+## Terminology
+`UrlValidator` - The existing framework validator `\Magento\Framework\Url\Validator`
+
+`RouteValidator` - The proposed `Magento\Framework\Url\RouteValidator`
+
+## Overview
+### Current situation
+In the Magento 2 framework we can only validate URLs `http://www.somedomain.com/some/path/route`.
+We don't have the possibility to configure this validator to test relative urls, aka: paths or routes.
+We should have such kind of validator available for cases where we would just need to validate a relative path and also
+eliminate the directory traversal when such an input comes from the user. 
+
+### Problem
+Such situation of validating a path did occur with the proposal
+[Domain Whitelist for Configurable 3rd Party Redirects](https://github.com/magento/architecture/pull/204).
+We just don't have a service class to validate paths, for code re-usage, with that can be injected in all resolvers or
+classes that need this validator such as PayPal Express and others.
+
+### Proposed solution
+As this validator doesn't have anything to do with GraphQl, the best place for it is to add a new validator in framework
+similar to existing `\Magento\Framework\Url\Validator`. So this new class in framework will need approval.
+
+Looking at the implementation of the `\Magento\Framework\Url\Validator` it's based on `\Zend_Validate_Abstract` so it
+complies with our standard of not using native php functions such as parse `parse_url`. Nor we would want to use regex.
+
+
+The new class `RouteValidator` will look like the following and it's namespace will be `Magento\Framework\Url`:
+```   
+   namespace Magento\Framework\Url;
+   
+   /**
+    * Validate route and paths
+    */
+   class RouteValidator extends \Zend_Validate_Abstract
+   {
+       /**
+        * @var \Zend\Validator\Uri
+        */
+       private $validator;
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=215) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=215) before we can accept your contribution.<br/>

design-documents/graph-ql/path-validator.md
```diff
@@ -0,0 +1,100 @@
+# Magento Framework validators
+The purpose of this document is to a validator for relative url routes not only full urls commonly referred as absolute url routes.
+
+## Terminology
+`UrlValidator` - The existing framework validator `\Magento\Framework\Url\Validator`
+
+`RouteValidator` - The proposed `Magento\Framework\Url\RouteValidator`
```
buskamuza: Why is it called `RouteValidator`? Looks like it validates path.

design-documents/graph-ql/path-validator.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+The new class `RouteValidator` will look like the following and it's namespace will be `Magento\Framework\Url`:
+```   
+   namespace Magento\Framework\Url;
+   
+   /**
+    * Validate url routes
+    */
+   class RouteValidator extends \Zend_Validate_Abstract
```
joni-jones: It introduces an additional dependency to Zend Framework 1, which we are trying to avoid.

design-documents/graph-ql/path-validator.md
```diff
@@ -0,0 +1,100 @@
...
+           $this->validator->setAllowAbsolute(false);
+           $this->_setValue($value);
+   
+           $valid = $this->validator->isValid($value)
+               && $this->validator->getUriHandler()->getQuery() === null
+                 //prevent directory traversal
+               && strpos($value, '..') === false;
+   
+           if (!$valid) {
+               $this->_error(Validator::INVALID_URL);
```
joni-jones: I think it doesn't make sense to add some errors because in general, the method's call result should be if the URL is valid or not. So there is no reason to have such methods like `getMessages` with errors.

kandy:
Can you add rules for valid route?

AlexMaxHorkun:
@cpartica Also please make sure the validator rejects path starting with javascript:/data: etc. since 1. they are not useable as redirects. 2. Potentially these URLs maybe used in templates/email templates

design-documents/graph-ql/path-validator.md
viktym: I propose to introduce RouteValidatorInterface, so in the future, we will be able to eliminate inheritance from \Zend_Validate_Abstract without the necessity to support its public methods.

kandy: We had \Magento\Framework\Validator\ValidatorInterface, just need to implement it

design-documents/graph-ql/path-validator.md
cpartica: it's the standard https://tools.ietf.org/html/rfc3986#section-3.3

design-documents/graph-ql/path-validator.md
cpartica: yes, it seems that we need a new class that can be configured to be used in both url and path cases
Note that a path is not really any string, and there's some validations involved for delimiters that make sense for a path:
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Examples_2
https://stackoverflow.com/questions/2163803/what-is-the-semicolon-reserved-for-in-urls

PR #217, opened 2019-07-30
Add `@spi` annotation

buskamuza:
## Problem

Versioning rules and backward compatibility guidelines grow. See [Versioning](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/). It becomes hard to remember what version the change increase, as well as on which version an extension should depend.
More exceptional cases are added in order to maintain balance between keeping extensions stable and allowing necessary changes in core.

## Solution

Allow to distinguish between API and SPI code and mark the code accordingly.
Then simplify rules in the documentation based on what API and SPI changes lead to. As a result, we avoid necessity to specify exceptions for some parts of the code in the documentation.

Note: details are more focused on PHP part. If the concept makes sense, JavaScript part can also be updated.

## Requested Reviewers

@antonkril 
@maghamed 
@paliarush 
@akaplya 
@melnikovi 
@joni-jones 
@kokoc

### DIFF
--- a/design-documents/backward-compatibility/api-vs-spi.md
+++ b/design-documents/backward-compatibility/api-vs-spi.md
@@ -1,0 +1,78 @@
+# API vs. SPI
+
+## Background
+
+Magento codebase consists of pubic and private code.
+Public code is an interface for extension developers they can rely on, and core developers should preserve those interfaces for as long as possible.
+[Module version dependencies](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/dependencies.html) describes which version of a module an extension developer should depend on based on the code used in the extension.
+The principles for public code are based on whether the core code is used as API (application programming interface) or SPI (service provider interface).
+Due to commonly used practice of inheriting classes in Magento, many of them are considered both API and SPI because some extensions may use them in one of another way.
+This makes those classes stuck in time.
+It is impossible to add new methods to the class because it is a breaking change for an SPI class, and it is impossible to remove methods because it is a breaking change for an API class.
+
+From another side, Magento [Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html) discourage usage of inheritance ("2.6. Inheritance SHOULD NOT be used. Composition SHOULD be used for code reuse.").
+This means that only interfaces should be considered SPI, while classes should be considered API-only code.  
+
+## Problem Statement
+
+As we're trying to find the balance between keeping extensions as compatible with core as possible, while allowing core developers to make some necessary changes, [Codebase changes](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/codebase-changes.html) and [Module version dependencies](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/dependencies.html) describe more and more rules and exceptions.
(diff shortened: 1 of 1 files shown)


maghamed:
In fact, we preserve the API/SPI segregation right now, but currently, it's the responsibility of 3rd party developers who work with Magento code. If they `use` module's API they can make composer dependency on MAJOR version of the particular module, and if they extend or overwrite existing out of the box implementation, or use inheritance from Magento classes they suppose to have a dependency on MAJOR.MANOR version of a module.

I like the idea to consider PHP classes (all those which don't belong to layer super types list) as APIs, but not API + SPI. But instead of introducing `@SPI and @API` annotations I prefer to use `final` either keyword (if we will manage to deal with Interceptor code generation) or at least with PHP annotation which will just describe our intention. That will provide us an ability to introduce new public methods to classes, which sometimes may be useful.

I am not a fan of introducing `@api/@spi` for interfaces. Some time ago we considered using @api  for Services and Data interfaces accessible via Web API, as in Service Isolated (microservice) environment, `@api` describes services which could be called remotely via a wire, but `@spi` should be extended in the scope of the same OS thread. 
So, API/SPI segregation if any should take into account service decomposition. 

That's why my suggestion so far is to discuss `final` keyword/annotation for non-layer super type classes and not introduce SPI/API for interfaces. 
For API/SPI interfaces we need to prepare designated proposal covering Service Isolated environment restrictions.

design-documents/backward-compatibility/api-vs-spi.md
```diff
@@ -0,0 +1,78 @@
...
+      * lists (e.g., `CommandList`) can be marked as `@spi`
+   * all classes intended for extension are SPI
+ 
+## Scope of Work
+
+### Initial Demarcation
+
+Initially, mark the code as follows:
+
+* For all interfaces marked with `@api`, also add `@spi`. Interfaces are intended for implementation, so they are SPI. Most of the interfaces are also intended for usage, so they are API. There might be exceptions, but those can be fixed in the future or used for new interfaces.
```
joni-jones: Just my opinion, to have both `@api` and `@spi` annotations for the same interface/class looks more complicated at least because they have different versioning rules.

design-documents/backward-compatibility/api-vs-spi.md
```diff
@@ -0,0 +1,78 @@
...
+Guidelines should describe:
+* API
+   * changes in API increase MAJOR version of the module
+   * extension developers should depend on MAJOR version if they call API methods
+* SPI
+   * changes in SPI increase MINOR version of the module 
+   * extension developers should depend on MINOR version if they implement/extend SPI
+   * XML configuration files are SPI. Special cases for `di.xml`
+      * virtual types are private code by default. Can be marked as `@api` and/or `@spi`
+      * lists (e.g., `CommandList`) can be marked as `@spi`
```
joni-jones: I don't see the difference between `@api` and `@spi` in this example. If we introduce the `CommandList` with the intention for the future extension/customization why it can't be just `@api`?

buskamuza: `@api` means somebody will be calling it. I don't see why non-owner of the `CommandList` should be calling it.
`@spi` means somebody adds more items to the list. This is its intention.

PR #223, opened 2019-08-01
Translations in di.xml

buskamuza:
## Problem

There are `translate` attribute supported in layouts, while `di.xml` supports `translatable` with different behavior. This is confusing and may lead to bugs.

## Solution

Unify translation attributes in XML files.

## Requested Reviewers

@melnikovi 
@maghamed 
@kokoc

### DIFF
--- a/design-documents/i18n/translations-in-di.xml.md
+++ b/design-documents/i18n/translations-in-di.xml.md
@@ -1,0 +1,31 @@
+# Translations in `di.xml`
+
+## Current Situation
+
+Currently all XML files that support types, allow `translate` attribute for `string` type. This attribute leads to:
+
+1. The specified string to be translated when XML file is read, so the client of the string receives a translated string automatically.
+2. A translation tool recognizes such strings as translatable and collect them in a translation file. 
+
+`di.xml`, instead, supports `translatable` attribute which:
+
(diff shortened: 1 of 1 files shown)


design-documents/i18n/translations-in-di.xml.md
```diff
@@ -0,0 +1,31 @@
...
+2. Somebody can make a mistake (because attribute names sound very similar). Though the translation tool can recognize both attributes, so it will not cause any issue.
+
+## Solution
+
+Unify translation behaviors:
+
+1. Allow `translate` attribute in `di.xml`.
+2. `translate` attribute leads to the string being automatically translated during XML parsing and passed to the client code as translated phrase.
+   2.1. If it is impossible to translate the phrase (e.g., the application is running in `global` area), the phrase is provided untranslated.
+3. Remove `translatable` attribute. Temporarily, support it as deprecated with existing behavior.
```
nathanjosiah: This would remove the ability to defer the translation. Is this an acceptable feature loss?

buskamuza:
* @kandy : keep behavior as is. It may be impossible to implement due to `di.xml` being loaded too early
* Look into all cases where translations are provided via `di.xml` and see if we can eliminate translations from `di.xml` at all.

PR #231, opened 2019-08-15
Introduce Magento Admin FE Guidelines

jedmao:
## Problem

There are already a handful of React applications built for Magento Admin use, but they have different dependencies and versions between them and we don't yet have a micro front-end architecture to manage them.

## Solution (for now)

We must do our best to align common dependencies manually.

### DIFF
--- a/design-documents/frontend/admin-guidelines.md
+++ b/design-documents/frontend/admin-guidelines.md
@@ -16,76 +16,179 @@ For more information behind these technology choices, refer to the [Technical Vi
 
 ## package.json
 
-```jsonc
+Each section of the `package.json` file has been broken down into different topics.
+
+- [version](#version)
+- [scripts](#scripts)
+- [browserslist](#browserslist)
+- [dependencies](#dependencies)
+  - [Utility functions](#utility-functions)
+  - [SSR](#ssr)
+  - [React](#react)
+  - [State management](#state-management)
+  - [Code splitting](#code-splitting)
+  - [i18n](#i18n)
+  - [Theming](#theming)
+  - [Relay](#relay)
+  - [TypeScript import helpers](#typescript-import-helpers)
+- [devDependencies](#devdependencies)
+  - [@types](#types)
+  - [Redux mock store](#redux-mock-store)
+  - [Testing library](#testing-library)
+
+### version
+
+The version of this `package.json` follows [semver](https://semver.org/) and will change with this document accordingly.
+
+```json
+"version": "0.1.0",
+```
+
+### scripts
+
+Customize your CRA configuration with [craco][].
+
+- [Don't eject](https://create-react-app.dev/docs/alternatives-to-ejecting).
+- Avoid forking, because you can oftentimes do the same thing by creating your own [craco][] plugin. See [craco plugins on npm](https://www.npmjs.com/search?q=craco%20plugin) and the [craco recipes](https://github.com/sharegate/craco/tree/master/recipes) for guidance.
+
+```json
+"scripts": {
+  "start": "craco start",
+  "build": "craco build",
+  "test": "craco test"
+},
+```
+
+### [browserslist](https://github.com/browserslist/browserslist#readme)
+
+Don't forget your [polyfills](https://create-react-app.dev/docs/supported-browsers-features#supported-language-features)!
+
+```json
+"browserslist": [
+  ">0.2%",
+  "not dead",
+  "not ie <= 11",
+  "not op_mini all"
+],
+```
+
+### dependencies
+
+#### Utility functions
+
+Use vanilla (native) JS if possible, utility functions if not. [You don't (may not) need Lodash/Underscore](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#you-dont-may-not-need-lodashunderscore-). See also [just](https://github.com/angus-c/just#just).
+
+```json
+"@queso/camel-case": "^0",
+"@queso/kebab-case": "^1",
+```
+
+#### SSR
+
+Prepare for [SSR](https://medium.com/walmartlabs/the-benefits-of-server-side-rendering-over-client-side-rendering-5d07ff2cefe8) by using [universal fetching](https://www.npmjs.com/package/ky-universal). Don't use [axios](https://www.npmjs.com/package/axios) or [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) directly.
+
+```json
+"ky": "^0",
+"ky-universal": "^0",
+```
+
+#### React
+
+```json
+"react": "^16",
+"react-dom": "^16",
+"react-helmet": "^5",
+```
+
+#### Routing
+
+See [The Future of React Router and @reach/router](https://reacttraining.com/blog/reach-react-router-future/). Basically, the new API will be much closer to `@reach/router` than React Router, so this prepares us for that future.
+
+```json
+"@reach/router": "^1",
+```
+
+#### State management
+
+[Immutability Helper](https://www.npmjs.com/package/immutability-helper) was [rewritten in TypeScript](https://github.com/kolodny/immutability-helper/pull/123) and strikes a balance between features and size.
+
+```json
+"immutability-helper": "^3",
+"react-redux": "^7",
+"redux": "^4",
+"redux-thunk": "^2",
+```
+
+#### Code splitting
+
+For context, read [Optimize your React application with Loadable Components](https://www.smooth-code.com/articles/code-splitting-react-loadable-components) and don't use dynamic imports in your components.
+
+```json
+"@loadable/component": "^5",
+```
+
+#### i18n
+
+```json
+"react-intl": "^3",
+```
+
+#### Theming
+
+```json
+"react-theme-context": "^2",
+```
+
+#### Relay
+
+```json
+"babel-plugin-relay": "^5",
+"react-relay": "^5",
+```
+
+#### TypeScript import helpers
+
+```json
+"tslib": "^1",
+```
+
+Combined with the following TypeScript [compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html):
+
+```json
 {
-  "version": "0.1.0",
-
-  "scripts": {
-    // https://www.npmjs.com/package/@craco/craco
-    "start": "craco start",
-    "build": "craco build",
-    "test": "craco test"
-    // don't eject
-  },
-
-  // npm install --save-dev core-js
-  // https://github.com/browserslist/browserslist#readme
-  "browserslist": [
-    ">0.2%",
-    "not dead",
-    "not ie <= 11",
-    "not op_mini all"
-  ],
-
-  "dependencies": {
-
-    // Some utility functions (use native JS, don't install Lodash)
-    "@queso/camel-case": "^0",
-    "@queso/kebab-case": "^1",
-
-    // Universal fetching (don't use axios or fetch)
-    "ky": "^0",
-    "ky-universal": "^0",
-
-    // React
-    "react": "^16",
-    "react-dom": "^16",
-    "react-helmet": "^5",
-    "react-router-dom": "^5",
-
-    // Redux
-    "immutability-helper": "^3", // dont' use immer or immutable
-    "react-redux": "^7",
-    "redux": "^4",
-    "redux-thunk": "^2",
-
-    // Code splitting (don't use dynamic imports in components)
-    "@loadable/component": "^5",
-
-    // Internationalization
-    "react-intl": "^3",
-
-    // Theming
-    "react-theme-context": "^2",
-
-    // Relay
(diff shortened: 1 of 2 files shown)


design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Routing
+
+See [The Future of React Router and @reach/router](https://reacttraining.com/blog/reach-react-router-future/). Basically, the new API will be much closer to `@reach/router` than React Router, so this prepares us for that future.
+
+```json
+"@reach/router": "^1",
+```
+
+#### State management
```
bradgarropy: At this point I exclusively use React's `context`. Thoughts?

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,194 @@
...
+#### Utility functions
+
+Use vanilla (native) JS if possible, utility functions if not. [You don't (may not) need Lodash/Underscore](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#you-dont-may-not-need-lodashunderscore-). See also [just](https://github.com/angus-c/just#just).
+
+```json
+"@queso/camel-case": "^0",
+"@queso/kebab-case": "^1",
+```
+
+#### SSR
```
bradgarropy: [Next.js](https://nextjs.org/) recommends the use of [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic-unfetch) for universal [data fetching](https://nextjs.org/docs#fetching-data-and-component-lifecycle). Might be another option to consider as it works very similar to the built in `fetch`.

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,194 @@
...
+
+```json
+"react-theme-context": "^2",
+```
+
+#### Relay
+
+```json
+"babel-plugin-relay": "^5",
+"react-relay": "^5",
```
bradgarropy: I'll just start the Relay vs Apollo discussion here...
With Apollo's adoption of [React hooks](https://www.npmjs.com/package/@apollo/react-hooks) in its latest release, it seems like Apollo will continue to evolve alongside React.

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Redux mock store
+
+This is [a TypeScript fork](https://www.npmjs.com/package/@jedmao/redux-mock-store) of a somewhat stale, but highly successful package that doesn't require much maintenance.
+
+```json
+"@jedmao/redux-mock-store": "^2",
+```
+
+#### Testing library
```
bradgarropy: Good call here 👍🏼

design-documents/frontend/admin-guidelines.md
jedmao: If you can swing it, that's fantastic! It's still a bit of a new concept, however, so I don't want to take away Redux just yet. I'd be interested in reviewing some code you have to see what the pros & cons are at this point.

design-documents/frontend/admin-guidelines.md
jedmao: Yep. I've been down this 🐰 🕳 a few times now and I keep coming to the same conclusion. Take a look at [ky's usage[(https://www.npmjs.com/package/ky#usage) compared to plain fetch. Much more [readable and expressive](https://github.com/jedmao/tsx/blob/e7fb79851f217c7a05af91ba44368cb363e44d2d/src/config/environment.ts#L12-L19) and comes with 1st-class TypeScript support to boot! Happy to open another conversation on the matter though if you have some arguments to present.

design-documents/frontend/admin-guidelines.md
jedmao: I was definitely torn about this particular decision, but I knew it would be one or the other. I don't know enough about the internals, but my concern with Apollo is that (if I'm not mistaken) their strategy for consolidating queries and optimizing the payload to the user wasn't as efficient and opinionated as Relay. I'd be glad to be proven wrong on this point. Ultimately, I want the _user_ to have the best experience and performance possible!

I will say, however, that the Relay documentation is pretty bad compared to Apollo.

jedmao: Referencing the conclusion in [this article](https://www.prisma.io/blog/relay-vs-apollo-comparing-graphql-clients-for-react-apps-b40af58c1534#conclusion):

> Apollo optimizes for ease of use and flexibility, thus allowing developers to get started quickly with their applications. However, making sure that the local cache is kept consistent with Apollo requires lots of manual effort and, especially in larger applications, can get pretty complicated.

> Relay on the other hand is a very sophisticated system that is designed from the ground up to maximize efficiency in the communication with the server. React components express their data requirements using GraphQL fragments that are co-located to each component. This makes it easier to ensure that each component gets the data it needs even when data requirements for other components in the hierarchy are changing. Relay is doing a lot of work behind the scenes to keep its local store consistent with the state on the server. The power of Relay comes at the price difficult to grasp complexity and a slow learning curve.

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,198 @@
...
+```json
+"@reach/router": "^1",
+```
+
+#### State management
+
+[Immutability Helper](https://www.npmjs.com/package/immutability-helper) was [rewritten in TypeScript](https://github.com/kolodny/immutability-helper/pull/123) and strikes a balance between features and size.
+
+```json
+"immutability-helper": "^3",
```
deloreyj: [Immer](https://github.com/immerjs/immer) is another immutability library that's been gaining a lot of steam. It allows you to write all of your code in the normal way [].push(), mutating object properties (even nested ones), etc and then makes the changes immutably under the hood. The value of being able to just write normal mutations and not have to learn new APIs (especially one as ugly as immutability helper) is worth the extra 2.8 kb IMO. Check out the [benefits/usage section](https://github.com/immerjs/immer#benefits) for more examples. As a side note, Immer is the immutability library that the Redux team decided to ship with Redux Starter Kit.

design-documents/frontend/admin-guidelines.md
jedmao: That's why I highlighted the article's date. I saw the F8 talk, which is what piqued my interest in Relay. I can confirm that the DX with Relay is not at all friendly, so if Apollo is better documented, it might be the right choice. Again, I was torn on this decision. It wasn't easy for me to find any fresh comparisons between what we saw at F8 and what Apollo currently offers. I'd really like to know the feature parity between them and the trajectories. I'm not sure that it matters though. Take a look at this!

Screen Shot 2019-08-23 at 1 36 31 PM

I'm pretty sure that spike lines up with the F8 talk and tells us pretty well how the community received it. Now compare that to the `apollo-client`:

Screen Shot 2019-08-23 at 1 36 12 PM

## Conclusion

I'm 100% OK changing this to the Apollo Client. Thanks for challenging this one!

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,194 @@
...
+```json
+"version": "0.1.0",
+```
+
+### scripts
+
+Customize your CRA configuration with [craco][].
+
+- [Don't eject](https://create-react-app.dev/docs/alternatives-to-ejecting).
+- Avoid forking, because you can oftentimes do the same thing by creating your own [craco][] plugin. See [craco plugins on npm](https://www.npmjs.com/search?q=craco%20plugin) and the [craco recipes](https://github.com/sharegate/craco/tree/master/recipes) for guidance.
```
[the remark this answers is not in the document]
jedmao: That's exactly what I'm challenging you to do here. I'm actually going to create a craco plugin for Linaria today if I have enough time.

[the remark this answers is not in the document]
deloreyj: Challenge accepted. I'll see if I can get some time on the board on in our grooming meeting Monday to work on this.

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,194 @@
...
+  "not ie <= 11",
+  "not op_mini all"
+],
+```
+
+### dependencies
+
+#### Utility functions
+
+Use vanilla (native) JS if possible, utility functions if not. [You don't (may not) need Lodash/Underscore](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#you-dont-may-not-need-lodashunderscore-). See also [just](https://github.com/angus-c/just#just).
```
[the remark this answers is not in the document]
jedmao: I can see how you need some clarification here.

1. We should not be loading the entire Lodash library on the client, which I think is obvious for many years now.
1. You shouldn't be installing utility functions, Lodash or otherwise, if you can get the same functionality in vanilla JS.
1. If you truly need some utility functions, it's probably only a handful of them. I don't think it makes sense to hoist these dependencies in a Micro FE environment, should that ever be introduced, so you can install whatever you want. My personal preference is to install something with TypeScript support.
   - Lodash has 3rd-party TS support.
   - [Just](https://github.com/angus-c/just#just) has "some" 3rd-party contributions, but [never plans to](https://github.com/angus-c/just/issues/126#issuecomment-470807747) support it officially.
   - [Queso](http://queso.surge.sh/) has 1st-class TS support and documentation, but is a new project of my own making (on paternity leave). I haven't even announced it, but it's super easy to contribute to and fully tested.

design-documents/frontend/admin-guidelines.md
jedmao: All good points. I've been holding onto this one for a while, but I think it's time to "let it go." Let me update this PR to recommend Immer. Thanks for challenging this point!

jedmao: Addressed in https://github.com/magento/architecture/pull/231/commits/b1c19bc24ff6a13764b7aa3d7e9a2c64daaebb16

design-documents/frontend/admin-guidelines.md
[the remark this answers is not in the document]
jedmao: Addressed in https://github.com/magento/architecture/pull/231/commits/b1c19bc24ff6a13764b7aa3d7e9a2c64daaebb16

design-documents/frontend/admin-guidelines.md
jedmao: Addressed in https://github.com/magento/architecture/pull/231/commits/b1c19bc24ff6a13764b7aa3d7e9a2c64daaebb16

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,215 @@
+# Magento Admin Guidelines
+
+## Problem
+
+There are already a handful of React applications built for Magento Admin use, but they have different dependencies and versions between them and we don't yet have a micro front-end architecture to manage them.
```
roma-glushko: @jedmao does it mean that upgrade of Magento Admin frontend (currently, UI component-based) is somewhere on the horizon?

jedmao: There are no plans for that as of yet.

design-documents/frontend/admin-guidelines.md
```diff
@@ -0,0 +1,215 @@
...
+// pending https://github.com/nfl/react-helmet/issues/465
+"react-helmet-async": "^1",
+```
+
+#### Routing
+
+See [The Future of React Router and @reach/router](https://reacttraining.com/blog/reach-react-router-future/). Basically, the new API will be much closer to `@reach/router` than React Router, so this prepares us for that future.
+
+```json
+"@reach/router": "^1",
```
deloreyj: Have you found a way you like to make @reach/router support hash routing?

jedmao: No. Is that an important feature for you?

deloreyj: I think so. We'll need client-side routing within the React app that doesn't interfere with Magento's built-in server-side routing. I would use a hash router for that. Did you have a different approach in mind?

jedmao: Short answer is no, has routing is not supported and seeing as react-router is going to merge with reach router, I guess we'll have to see if it survives the merge. Take a look at [this thread](https://github.com/reach/router/issues/25) that suggests using a memory router; though, also points out the fact that it won't survive a page refresh.

jedmao: There are two big reasons to use reach router:
1. The API is most compatible with the future merge of react router.
1. Nested routes are way nicer.

That said, if you really can't make it work with reach router, I guess you'll have to use react router (for now), understanding that the API will change significantly and you might even lose your hash routing capabilities, preventing you from upgrading it.

No great answer here. It's a double edged ⚔️

deloreyj: Yeah I'd read that thread when I was trying to figure out how to solve this issue. I think the value of having the hash router now is worth a slightly more involved migration later on to my team.

PR #233, opened 2019-08-19
MySQL 8: auto_increment

buskamuza:
## Problem

MySQL 8 added caching layer on top of table status, which is refreshed from time to time (default `86400` = `1 day`).
This leads to the behavior where returned `auto_increment` value is `null` even though it shouldn't be (there is an auto_increment column in the table).
The table status information is not calculated until the table is accessed, and (probably) the expiration time is not passed.

## Solution

1. Disable MySQL cache for table status
2. Remove auto_increment from DB adapter interface

### DIFF
--- a/design-documents/database/mysql-8-status-support.md
+++ b/design-documents/database/mysql-8-status-support.md
@@ -14,8 +14,9 @@ The table status information is not calculated until the table is accessed, and 
 
 ### Real Use Cases
 
-1. Magento native import relies on auto_increment for entity relations. To do: get confirmation from Stepan Furman that import is broken on MySQL 8.
-2. Extensions may rely on the adapter interface for their scenarios.
+1. Magento native import relies on auto_increment for entity relations. See `\Magento\ImportExport\Model\ResourceModel\Helper::getNextAutoincrement()`.
+2. Mview. See `\Magento\Framework\Mview\View\Changelog::getVersion()`.
+3. Extensions may rely on the adapter interface for their scenarios.
 
 ## Possible Solutions
 
--- a/design-documents/database/mysql-8-status-support.md
+++ b/design-documents/database/mysql-8-status-support.md
@@ -1,0 +1,56 @@
+# MySQL 8 - Table Status Support
+
+## Background
+
+Magento DB adapter interface provides method for obtaining table status: `\Magento\Framework\DB\Adapter\AdapterInterface::showTableStatus()`.
+For MySQL adapter it's implemented as `SHOW TABLE STATUS FROM <db> LIKE <table>`.
+Note: the interface doesn't prescribe what data and in which format can be returned.
+
+## Problem Statement
+
+MySQL 8 added caching layer on top of table status, which is refreshed from time to time (default `86400` = `1 day`).
+This leads to the behavior where returned `auto_increment` value is `null` even though it shouldn't be (there is an auto_increment column in the table).
+The table status information is not calculated until the table is accessed, and (probably) the expiration time is not passed.
+
+### Real Use Cases
+
+1. Magento native import relies on auto_increment for entity relations. To do: get confirmation from Stepan Furman that import is broken on MySQL 8.
+2. Extensions may rely on the adapter interface for their scenarios.
+
+## Possible Solutions
+
+### Remove auto_increment from the interface
+
+Reasoning: application should not deal with auto_increment, nor rely on it.
+
+This decision will require reworking of import. This may introduce performance degradation for import.
+
+This is a breaking change. Because we don't have clear table status interface, docblock and devdocs documentation are the places for communication to developers.
+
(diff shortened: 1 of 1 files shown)


Stepa4man:
MySQL (not)issue: https://bugs.mysql.com/bug.php?id=91038

image

### Steps to reproduce with MySQL-8:
```
# 1. Create a table with auto_increment
CREATE TABLE `test`
(
    `test_id` int(10) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Test ID',
    PRIMARY KEY (`test_id`)
) ENGINE = InnoDB
  DEFAULT CHARSET = utf8 COMMENT ='Test Table';

# 2. Check auto_increment
SHOW TABLE STATUS LIKE 'test';
# auto_increment = null, expected auto_increment = 1

# 3. Insert values
INSERT INTO `test` VALUES (null), (null), (null);

# 4. Check actual max ID
SELECT max(`test_id`) FROM `test`;
# returns 3 as expected

# 5. Check auto_increment
SHOW TABLE STATUS LIKE 'test';
# auto_increment = null, expected auto_increment = 4 (value+1)

# 6. Drop table
DROP TABLE `test`;

# 7. Disable caching
SET @@SESSION.`information_schema_stats_expiry` = 0;

# 8. Create a table with auto_increment
CREATE TABLE `test`
(
    `test_id` int(10) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Test ID',
    PRIMARY KEY (`test_id`)
) ENGINE = InnoDB
  DEFAULT CHARSET = utf8 COMMENT ='Test Table';

# 9. Check auto_increment
SHOW TABLE STATUS LIKE 'test';
# still returns auto_increment = null, expected auto_increment = 1

# 10. Run a select query to update stats in the information schema
SELECT * FROM `test`;

# 11. Check auto_increment
SHOW TABLE STATUS LIKE 'test';
# still returns auto_increment = 1 as expected

# 12. Insert values
INSERT INTO `test` VALUES (null), (null), (null);

# 13. Check actual max ID
SELECT max(`test_id`) FROM `test`;
# returns 3 as expected

# 14. Check auto_increment
SHOW TABLE STATUS LIKE 'test';
# auto_increment = 4, as expected (value+1)
```

 _Please, take into in account_ that even with setting `information_schema_stats_expiry=0` newly created table can have `auto_increment` = null at `information_schema`.`TABLES` until it will be used at least somehow. Check out steps 8-11


### Usages

- \Magento\Framework\DB\Adapter\Pdo\Mysql::showTableStatus
- \Magento\Framework\Mview\View\Changelog::getVersion

dhorytskyi:
Another possible solution is to analyze the table before getting the table status.

`
CREATE TABLE `test` (
  `id` int NOT NULL AUTO_INCREMENT,
  `value` varchar(255),
  PRIMARY KEY(`id`)
);

SHOW TABLE STATUS WHERE `name` LIKE 'test';

INSERT INTO `test` (`value`) VALUES ('v1');
INSERT INTO `test` (`value`) VALUES ('v1');
SHOW TABLE STATUS WHERE `name` LIKE 'test';

ANALYZE TABLE `test`;
SHOW TABLE STATUS WHERE `name` LIKE 'test';
`

https://www.db-fiddle.com/f/vgPjLZnkrvoPJ6pLaBJop/0

Stepa4man:
> Another possible solution is to analyze the table before getting the table status.
> 
> CREATE TABLE test (
> id int NOT NULL AUTO_INCREMENT,
> value varchar(255),
> PRIMARY KEY(id)
> );
> 
> SHOW TABLE STATUS WHERE name LIKE 'test';
> 
> INSERT INTO test (value) VALUES ('v1');
> INSERT INTO test (value) VALUES ('v1');
> SHOW TABLE STATUS WHERE name LIKE 'test';
> 
> ANALYZE TABLE test;
> SHOW TABLE STATUS WHERE name LIKE 'test';
> 
> https://www.db-fiddle.com/f/vgPjLZnkrvoPJ6pLaBJop/0

Hi @dhorytskyi,
Thanks for your suggestion.
I believe it can be even a better solution for this case.

design-documents/database/mysql-8-status-support.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Real Use Cases
+
+1. Magento native import relies on auto_increment for entity relations. See `\Magento\ImportExport\Model\ResourceModel\Helper::getNextAutoincrement()`.
+2. Mview. See `\Magento\Framework\Mview\View\Changelog::getVersion()`.
+3. Extensions may rely on the adapter interface for their scenarios.
+
+## Possible Solutions
+
+### Remove auto_increment from the interface
```
melnikovi: What if we deprecate the method on the interface and use [this](https://github.com/magento/architecture/pull/233#issuecomment-524468006) approach to avoid introducing separate adapter for MySQL 8 where we would change implementation of the method to clear table cache? Need to do performance testing. If ANALYZE TABLE is slow, then introduce adapter for MySQL 8.

PR #254, opened 2019-08-30
Contract for JSON encoding/decoding

lenaorobei:
Since I forgot about https://github.com/magento/architecture/pull/161 and force-pushed to my `master`, creating new PR.

Hopefully this PR will get the second wind 🙂 
## Problem
The use of JSON data format might be wider than just serialization and current `Magento\Framework\Serialize\Serializer\Json` does not cover all those needs.

## Solution
New contract for JSON encoding/decoding.

## Requested Reviewers
@maghamed
@Vinai
@fooman

### Related Discussion
https://github.com/magento/magento-coding-standard/pull/91

### DIFF
--- a/design-documents/data-format/json.md
+++ b/design-documents/data-format/json.md
@@ -1,0 +1,245 @@
+# Contract for JSON encoding/decoding
+
+This proposal is based on [Design Document for changing SerializerInterface](https://github.com/magento-engcom/msi/wiki/Design-Document-for-changing-SerializerInterface) written by [Igor Miniailo](https://github.com/maghamed).
+
+## Current state
+
+Magento 2.2 changed the default serialization implementation in favour of JSON encoding in order to make code more secure and avoid vulnerabilities related to PHP `unserialize` function.
+
+For this purpose `SerializerInterface` was introduced.
+
+``` php
+/**
+ * Interface for serializing
+ *
+ * @api
+ * @since 100.2.0
+ */
+interface SerializerInterface
+{
+    /**
+     * Serialize data into string
+     *
+     * @param string|int|float|bool|array|null $data
+     * @return string|bool
+     * @throws \InvalidArgumentException
+     * @since 100.2.0
+     */
+    public function serialize($data);
+
+    /**
+     * Unserialize the given string
+     *
+     * @param string $string
+     * @return string|int|float|bool|array|null
+     * @throws \InvalidArgumentException
+     * @since 100.2.0
+     */
+    public function unserialize($string);
+}
+```
+
+Its default implementation of `Magento\Framework\Serialize\Serializer\Json` became an independent contract for all the business logic where `json_encode`/`json_decode` can be applied.
+
+``` php
+/**
+ * Serialize data to JSON, unserialize JSON encoded data
+ *
+ * @api
+ * @since 100.2.0
+ */
+class Json implements SerializerInterface
+{
+    /**
+     * @inheritDoc
+     * @since 100.2.0
+     */
+    public function serialize($data)
+    {
+        $result = json_encode($data);
+        if (false === $result) {
+            throw new \InvalidArgumentException("Unable to serialize value. Error: " . json_last_error_msg());
+        }
+        return $result;
+    }
+
+    /**
+     * @inheritDoc
+     * @since 100.2.0
+     */
+    public function unserialize($string)
+    {
+        $result = json_decode($string, true);
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw new \InvalidArgumentException("Unable to unserialize value. Error: " . json_last_error_msg());
+        }
+        return $result;
+    }
+}
+```
+
+## Problem overview
+JSON as data format can be used in other different from serialization contexts so `Magento\Framework\Serialize\Serializer\Json` class implementation is not sufficient enough to cover all those scenarios, e.g. it is impossible to pass `$options` or `$depth`.
+
+Native PHP [json_encode](https://www.php.net/manual/en/function.json-encode.php) function:
+``` php
+string json_encode ( mixed $value [, int $options = 0 [, int $depth = 512 ]] )
+```
+
+Native PHP [json_decode](https://www.php.net/manual/en/function.json-decode.php) function:
+``` php
+mixed json_decode ( string $json [, bool $assoc = FALSE [, int $depth = 512 [, int $options = 0 ]]] )
+```
+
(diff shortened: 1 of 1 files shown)


Vinai:
My previous comment as written on the thread for #161 still applies:
https://github.com/magento/architecture/pull/161#issuecomment-492131243:

-----
In a nutshell, I'm for *alternative solution 1* - keep the `Json` class and add an optional `$option` parameter. From a cost/benefit perspective that seems like the best solution to me.

Introducing a `JsonEncoderInterface` to me is using abstraction for a stable part of the system. Abstraction makes the most sense in areas that change a lot. The less something changes, the more concrete it should be.
In this case, chances are high there will never be an alternative Json implementation. It is coupled to the low level PHP function signature due to the options parameter syntax. For this reason I feel introducing a new interface is not required, in fact - it introduces more code when less code is sufficient. This violates Kent Beck's fourth rule of simple code: fewest possible elements.

Using virtual types is even more complex. I find that solution harder to understand as I would need to search di.xml to seem which options are set for a given type. It also puts additional weight on developers because they have to understand virtual types, arguably one of the more complex areas of object manager configuration. Keeping things simpler is usually better.

Personally I still will probably continue to use PHP native `json_encode` and `json_decode` functions when possible, but I do get the motivation behind wanting to protect the platform against vulnerabilities in code you don't control. For core development using the `Json` class seems like the right choice.
One of the things I struggle with when doing Magento development is that the whole system feels heavy and clunky, and isolating my own code against Magento by using native PHP functions and custom classes helps alleviate this.

-----

navarr:
An interface might be useful for alternative implementations - such as [JSON5](https://json5.org/)

Vinai:
@navarr The point of using a known interchangeable serialization format like JSON is that external tooling besides Magento can use the data. It needs to be dependable. Switching to a different format like JSON5 would break all that. For example JSON5 doesn’t work with MySQL, Postgres, Mongo or other storage engines with JSON support. I think it’s not a good scenario.
JSON is like the lowest common denominator of serialization formats, even though much better ones exist. But that is actually one of the reasons it’s successful - JSON is understood pretty much everywhere.

JSON and the other formats are standardized, so switching implementations will not change the interpretation of the data. Switching the format is a huge breaking change.
Using a different serialization format like JSON5 or EDN as a default would be fine, too, but not by switching out the interface implementation, but rather by using a different concrete class and having the code depend on that, e.g. EdnSerializer or Jsin5Serializer.
The data serialization format is important and needs to be dependable.

design-documents/data-format/json.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+On the other hand native PHP implementation requires us to call `json_last_error` in order to handle errors instead of throwing an exception. It was [changed in PHP 7.3](https://wiki.php.net/rfc/json_throw_on_error), but this is still not the default behaviour.
+
+## Solution
+
+Since we want to protect our platform from any possible security vulnerabilities related to the code we do not control and have expected implementation we propose to add new dedicated contract for JSON.
+
+``` php
+interface JsonEncoderInterface
```
ihor-sviziev: I would like to add an ability to use it as viewModel in phtml files but extending `\Magento\Framework\View\Element\Block\ArgumentInterface`. What do you think?

design-documents/data-format/json.md
```diff
@@ -0,0 +1,245 @@
...
+#### Pros
+- Correct from OOP point of view as we have two independent interfaces for two different business operations.
+- Follows the Interface Segregation Principle (SOLID).
+- Get us closer to the desired state of serializer/encoding design.
+- Proper error handling.
+
+#### Cons
+- Backward incompatible as introduces a lot of refactoring substituting existing direct usages of JSON object with newly introduced contract `JsonEncoderInterface`.
+
+### Other possible solutions
```
melnikovi: Discussed with @Vinai and @buskamuza and narrowed down to 2 options.

Deprecated common interface in the patch release. Remove common interface in minor release and add optional parameter to Json class.

Deprecate common interface. Do not remove interface in the minor release, keep it for some time. Introduce new methods SerializeWithOptions/UnserializeWithOptions in Json class.

Could you please update proposal with these options, so we can discuss them on the next architecture meeting?

melnikovi: Discussed on ad hoc meeting with @paliarush and @joni-jones and decided to deprecate interface in the next minor release, provide new JsonSerializer with serialize/unserialize methods and in minor release after next remove the interface.

Vinai: Will the deprecated interface stay around, or are you still planing to remove it later? If so, when?
What is the benefit of providing a new `JsonSerializer` class over adding two methods to the existing `Json` class? Will the `Json` class be deprecated, too?

From my perspective using the existing `Json` class is preferable, instead of introducing a new class.
To allow specifying serialization options, two new dedicated methods should be added. This keeps backward compatibility and provides a clean interface.

The biggest benefit is that all the existing code that uses the `Json` class does not need to be refactored. This kind of refactoring is useless busy work that doesn't add any value in return.
Another benefit is that it's less confusing to have a single implementation rather then two, even if one of them is deprecated.

Personally I find a class with the methods `Json::serialize` and `Json::serializeWithOptions`  (plus matching deserialization methods) far more expressive and clean compared to a class `JsonSerializer::serialize` with optional options arguments.

Anyway, just my .02$. I really feel I'm repeating myself.
In a nutshell, I advocate avoiding useless bc breaks (to avoid angering developers building on the platform), and simple is better.

fooman:
I am with @Vinai  and I think this discussion might be a good starting point for broadening it.

The current Magento approach is to avoid adding to existing interfaces as this would break other  implementations of said interface.

Instead we are introducing a new interface and deprecate the old (it's really just replacing one way of breaking with a different one).

```
/** @deprecated */
interface ExistingInterface
{
    public methodOne();
}

interface NewInterface
{
    public methodOne();
    public MethodTwo();
}

class DefaultClass implements NewInterface
{
}
```

On balance I believe this is optimising for the wrong audience. I am convinced that there are more users of `ExistingInterface` than there are alternative implementations of `DefaultClass` that do not look like this:

```
class ReplacedClass extends DefaultClass
{
}
```

I think we should take a look at the polices and approach so that we can end up with:

```
interface ExistingInterface
{
    public methodOne();
    public MethodTwo();
}
```
as that would create less work over all.

lenaorobei:
@Vinai @fooman thank you for the feedback, your arguments were compelling enough and I updated the proposal.

Now the proposal states:
- Deprecate the `Magento\Framework\Serialize\SerializerInterface` and add annotation `@see use Magento\Framework\Serialize\Serializer\Json instead`.
- Add two new methods to the` Magento\Framework\Serialize\Serializer\Json`. Implementation will use low-level PHP `json_encode`/`json_decode` under the hood and will throw `InvalidArgumentException` if encoding/decoding cannot be performed.

Vinai:
Thanks for listening!

PR #261, opened 2019-09-06
Implement PSR-11: Container Interface

tdgroot:
## Problem
PSR-11 is not implemented. It would be relatively easy to do so, while providing compatibility for general purpose PHP packages.

## Solution
Haven't seen an implementation around yet, but I think I could provide one, if necessary. Let me know :)

## Requested Reviewers
- @kandy 
- @buskamuza

### DIFF
--- a/design-documents/implement-psr-11-container-interface.md
+++ b/design-documents/implement-psr-11-container-interface.md
@@ -1,0 +1,42 @@
+# Implement PSR-11: Container Interface
+PSR-11 is a set of basic interfaces to standardize dependency injection (obtaining objects and parameters) across frameworks and libraries. 
+
+## Why?
+Say a general purpose PHP package makes use of the PSR-11 Container Interface(CI) to instantiate objects in a dependency injection(DI) friendly manner:
+
+``` php
+<?php
+
+namespace Acme\SomeComponent;
+
+class SomeService
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
+
+    public function doSomething(string $message)
+    {
+        ...
+
+        if ($this->container->has('\\Psr\\Log\\LoggerInterface')) {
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=261) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=261) before we can accept your contribution.<br/>

buskamuza:
Looks straightforward for service classes.
Could you please also cover factories?

Do we preserve both `get()` and `create()` method of Object Manager, while making it implement `Psr\Container\ContainerInterface`? So `create()` method won't be part of the interface, but will be supported for factories, so factories should continue depending on `\Magento\Framework\ObjectManagerInterface`.

Actually it would help with static testing, we could limit non-factory classes to using `Psr\Container\ContainerInterface`, while allow `\Magento\Framework\ObjectManagerInterface` in factories.

A breaking change will be adding `has()` method to `\Magento\Framework\ObjectManagerInterface`.

Could you please address these points in the document based on your vision?

akaplya:
> A breaking change will be adding `has()` method to `\Magento\Framework\ObjectManagerInterface`.
Actually, ObjectManagerInterface should not necessarily implement ContainerInterface it could be a wrapper, right? Adding a new interface will be a pretty MINOR change.

Usage of the container, the same as the usage of object manager, will be not recommended inside of service classes and will require a corresponding change in static tests.

Vinai:
After a little discussion on slack I understand this would be mainly for providing Magento objects to third party non-Magento libraries that use PSR-11.
Is there a concrete requirement to do that at the moment?
If not, I think it would be better to hold off adding PSR-11 support in this way, because

- YAGNI
- The fourth rule of simple code by Kent Beck (fewest possible elements)
- It is simple to add later when needed, e.g. no choices within Magento depend on it

If there is a concrete requirement for providing PSR-11 support, it would be good to add it to the description in the PR.

tdgroot:
In the light of Vinai's input I will close this issue.

PR #263, opened 2019-09-06
GraphQL versioning proposal

paliarush:
GraphQL versioning proposal.

## Requested Reviewers

@buskamuza 
@maghamed 
@mslabko

### DIFF
--- a/design-documents/graph-ql/versioning.md
+++ b/design-documents/graph-ql/versioning.md
@@ -6,7 +6,7 @@ Some GraphQL schema changes can be covered using versionless strategy, which [is
  - new fields in return types
  - deprecation and subsequent removal of fields in return types
  - new optional arguments of the fields, queries and mutations
- - replacement of existing query or mutation with a new semantically different one. Example is replacing a set of mutations for adding products to cart with a single unified mutation which can add products of any type to cart. Semantics of the operations is different in this case, that is why it is not difficult to come up with a new meaningful name.
+ - replacement of existing query or mutation with a new semantically different one. Example is replacing a set of mutations for adding products to cart with a single unified mutation which can add products of any type to cart. Semantics of the operations is different in this case, that is why it is not difficult to come up with a new meaningful name. The old mutations and queries should be deprecated and eventually removed.
  
 Sometimes a new query which is meant to replace existing one has the same semantics and thus it is difficult to give the new query meaningful name. Example is new `products` query, which has a set of filters incompatible with existing implementations, thus it is impossible to comply with backward compatibility and deprecation policies.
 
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/versioning.md
```diff
@@ -0,0 +1,28 @@
+## Problem 
+
+GraphQL APIs are constantly evolving and require changes to be made to existing queries, mutations and data types.
+
+Some GraphQL schema changes can be covered using versionless strategy, which [is considered best practice](https://graphql.org/learn/best-practices/#versioning) for GraphQL APIs. Examples of such changes include:
+ - new fields in return types
+ - deprecation and subsequent removal of fields in return types
+ - new optional arguments of the fields, queries and mutations
+ - replacement of existing query or mutation with a new semantically different one. Example is replacing a set of mutations for adding products to cart with a single unified mutation which can add products of any type to cart. Semantics of the operations is different in this case, that is why it is not difficult to come up with a new meaningful name.
```
buskamuza: This items should probably mention that the old mutation should be first deprecated (and continue being supported) and then removed (same as in item 2).

design-documents/graph-ql/versioning.md
```diff
@@ -0,0 +1,28 @@
...
+ 
+Sometimes a new query which is meant to replace existing one has the same semantics and thus it is difficult to give the new query meaningful name. Example is new `products` query, which has a set of filters incompatible with existing implementations, thus it is impossible to comply with backward compatibility and deprecation policies.
+
+## Proposed solution
+
+It is proposed to suffix names of the new fields/mutations/queries which are semantically close to existing ones with  `v2`, `v3`, ..., `vX`. In the example above, a new query will be named `productsV2`, while existing `products` should be deprecated. Note that version suffixes should start from `v2`, while suffixless fields will be considered `v1` by default.
```
jedmao:
```suggestion
It is proposed to suffix names of the new fields/mutations/queries which are semantically close to existing ones with version numbers. In the example above, a new query would be named `products2`, while existing `products` should be deprecated. Note that version suffixes should start from version `2`, incrementing up until all deprecations are safely removed, after which the suffix should be removed in the next version (e.g., back to `products`).
```
1. You are using a lowercase `v` in your `vX` examples, but an uppercase `v` in `productsV2`.
1. I'm not even sure that the `v` holds any value. Wouldn't `products2` convey the same concept? Numbers are so rare in field names, I don't see it being an issue.
1. Once previous versions have ended their life and been removed, consider removing the version number entirely in the next release (e.g., `products3` becomes just `products` again).
   - Assuming we have verified _nobody_ is using `products` and `products2` any more.
   - Much easier for the human mind to track.
   - Once things settle, all the version numbers should be removed.

paliarush: 1. Lower vs upper case depends on wether we are talking about query (camel-cased) or field (snake-cased)
2. I personally feel like it is clearer with the `v` in the suffix, but it is subjective

PR #270, opened 2019-09-11
Align information in relation to new architecture

nuzil:
## Description

I aligned information about architectural vision of Asynchronous Import.

## Reviewers 
@naydav @paliarush

### DIFF
--- a/design-documents/asynchronous-import/base-extension.md
+++ b/design-documents/asynchronous-import/base-extension.md
@@ -202,17 +202,18 @@ Start Import
         "validation_strategy": "string",
         "allowed_error_count": 0,
         "import_image_archive": "string",
-        "import_images_file_dir": "string"
-    },
-    "convertingRules": [
+        "import_images_file_dir": "string",
+	"convertingRules": [
           {
               "name": "string",
-              "paramethers": [
+              "parameters": [
                   ..... free array format here .....
               ],
               "sort": "string",
           }
-    ]
+    	]
+    }
+    
 }
 ```
 
@@ -232,7 +233,7 @@ Start Import
 | Key | Value |
 | --- | --- |
 | name | Name of the rule - will be defined in configuration xml |
-| paramethers | array of paramethers. Flexible array here, will depend from method used |
+| parameters | array of paramethers. Flexible array here, will depend from method used |
 | sort | Sorf order of rules execution |
 
 
--- a/design-documents/asynchronous-import/base-extension.md
+++ b/design-documents/asynchronous-import/base-extension.md
@@ -1,16 +1,19 @@
-# Phase 1
+# Phase 1 - MVP
 
 With phase 1 we are planning to develop next functionality: 
-- Endpoint to receive file
-- Endpoint to start processing
-- Endpoint to receive status
+- Retrieving source data endpoints
+    * Source Retrieving: (local, remote file, base64 encoded)
+- Import process
+    * Import of Product Advanced Prices
+    * Import of Stock / Sources 
+    * Simple products import
 
-## File Upload Endpoint
-### Source Upload endpoint
+## Source management
+### Retrieving source data endpoints
 
-So import will start from uploading import Source file. Currently we will support "csv" files format
+So import will start from uploading import Source file. Currently we will support "csv" files format only
 
-POST  `/V1/import/source/csv`
+POST  `/V1/import/sources/csv`
  
 This request can accept files from different sources:
 - Local file path
@@ -23,16 +26,15 @@ Path is relative from Magento Root folder
 
 ```
 {
-  "source": {
-      "import_data": "var/catalog_product.csv",
-      "import_type": "local_path",
-      "uuid": "UUID",
-      "format": {
-          "csv_separator": "string",
-          "csv_enclosure": "string",
-          "csv_delimiter": "string",
-          "multiple_value_separator": "string"
-      }
+  "uuid": "UUID",
+  "source_data": {
+    "source_type": "local",
+    "source_data": "var/catalog_product.csv" 
+  } 
+  "format": {
+    "csv_separator": "string",
+    "csv_enclosure": "string",
+    "multiple_value_separator": "string"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


design-documents/asynchronous-import/base-extension.md
```diff
@@ -23,16 +26,15 @@ Path is relative from Magento Root folder
...
-          "csv_enclosure": "string",
-          "csv_delimiter": "string",
-          "multiple_value_separator": "string"
-      }
+  "uuid": "UUID",
+  "source_data": {
+    "source_type": "local",
+    "source_data": "var/catalog_product.csv" 
+  } 
+  "format": {
```
tariqjawed83: I think we should have format_type like csv, xml etc.; I do understand we will currently only support csv.
Instead of calling "csv_separator", we can say use "delimiter", which is commonly understood term. And we could also remove "csv_" from beginning.

tariqjawed83: https://www.php.net/manual/en/function.fgetcsv.php

we can also look at some optional parameters in the link above, for instance "escape, enclosure, length"; some of them are very commonly used in file-upload scenarios.

design-documents/asynchronous-import/base-extension.md
```diff
@@ -23,16 +26,15 @@ Path is relative from Magento Root folder
 
 ```
 {
```
tariqjawed83: do we think that this can be used in multi-select or multi-files upload?

design-documents/asynchronous-import/base-extension.md
nuzil: About "csv_" - agree here, as we have type defined in URL, we can use just params as "delimiter", "enclosure" and I will add "escape", make sense.
Length I guess its not really required param in this case. Cause we are basically defining how we will read file, which parts of file, etc. I never had a real usage of "length" param in csv import workflows.

design-documents/asynchronous-import/base-extension.md
nuzil: Not really. Again it was build based on practical experience. One file transfer already can be quite resource consuming. Multiple its already risky.
If customer want to have multiple files upload he has BULK API available for this case

design-documents/asynchronous-import/base-extension.md
```diff
@@ -207,70 +160,88 @@ Will return list of Source that was uploaded before.
 }
 ```
 
-## Start File Import Endpoint
+### Base CSV Parser
```
tariqjawed83: Usually in the streaming based file import process, we take the records from the data stream and process them as we are fetching, parsing and pushing those records. Is there a specific customer requirement, to separate the upload and import from each other. Because this design will inherently make the process slower, since you have to upload the file first, and then when it's all completed, only then you will start the actual import process. From customer perspective, they want to import the data into Magento as quickly as possible. 

On the other hand, since uploading might involve network latency and disk read (I/O latency), it is usually time consuming process, it is usually advisable to use streaming based approach end to end; meaning since lot of time spend waiting on I/O, CPU can be better off continue processing what it has already in-hand; so as you get records you can parse and push them simultaneously.

Since ultimate import will be based on Queue Processing, it should not cause dead-lock issues.

design-documents/asynchronous-import/base-extension.md
tariqjawed83: ok, I was just thinking from UI perspective, sometimes we have multi-select option enabled for small files. But in this case it can be risky and may overcomplicate.

design-documents/asynchronous-import/base-extension.md
nuzil: We had a long and multiple discussion related on this topic.
Main idea of split process is to separate 1 actions that can be independent from each other.
We are talking about decomposition of processes and this approac have multiple profits.
1) We have own set of imported data, so we can always see which data were uploaded, when and by whom
2) Each file can be run independently, so you can restart import for a file as many times as you want, define different mapping for same file, fix some errors in your mapping or processing rules without requirements to re-upload whole file again and again
3) Its is possible, and there are a cases where same file can be used for processing different import types: for example - Stock and Price can be 2 different imports but they could be located in same file
4) Source - its a separate module, this mean you can just delete it. For Import its no matter where you received your file from, you can use 3rd party extension, you can create your own module that will place file in correct location and will create required entity in database. and it still will work. The same process works in another direction. You can upload file and then create any custom script that will process it. And in this case you will use all standard Magento security features for file valiation and upload.

I hope it makes sense :-)

design-documents/asynchronous-import/base-extension.md
nuzil: Yep, I would not stick to multiple files in MVP in any case. Especially when we have BULK API on-board. But this can be discussed in a future steps if we will get request from users

design-documents/asynchronous-import/base-extension.md
tariqjawed83: I believe we can still achieve all these points using a streaming based approach to this problem. Allow import again for a file that has been imported, by updating only import settings or format, there is no stopping in that. I have been involved in numerous data migration / ETL projects in the past, and have seen these issues many times.

Maybe we can improve this process later on after MVP, but I just want to make it clear, that this will not be performant for customers in many scenarios, and it will not be using the streaming based approach potential.

But I am okay, if you want to move forward initially with approach and test the waters.

nuzil: Agree!! One more point I forgot to mention is a complexity of such request. We already have an issue that for each file format we need own endpoint and own interfaces cause they have a custom format. Then also add to the body mapping information and body already become more complex and heavy. 
As we are providing API for that, I believe that if somebody will integrate it with own system, he can also build in a way of one request. 

I any case, agree with you that let’s try it how it will work and how the user experience will be. For now we had a good experience with this on our project. Let’s see

tariqjawed83: Sure, I am okay. For large data-sets or files, this sequential implementation never performed at the desired rate for me in the past; but we can always improve on, this is not something fundamentally right or wrong. This is based on needs, streaming based approach end to end, is much more efficient in terms of memory footprint & cpu processing, that all modern systems are using it.

design-documents/asynchronous-import/base-extension.md
```diff
@@ -23,16 +26,16 @@ Path is relative from Magento Root folder
...
-      "import_data": "var/catalog_product.csv",
-      "import_type": "local_path",
-      "uuid": "UUID",
-      "format": {
-          "csv_separator": "string",
-          "csv_enclosure": "string",
-          "csv_delimiter": "string",
-          "multiple_value_separator": "string"
-      }
+  "uuid": "UUID",
```
tariqjawed83: will this UUID be returned by the system/api itself rather than the user? if that is true, then we can remove it from the Request, and also we can name it just an ID, rather than UUID. Since UUID is one of the strategy to generate the IDs, which can be changed in the future. We should not name the underlying strategy of the ID generation, for the name of the Entity ID.

design-documents/asynchronous-import/base-extension.md
```diff
@@ -207,70 +166,88 @@ Will return list of Source that was uploaded before.
...
-                              "function": "string",
-                              "args": "array"
-                        }
-                  ]
-              }
-        ]
-    }
+        "import_images_file_dir": "string"
+    },
+    "convertingRules": [
```
tariqjawed83: 1- Mapping is a commonly used term for Data Mapping, 
2- Not sure, if I can understand how to call this API, it is not clear the intend of the parameters to be honest.
3- What is the use-case of sorting while importing? usually sorting of any kind doesn't happen during data importing.

design-documents/asynchronous-import/base-extension.md
```diff
@@ -41,71 +44,38 @@ Path is relative from Magento Root folder
...
+  "uuid": "UUID",
+  "source_data": {
+    "source_type": "upload_file",
+    "source_data": "c2t1LHN0b3JlX3ZpZXdfY29kZSxhdHRya..." 
+  } 
+  "format": {
+    "delimiter": "string",
+    "enclosure": "string",
+    "escape": "string",
+    "multiple_value_separator": "string"
```
tariqjawed83: what is the purpose of this "multiple_value_separator"?

design-documents/asynchronous-import/base-extension.md
```diff
@@ -279,30 +256,79 @@ Will be returned list of objects that we tried to import
...
+  "errors": [],
+  "created_at": "string",
+  "finished_at": "string",
+  "import": {
+        "uuid": "",
+        "source_uuid": "",
+        "import_type": "catalog_product",
+        "validation_strategy": "string",
+        "allowed_error_count": 0,
+        "import_image_archive": "string",
```
tariqjawed83: are we planning to import images as well? if yes, not sure where they will be landing, it could be some CDN etc.

design-documents/asynchronous-import/base-extension.md
nuzil: UUID can be defined by the client, but at the same time IF its not set, we are generating it on Magento side, thats why this param is here

design-documents/asynchronous-import/base-extension.md
nuzil: For example Multiselect attribute value: 
"color": "Red|Green|White"
In this case - | is a multiple value separator

design-documents/asynchronous-import/base-extension.md
nuzil: Yes we are planning and currently they will be imported in Magento.
Magento our of box is saving all images under pub/media/
Magento API same time accepting images as base64_encoded content.
So form our side are not really care about where are the images saved, we just taking what customer giving to us, and forwarding them to Magento in base64 view. 

Customer providing images by uploading them on Magento filesystem. and they will be deleted after import.

Yes, in perfect way we have to support CDN storage, and for Magento we already have some support. But for our module - its currenly I guess out of scope of MVP.

Interesting point for future, we can groom locating where User can user place images for Import

design-documents/asynchronous-import/base-extension.md
nuzil: Thats fine, its not so easy from first glance :-)
convertingRules are defining how values for import can be changed to archive our goals. 
Same time its not require param.
Out of box, without this param we are supporting standard Magento CSV format.
But imaging real world where each system export their own formats. This mean we have to provide possibility to configure import source so, that customer can import any csv data structure he wants.
Here are coming convertingRules. 

So we have in the code some sub-module: RulesConverter.
Module takes all data we giving him (in this case read from the source file).
Then converter apply rules for this data.
Example: In import Status == Yes/No, we have a rule that convert this to 0/1 .... and so on for all required NOT standard Magento fields. 
And it returns already changed input data, but those data are ready for import to Magento.

Sort order is required, cause you may have multiple manipulation with same field, so order of execution have to be correct.

design-documents/asynchronous-import/base-extension.md
nuzil: I understand your point related to IDs naming.
Its correct idea if we are talking about system that have this pattern everywhere. 
Unfortunatelly its not in Magento, so _id is everywhere in the system are Database Ids. 
Same time you are right that user should not be aware of way how ID is generated. 

In our implementation its explicitly UUID, cause its currently on our code level.
May I ask you that we leave it as it is for now, and will have another discussion.

Btw. you can raise this topic on next Architectural meeting so there will be a lot of people to discuss and this may define pattern for future Magento standard of IDs naming

tariqjawed83: ok I will raise this within architecture meeting

design-documents/asynchronous-import/base-extension.md
tariqjawed83: okay, so it should be multiple_value_delimiter? to match above convention

design-documents/asynchronous-import/base-extension.md
tariqjawed83: thanks for clarification,

design-documents/asynchronous-import/base-extension.md
nuzil: Make sense, aligned

tariqjawed83:
@nuzil I understand there were changes to the design, should we need to update this pull accordingly? or we should use this pull https://github.com/magento/architecture/pull/330/ instead?

nuzil:
This one can be closed. New one will come

PR #271, opened 2019-09-11
Reduce q-ty of error reports created in <magento_root>/var/report

arhiopterecs:
## Problem
Sometimes some clients have problems with this error reporting mechanism.
For example:
In the case when there are many requests to the Http application where there is a critical error, a huge number of reports about this error will accumulate in the directory `<magento_root>/var/report`. Magento does not have a mechanism for cleaning the report directory. System Integrator does not spend the time for monitoring the file reports, in most cases. As a result, report files are accumulated. The number of report files grows to hundreds of thousands, and sometimes it happens that the number of report files reaches a million or more.
With such a number of files in the folder, it becomes more difficult to work with error reports. It is not possible to look at the contents of the directory, and it is even difficult to remove the reports. There are cases that report files to occupy all the free space in the storage, which leads to the crash of the application.

## Solution
The solution to this problem is to save only unique reports `var/report/<unique_hash>` and associate them with the messages in the exception log `var/log/exception.log`. For associations, we may use a unique hash that will be generated using an exception trace, an exception message, $_SERVER['REQUEST_URI'] and $_SERVER['SCRIPT_NAME'].
The hash will be used to name the report file and this hash will be indicated in the report and the log message. And, we can save reports in subdirectories with a configurable nesting level, this avoids the overflow of the <magento_root> var/reports` directory with millions of files

## Requested Reviewers
@buskamuza 
@kandy


larsroettig:
Maybe it will be also good to define different output adapter not only file maybe it nice to put this in elastic search cluster.

design-documents/error-reporting/reduce-q-ty-of-error-reports.md
```diff
@@ -0,0 +1,82 @@
...
+
+***For Example:***
+
+If we have hash sha256('')= 44ffb1087a44e61b018b3cdee72284d017f22e52755c24e5c85cbac1647ae7a7
+
+1. MAGE_REPORT_DIR_NESTING_LEVEL=0 -> <magento_root>/var/report/44ffb1087a44e61b018b3cdee72284d017f22e52755c24e5c85cbac1647ae7a7
+1. MAGE_REPORT_DIR_NESTING_LEVEL=1 -> <magento_root>/var/report/**44**/44ffb1087a44e61b018b3cdee72284d017f22e52755c24e5c85cbac1647ae7a7
+1. MAGE_REPORT_DIR_NESTING_LEVEL=2 -> <magento_root>/var/report/**44**/**ff**/44ffb1087a44e61b018b3cdee72284d017f22e52755c24e5c85cbac1647ae7a7`
+....
+1. MAGE_REPORT_DIR_NESTING_LEVEL=32 -> <magento_root>/var/report/**44**/**ff**/**b1**/**08**/**7a**/**44**/**e6**/**1b**/**01**/**8b**/**3c**/**de**/**e7**/**22**/**84**/**d0**/**17**/**f2**/**2e**/**52**/**75**/**5c**/**24**/**e5**/**c8**/**5c**/**ba**/**c1**/**64**/**7a**/**e7**/**a7**/44ffb1087a44e61b018b3cdee72284d017f22e52755c24e5c85cbac1647ae7a7
```
paliarush: We need to come up with recommended nesting level or some formula for calculating optimal nesting level. It should be published in docs.

PR #277, opened 2019-09-16
Updated Magento Commerce Testing Strategy

nathanjosiah:
## Problem

Magento test coverage is sporadic and somewhat arbitrary and our testing strategy is unconventional, not very effective, and inconsistent/conflicting across 4 sets of testing requirements.

## Solution

Implement a new incremental testing strategy that focuses on layered coverage with each level containing decreasing levels of granularity and quantity of tests.

## Requested Reviewers

@maghamed @buskamuza @psullivan @irenedhong


nathanjosiah:
Adding to this following a recent meeting, in the actual definition of done I will expand the definition for how "Code MUST be tested on its own" as well as more information about tests for bug fixes. I will also add a note that MFTF may be the appropriate test coverage depending on what code was written (such as wiring code that is otherwise difficult or impossible to test via non-functional tests) which would be an exception to the no P0/P1 rule.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+
+<!-- Describe any new terms used in the document -->
+
+### Overview
+
+<!-- Describe what we are trying to solve and a few sentence summary of design approach. -->
+
+Despite running 20,000 - 30,000 tests per pull request, we have little to no confidence in the stability of the product. This means that we still need a manual regression process that lasts weeks in the best of cases. 
+
+A sizable portion of this can be attributed to our current high-level testing mentality that prioritizes system-level black box testing over other types of tests. While these types of tests are necessary and good to have, they only provide implicit coverage to the code paths of the concrete implementations that happen to be loaded in the high-level black box tests (they are also expensive/difficult to maintain). We still need explicit coverage for the concrete implementations that extension developers have hard dependencies on. This is true for the simple code paths as well as variations both positive and negative flows.
```
melnikovi: What do you mean by high level black box tests, functional tests? In general black box tests are good, and they should be easy to maintain compare to white box tests that knows about implementation. Because every change to implementation (including those that don't change functionality) will break the test. Also, I believe current approach is to use appropriate type of test for each case. It might not be well explained, I can agree with that.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Design
+
+#### New Testing Strategy:
+
+The new strategy for testing relies on having test coverage at each layer. The idea is to have many small and fast tests. Therefore, tests MUST be written in the following order of granularity and the higher-level the test, the fewer tests there will be:
+
+1. Code MUST be tested on its own. This MAY be done through unit or integration testing as appropriate. This level of test coverage ensures that the code can properly function on its own.
```
melnikovi: I would suggest to clarify that we want to write black box unit tests. Currently we have a lot of unit tests that are white box tests. They don't test much and break when you make changes (including those that don't affect functionality) because everything is mocked including business logic.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Design
+
+#### New Testing Strategy:
+
+The new strategy for testing relies on having test coverage at each layer. The idea is to have many small and fast tests. Therefore, tests MUST be written in the following order of granularity and the higher-level the test, the fewer tests there will be:
+
+1. Code MUST be tested on its own. This MAY be done through unit or integration testing as appropriate. This level of test coverage ensures that the code can properly function on its own.  
+1. Code MUST be shown to work with known collaborators. This MUST be done through integration tests and ideally SHOULD involve as few components as possible per test case. This level of tests ensures that the code will work with its known associates. Examples of such associates would include aggregators, composites, etc.
```
melnikovi: I would suggest to add more clarity on how using integration test here would be different than in previous item if you think that it's important. If following previous item sometimes would mean following this one as well, than it might be ok.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+1. webapi behavior MUST be covered by blackbox tests to assert that the service contracts are properly satisfied. This ensures that the software package as a whole functions properly when everything is compiled together. This includes REST/SOAP/GraphQl etc. To be clear, these tests do not exclude the underlying concrete implementations from being tested. 
+1. MFTF tests SHOULD only be used to cover P0/P1 testing scenarios. Until these scenarios are all completed, a product owner MUST be consulted for approval of new tests. Additionally, MFTF MUST NOT be used for testing scenarios that are not P0/P1
```
melnikovi: I believe this should be severity, not priority scenarios. Severity defines impact on product, priority is used to specify order in which issues that have the same priority should be fixed.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+
+<!-- Describe what we are trying to solve and a few sentence summary of design approach. -->
+
+Despite running 20,000 - 30,000 tests per pull request, we have little to no confidence in the stability of the product. This means that we still need a manual regression process that lasts weeks in the best of cases. 
+
+A sizable portion of this can be attributed to our current high-level testing mentality that prioritizes system-level black box testing over other types of tests. While these types of tests are necessary and good to have, they only provide implicit coverage to the code paths of the concrete implementations that happen to be loaded in the high-level black box tests (they are also expensive/difficult to maintain). We still need explicit coverage for the concrete implementations that extension developers have hard dependencies on. This is true for the simple code paths as well as variations both positive and negative flows.
+
+### Design
+
+#### New Testing Strategy:
```
melnikovi: I would recommend to run this through developers to see if it's clear. I personally think that it would be good to include some instructions and examples, it will make easier to apply the strategy.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+1. MFTF tests SHOULD only be used to cover P0/P1 testing scenarios. Until these scenarios are all completed, a product owner MUST be consulted for approval of new tests. Additionally, MFTF MUST NOT be used for testing scenarios that are not P0/P1
+
+In addition, regardless of the interpretation of the strategy above: Any class marked with an `@api` annotation MUST be covered with explicit test coverage via integration or unit tests. These tests MUST test the concrete implementation's behavior in a way that can not be inadvertently changed outside of the test itself. For example: testing a concrete class through DI preference, is not safe because it could be overridden via configuration of another loaded module.
+
+#### What to test when changing code
+1. Before changing code or writing new code, write passing tests (either blackbox or whitebox as appropriate) for existing behavior that doesn't already have test coverage.
```
melnikovi: What would be an example of white box test? Integration and functional tests generally considered blackbox tests as far as I know. A unit test? Writing white box unit test will not help much as it will break when you start making changes (including those that don't affect functionality). I suggest to clarify that we want to write black box tests in general and white box tests only to test SPI. And instead of black and white box tests here I would just name different types of tests unit, integration, API and functional.

design-documents/testing-strategy.md
```diff
@@ -0,0 +1,39 @@
...
+1. webapi behavior MUST be covered by blackbox tests to assert that the service contracts are properly satisfied. This ensures that the software package as a whole functions properly when everything is compiled together. This includes REST/SOAP/GraphQl etc. To be clear, these tests do not exclude the underlying concrete implementations from being tested. 
+1. MFTF tests SHOULD only be used to cover P0/P1 testing scenarios. Until these scenarios are all completed, a product owner MUST be consulted for approval of new tests. Additionally, MFTF MUST NOT be used for testing scenarios that are not P0/P1
+
+In addition, regardless of the interpretation of the strategy above: Any class marked with an `@api` annotation MUST be covered with explicit test coverage via integration or unit tests. These tests MUST test the concrete implementation's behavior in a way that can not be inadvertently changed outside of the test itself. For example: testing a concrete class through DI preference, is not safe because it could be overridden via configuration of another loaded module.
```
melnikovi: Most if not all of the classes with @api annotation will not be beneficial to cover with unit tests. There are some classes marked with @api annotation which are good candidates for unit tests, but we have handful of them. I think it's very important to distinguish when to use appropriate type of test.

Classes may not necessarily have @api annotation, for example `\Magento\Catalog\Model\CategoryRepository`. We may want to say that all implementations of the interfaces marked with @api annotation should be covered instead. But some of these interfaces covered with other types of tests, for instance service contracts are covered with API tests. Need to make sure everyone will be ok with that.

design-documents/testing-strategy.md
nathanjosiah: I agree, we are going to emphasize that unit tests are going to be fairly rare in Magento code. Essentially code tests will default to integration tests unless there is an obvious use case for a unit test.

design-documents/testing-strategy.md
nathanjosiah: I'm not sure I understand this question. The idea here is the verifying that code works in the ways that it could be used. First it could be used on its own (such as a di injected hard dependency) so we need to make sure it behaves correctly. Secondly, it could be used with something else (such as with a pool, aggregator, composite, renderer, chain, etc) and it may not interact well with other components in this way. This is the more traditional idea of an integration test.

design-documents/testing-strategy.md
nathanjosiah: > I believe this should use severity, not priority. 
and 
> coverage for S2 scenarios might be important for regression.

I will defer these to @irenedhong for this. She defined this policy. However I will comment that this is something that I'm personally conflicted with which direction we should go because while I understand Irene's policy I also see that we have separate policy for how we handle P2/S2 bugs compared to 3's and 4's and I'm not sure it's clear why we would have a separate policy for writing the tests versus fixing the bugs.

> One of the benefits of functional tests is that they are declarative and can be extended by extension developers. What is our strategy regarding other types of tests that are not declarative?

This is something that has been discussed in depth. It's mentioned in the proposal but this is only a theoretical problem currently. We understand this could potentially be something that needs to be addressed in the future but we have been unable to find any examples (even hypothetical) of when this was actually a problem and so we can't try to create some solution for it. Also, should this ever become a problem, we aren't eliminating functional tests so they could be used as a solution while some extensibility is built to accommodate these issues.

design-documents/testing-strategy.md
nathanjosiah: Yes, I agree. This proposal is not intended to be a final document. Before this PR was created the discussions surrounding this new strategy have been abstract, high-level, and theoretical. This PR was the first time it has been written down in some explicit form. There will be further PR's to come in the future with more details based on these discussions.

design-documents/testing-strategy.md
nathanjosiah: Again, referring to my previous comments above, there are times when black box testing doesn't make any sense. For example, if you have a method signature like `filter($value):string` the contract is pointless to test totally blackbox in the sense that you can't simply verify basic input=>output. You should definitely do that but it isn't enough in most cases. The implementation will dictate how complicated the testing scenarios will be. 

What seems to be the root issue here is there is a Magento-specific terminology for what "whitebox test" means. For Magento this seems to be defined as test that is a direct mirror of the code. But that isn't true, a whitebox test would be one where it has some level of knowledge about the implementation. Such cases are good for making sure the implementation handles corner cases (unique to the implementation) correctly.

Again, the idea is that contracts are great but there are still concrete dependencies on implementations and implementation can have individualized flaws/weaknesses specific to the details of that implementation.

design-documents/testing-strategy.md
nathanjosiah: I think you misread this section. 

> coverage via integration or unit tests

This says it must be covered with some type of non-functional test. This is not a final document but wouldn't contain specific details for when to use which type of test anyway. This document and the subsequent DoD will contain the testing strategy. The implementation of that strategy will depend on the code being written. We plan on making updates to the testing documentation to make it more clear when to use different types of tests.

design-documents/testing-strategy.md
irenedhong: @maghamed recommended that we use Priority over Severity because something can be low priority overall but high severity for a specific client. I leave this more up to those who have more knowledge between the differences and I have no opinion on this whether we choose to say it is P0/P1 or S0/S1.

For the distinction that we do this arbitrary cut off at P2 for UI tests was a recommendation by @maghamed that I agreed with. Please remember that we will continue to test P2 and are actually investing quite a lot of effort to shore these variations/edge cases/negative cases/permutations in our code base for the most critical modules. The note here is that we want to cover these with cheaper and faster non GUI tests as there can be hundreds of permutations/edge cases/negative/variations. We want to save the use of highly valuable but very expensive GUI tests for the most important P0/P1 scenarios where we are willing to pay the price of maintenance, creation, and run time. As part of this effort, we are spelling out what the most critical P0/P1 scenarios even are. Also we allow for exceptions to the rule by asking the product owner if we think it is critical to have a P2 covered by a GUI test. But then that begs to ask, if it is so critical why is it not a P0/P1?

For our bug fixing policy, I believe this document aligns with that. We hotfix for any P0 scenarios and we will stop a release for P1 scenarios. While we do fix P2's, we don't fix all P2's because the business value is not always there as they are often variations/permutations/edge cases/negative cases.

@melnikovi can you please clarify what you mean 'coverage for S2/P2 scenarios might be important for regression'. Which scenarios? Would you say that automated coverage for S2/P2 scenarios are not enough with the effort underway? Who would define all these P2 scenarios that need to be covered in regression specifically?

joanhe: I don't think it is good idea to restrict adding tests. We want to increase test coverage. If developer and/or QA determine function test is good type of test to ensure coverage of product change. We shouldn't prevent the test to be added by extra approval process.

nathanjosiah: This goes against one of the main goals of this initiative. Adding additional more expensive coverage for unnecessary flows creates more of the problem we are trying to solve. @irenedhong

fooman:
One thing that I am not sure if its currently done or not but I always wonder if a lot of the existing tests are somewhat disconnected from what the end users end up using. A couple of examples of this:

1. the test suite executes against the codebase on github - however most use Magento as retrieved from repo.magento.com. The repo.magento.com version differs in a few substantial ways (extra core bundled extensions and modules are packaged). Would love to see the following added:
Install Magento via 
```
composer create-project --repository=https://repo.magento.com/ magento/project-community-edition <install-directory-name>
bin/magento setup:install [...]
execute integration tests
execute mftf tests
```
2. most users do not run a new install but instead upgrade from prior versions - ideally we would match this, ie
```
composer create-project --repository=https://repo.magento.com/ magento/project-community-edition <install-directory-name>
bin/magento setup:install [...]
composer require magento/product-community-edition:2.3.3 --no-update
composer update
execute mftf tests
```
3. Magento ships with `.htaccess` files as well as a suggested `nginx.conf` are these covered? Same for docroot in the pub/ folder or not?

A few other areas where I think tests are light: Multi-store, Non-English, Emails.

nathanjosiah:
@fooman These are great examples of why we need this new testing strategy. 

1. yes we are fully aware of this. It's actually much worse than even that and in fact we run 0 builds on github packages for releases but it's also more complicated than it seems. There is another initiative in the works to address this.

2. we do actually have some upgrade tests but they are focused on the DB and semantic versioning  so I would be interested in hearing what we could do to improve coverage for these use cases.

3. This is actually kind of complicated as well and I would love to see more coverage specifically related to environment code but we currently don't have test coverage because we don't have build environments to support these types of tests. Our build environments use nginx.

nathanjosiah:
The first concrete iteration of the new documentation can be found here https://github.com/magento/devdocs/pull/5534 and needs more comments before it get's merged. It's approved from a directional standpoint but it would be good to have some feedback on clarity and missing points.

fooman:
For 2. what I mentioned above would already be a good base line. It would tease out issues with the upgrade procedure (ie if the above steps are not enough and/or it's correctly picked up by https://devdocs.magento.com/guides/v2.3/comp-mgr/cli/upgrade-with-plugin.html)

For 3. maybe a first good step would be to copy this file https://github.com/magento/magento2/blob/2.3-develop/nginx.conf.sample during the build.

nathanjosiah:
@fooman 2. I will make sure this is investigated further but we may have something already in place for this. Though, it's not comprehensive for sure if we do.
3. It's not quite this simple. The build environments have tons of custom configuration so copying the sample configuration doesn't really make sense. Those files are meant to be used as a sample which is why they are named as such. However, the broader point about automated environment tests is still very true and a definite weak point.

PR #282, opened 2019-09-18
GraphQL custom attributes container proposal

paliarush:
GraphQL custom attributes containers proposal

### DIFF
--- a/design-documents/graph-ql/custom-attributes-container.md
+++ b/design-documents/graph-ql/custom-attributes-container.md
@@ -1,0 +1,31 @@
+## Problem statement
+
+Some Magento installations may have tens of thousands of custom (EAV) product attributes. This is possible when the merchant is selling thousands of different product types with unique attribute sets.
+
+Custom attributes in GraphQL schema are presented in flat structure, which may be a problem for the client application when performing search in catalog. GraphQL requires all needed fields to be explicitly listed in the query, "get all fields" queries are not supported. At the same time Magento is expected to return all EAV attributes applicable to each product. 
+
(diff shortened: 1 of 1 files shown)


AlexMaxHorkun:
Different products having different set of properties (custom attributes) sounds a lot like products having different interfaces and GraphQL has tools for clients to request certain fields based on data type/interface. I think that we could generate interfaces based on Attribute Sets so that clients could do something like this:
```
products (filter: $filter) {
    items {
        sku
        ... on AttrSetDefaultProductInterface {
            color
        }
        ... on AttrSetCustomSimpleProduct {
            custom_attr
        }
    }
```
That would be a more elegant and GraphQL compliant solution then duplicating bunch of fields as custom_attributes list. @paliarush what do you think?

paliarush:
@AlexMaxHorkun unfortunately multiple interfaces do not solve the issue described in the proposal: the client will still have to list all possible fields, even worse, now the client will have to list all known attribute sets as well. In addition, knowledge about attribute sets is admin-only.

AlexMaxHorkun:
> @AlexMaxHorkun unfortunately multiple interfaces do not solve the issue described in the proposal: the client will still have to list all possible fields, even worse, now the client will have to list all known attribute sets as well. In addition, knowledge about attribute sets is admin-only.

Clients are probably aware of all the fields - they'd have to display them after all. I still think products of different sets can be treated as basically different product classes - they posses different properties and most likes differ in presentation.

AlexMaxHorkun:
@paliarush Have you considered introducing a new Query with custom attributes and their metadata for clients to utilize when querying products? Clients might want to display attributes differently based on their type and frontend label etc. I think that would solve this issue. Perhaps such query already exists, I haven't checked

paliarush:
@AlexMaxHorkun There is metadata query and there is a plan to add attribute metadata to schema.
I think we are not on the same page regarding the issue being solved here. 
Example: client has to send query with 10000 attributes during product search. In the response it gets products with 100 different attribute values and 9900 nulls per product. This is an issue and I don't see how it is solved with your proposal.

PR #286, opened 2019-09-22
Add default schema for a Processing Rules

nuzil:
## Problem

Default Magento CSV formal contains data, which are cannot be directly imported into Magento.

For processing different types of Import, we have to define default rules processing, for a case when User want to import CSV data in default Magento Format.

In additional, we want to provide possibility to developers to define own default rules sequence for minimize requests body and define own rules.

## Requested Reviewers
@tariqjawed83 @naydav @paliarush


nuzil:
Hi @tariqjawed83 
I will close it for now, we groomed a little bit and will align content before providing new PR

PR #291, opened 2019-09-24
GraphQL cart error messages

lenaorobei:
## Problem
Now in GraphQL when multiple errors occur with cart there is no way to read all cart messages in order to understand the whole context.  
53808295-8c011b00-3f5a-11e9-9372-d84fb4464e89

## Solution
@XxXgeoXxX proposes to add additional filed to the `Cart` type https://github.com/magento/graphql-ce/pull/475
Names changed by me but, PR will be adjusted once approved here.

## Requested Reviewers
@paliarush

### DIFF
--- a/design-documents/graph-ql/coverage/Cart.graphqls
+++ b/design-documents/graph-ql/coverage/Cart.graphqls
@@ -27,6 +27,8 @@ type Cart {
 
     is_multishipping: Boolean!
     is_virtual: Boolean!
+
+    errors: [CartError]
 }
 
 type CheckoutCustomer {
@@ -56,4 +58,9 @@ type CheckoutPaymentMethod {
 
 type CartGiftCard {
     code: String!
+}
+
+type CartError {
+    identifier: String
+    text: String
 }


design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -57,3 +59,8 @@ type CheckoutPaymentMethod {
 type CartGiftCard {
     code: String!
 }
+
+type CartError {
+    identifier: String
```
paliarush: Please provide a use case when `identifier` is needed on storefront.

It seems like it should not be exposed to storefront. If so than this `CartError` type is excessive and its usages can be replaced with `String`

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -57,3 +59,8 @@ type CheckoutPaymentMethod {
 type CartGiftCard {
     code: String!
 }
+
+type CartError {
```
paliarush: Please explain in proposal why it is not possible to use built-in mechanism for GraphQL errors along with `\Magento\Framework\Validation\ValidationException::getErrors`

paliarush: Please explain in the proposal why we should ignore other message types like `success`, `notice`, `warning`

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -57,3 +59,8 @@ type CheckoutPaymentMethod {
...
+    identifier: String
+    text: String
```
paliarush:
```suggestion
    text: String!
```

design-documents/graph-ql/coverage/Cart.graphqls
paliarush:
```suggestion
    identifier: String!
```

design-documents/graph-ql/coverage/Cart.graphqls
XxXgeoXxX: The main reason that resolvers have conditions to prevent the execution of add product to cart and show items in the cart(example app/code/Magento/QuoteGraphQl/Model/Cart/AddProductsToCart.php). That behavior difference with frontend where the customer could see products in cart with messages on frontend but in graphql query he got the error.

XxXgeoXxX: In base PR https://github.com/magento/graphql-ce/pull/475 we expected all kind of messages. Looks like we need rename CartError to CartMessages.

design-documents/graph-ql/coverage/Cart.graphqls
XxXgeoXxX: The main reason to add the identifier is localizing part of process where errors were thrown. It should fetch profit for developers who extend base functionality and implements new features.

lenaorobei:
@paliarush @XxXgeoXxX we have an alternative proposal here https://github.com/magento/graphql-ce/issues/970

It will look like this:
```
{
  "errors": [
    {
      "message": "First error.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    },
    {
      "message": "Second error.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```
Please let me know what do you think.

lenaorobei:
Closing in favor of https://github.com/magento/graphql-ce/issues/970

paliarush:
@lenaorobei it looks to me that there is a bug in the native error handling functionality which should be fixed. We should not introduce custom error messages.

PR #296, opened 2019-09-26
PHPCompatibility

buskamuza:
## Problem

Add PHPCompatibility to code style ruleset.

### DIFF
--- a/design-documents/testing/coding-standards/phpcompatibility.md
+++ b/design-documents/testing/coding-standards/phpcompatibility.md
@@ -1,0 +1,28 @@
+# PHPCompatibility in Code Style Ruleset
+
+## Goal
+
+Utilize PHPCompatibility tool in `phpcs` rules, so that rules specific to PHP version are validated.
+
+Requirements include determining PHP version from `composer.json` of the Magento project or Magento module, so validation against correct version happens automatically.
+This requirement though introduces additional complexity to the ruleset/sniff implementation:
+- `composer.json` may be absent
+- it requires custom bootstrapping logic for `phpcs`
+- it's not clear whether we should analyze different files differently in case they are from different modules and those modules hav different PHP constraints
+
+Based on the above it is proposed to put the responsibility of specifying PHP version to the user (owner of the project).
+
+## Implementation Details
+
+In https://github.com/magento/magento-coding-standard repository:
+1. Add `phpcompatibility/php-compatibility` to Composer dependencies
+2. Add necessary configuration to the ruleset, so that PHPCompatibility ruleset is included
+   2.1. Add `<config name="installed_paths" value="./../../../../../vendor/phpcompatibility/php-compatibility" />`
+   2.2. Add `<rule ref="PHPCompatibility"/>`
+
+There are two options for https://github.com/magento/magento2 repository:
+1. Add `<config name="testVersion" value="7.1.3-7.3.0"/>` to Magento static tests ruleset
+   1.1. Different branches will have different PHP versions
+   1.2. Drawback: PHP versions in the ruleset should be updated every time Magento changes supported PHP versions
+2. Determine PHP version in `\Magento\Test\Php\LiveCodeTest::testCodeStyle()` based on the project's `composer.json` and pass it as an argument to `phpcs`
+   2.1. Drawback: IDEs using Magento ruleset won't take into account any PHP version. This should not be a big issue as usually IDEs have settings for runtime version and highlight incompatibilities with selected version. If this is not enough, a project owner can add `<config name="testVersion" value="their-php-version"/>` in their ruleset
--- a/design-documents/testing/coding-standards/phpcompatibility.md
+++ b/design-documents/testing/coding-standards/phpcompatibility.md
@@ -2,7 +2,7 @@
 
 ## Goal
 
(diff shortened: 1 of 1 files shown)


design-documents/testing/coding-standards/phpcompatibility.md
```diff
@@ -0,0 +1,28 @@
...
+1. Add `phpcompatibility/php-compatibility` to Composer dependencies
+2. Add necessary configuration to the ruleset, so that PHPCompatibility ruleset is included
+   2.1. Add `<config name="installed_paths" value="./../../../../../vendor/phpcompatibility/php-compatibility" />`
+   2.2. Add `<rule ref="PHPCompatibility"/>`
+
+There are two options for https://github.com/magento/magento2 repository:
+1. Add `<config name="testVersion" value="7.1.3-7.3.0"/>` to Magento static tests ruleset
+   1.1. Different branches will have different PHP versions
+   1.2. Drawback: PHP versions in the ruleset should be updated every time Magento changes supported PHP versions
+2. Determine PHP version in `\Magento\Test\Php\LiveCodeTest::testCodeStyle()` based on the project's `composer.json` and pass it as an argument to `phpcs`
```
jean-bernard-valentaten: I'd argue that using the namespace `\Magento\Test` is not such a good idea as it may be confused with the namespace `\Magento\Tests` in which the automatic tests are defined.
The same argument stands valid for the class name `LiveCodeTest` and the method name `testCodeStyle` as they are too close to the naming convention for test classes and cases.  

If I understand you right, you'd propose to use this class as a kickstarter  for `phpcs`. This would indeed have the drawback that IDEs would not take into account any PHP version and it also has the drawback, that poeple need to call another command than `phpcs` to have PHP versions taken into account.  
In order to eliminate those drawbacks, I'd suggest to use a config [bootstrap](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-bootstrap-file) to execute class in the ruleset. This can be specified as an argument to `phpcs`and can thus be contained in the `ruleset.xml`:
```xml
  <arg name="bootstrap" value="php-compatibility-bootstrap.php" />
```

In turn I'd suggest to rename `\Magento\Test\Php\LiveCodeTest::testCodeStyle()` to `\Magento\Bootstrap\PhpCompatibility::execute()` and have the bootstrap manipulate the configuration settings.

The file `php-compatibility-bootstrap.php` would then contain the following code:
```php
<?php
/**
 * Manipulate config settings for package 'phpcompatibility/php-compatibility'
 *
 * @var \PHP_CodeSniffer\Runner $this
 */
use Magento\Bootstrap\PHPCompatibility;

$bootstrap = new PHPCompatibility($this);
$bootstrap->execute();
```

The variable `$this` should be available in the context of the bootstrap file as the file is being `include`ed in `\PHP_CodeSniffer\Runner::run()`.

buskamuza: The class I mentioned already exists and it's used for starting `phpcs` in our builds.
My idea is:
- automate PHP version detection for our builds
- keep implementation as simple as possible for 3rd-party. The idea here is that we just include in our docs something like "Now we have PHPCompatibility rules included, please go read about it in its official documentation". Then people can follow PHPCompatibility docs and either specify or not specify PHP version. I'd really like to avoid overcomplicating the implementation and especially the case when it behaves non-standard or does not allow to configure it in standard way (e.g., due to a bug or incorrect assumption, it becomes impossible to reconfigure PHP version and it will be always auto-calculated).

jean-bernard-valentaten: Ok, I wasn't aware that the class already exists nor could I find it in the repo `magento-coding-standard`. Obviousel I searched in the wrong place. Sorry for the inconvenience.

ihor-sviziev:
@buskamuza,
I don't think it's a good idea to mix coding standards with php compatibility checker. I'd suggest to add one more check to static tests:
https://github.com/magento/magento2/pull/24773/checks?check_run_id=239698431
Capture 2019-09-30 at 8 57 28
+ add instructions how to run these tests to devdocs

buskamuza:
@ihor-sviziev , why do you think it's bad? PHPCompatibility itself has guidelines about how to use it as part of phpcs.
We can add it as a separate static test, but then it will be only executed if somebody knows about it.

lenaorobei:
I would like to ask someone from EQPS team to review this PR as well, since it will have impact on Marketplace verification flow.

@igorin @ravmenon

ihor-sviziev:
> We can add it as a separate static test, but then it will be only executed if somebody knows about it.


As I understood - you want to PHPCompatibility checks to Magento Coding Standard static tests.

Current version contains only typical issues in code. 

If php compatibility check will be added - we can't use these tests.

**Use case:**
We're using it for internal modules.

Now our modules runs only on php 7.2 and we're not support php 7.1 and not implemented php 7.3 support until it become important for us.

buskamuza:
@ihor-sviziev , you're right. If no PHP version is specified, PHPCompatibility reports all issues as if you use latest PHP they support (looks like it's PHP 7.4 now).

Then we could add it on the level of Magento repo (each branch could specify its own PHP version).
Or we could implement it as a separate test with PHPCompatibility ruleset only. It would be included in `require-dev` of Magento, but will not impact base Magento ruleset.

ihor-sviziev:
@buskamuza thank you!

PR #299, opened 2019-09-27
GraphQL Directives for EAV attribute metadata

paliarush:
GraphQL Directives for EAV attribute metadata

### DIFF
--- a/design-documents/graph-ql/framework/attributes-metadata.md
+++ b/design-documents/graph-ql/framework/attributes-metadata.md
@@ -1,4 +1,4 @@
-## Problem statement
+# Problem statement
 
 Storefront applications built on top of Magento GraphQL web APIs need to support the following scenarios:
  - Render advanced catalog search form, which can be configured by the merchant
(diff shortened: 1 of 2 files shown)


nrkapoor:
Looks good.

design-documents/graph-ql/framework/attributes-metadata.md
```diff
@@ -0,0 +1,58 @@
...
+ - Display only a subset of all available product attributes in catalog listing based on the configuration
+ - Render product compare page, which should only display attributes specified by the merchant
+ 
+All these scenarios require specific EAV attribute metadata to be accessible from the storefront application.
+
+Additionally, there should be a way to retrieve metadata for all storefront custom attributes in the system. This information should be cached on the server and on the client. 
+
+# Proposed solution
+
+Relaxing signature of existing `customAttributeMetadata` query by making its `attriubtes` argument optional will allow to fetch all storefront attributes metadata.
```
buskamuza: Could you please add current representation of `customAttributeMetadata` to the document?

design-documents/graph-ql/framework/attributes-metadata.md
```diff
@@ -0,0 +1,58 @@
...
+Additional fields should be added to the metadata response (`Attribute`  type), for example `is_dynamic`, `use_in_compare_products`, `display_in_product_listing`, `use_in_advanced_search`, `advanced_search_input_type`. The exact list of fields must be discussed and approved separately.
+
+Introduction of the following query will allow fetching lists of attributes applicable to specific pages:
+```graphql
+pageSpecificCustomAttributes(
```
buskamuza: Can we use filter instead? So filter will be based on metadata itself. E.g.:
```
customAttributeMetadata(filter: {is_dynamic: {eq: true}, use_in_compare_products: {eq: true}}) { 
    attribute_code,
    attribute_options { ... }
}
```

This would provide flexibility and FE developers can build queries optimal for their scenarios.

paliarush: I see flexibility as a benefit.

On the other hand this query will be more complex to use and to implement. This query still needs to support `attributes: [AttributeInput!]` argument. And these arguments could be combined.

I would say we should decide based on the need of flexibility here. Looks like we have very specific scenarios to support and flexibility is not really important. 
@buskamuza @nrkapoor what do you think?

design-documents/graph-ql/framework/attributes-metadata.md
paliarush: Done

PR #301, opened 2019-09-30
MC-19826: Adds investigation on removal of dependency on Auto_Increment

jean-bernard-valentaten:
## Problem

When trying to upgrade Magento to MySQL v8 we noticed that `AUTO_INCREMENT` is being relied on in different places. Since no decision could be taken directly, we were asked to investigate and report on this subject.

## Solution

No solution is provided by this PR, just a document that can help taking the decision.

## Requested Reviewers

@buskamuza

### DIFF
--- a/design-documents/Investigation-on-removal-of-dependency-on-Auto_Increment-Id-in-application.md
+++ b/design-documents/Investigation-on-removal-of-dependency-on-Auto_Increment-Id-in-application.md
@@ -1,0 +1,443 @@
+## Investigate removing dependency on autoincrement id in application
+
+### Overview
+
+When trying to upgrade Magento to MySQL v8 an issue was encountered that relates to relying on the `AUTO_INCREMENT` of
+a table in several places. This document investigates on where this dependency is being used, why it is used and how
+difficult it will be to get rid of this dependency.
+Note that this document does not discuss the whether or not the dependency shall be removed but rather helps in making
+the decision.
+
+### List of places where `AUTO_INCREMENT` is used and for which purpose
+
+Currently the `AUTO_INCREMENT` is read from database in the following two methods:
+* `\Magento\ImportExport\Model\ResourceModel\Helper::getNextAutoincrement()`: Attempts to extract the `AUTO_INCREMENT`
+    from table status and returns it so that it can be used for predictive purposes.
+* `\Magento\Framework\Mview\View\Changelog::getVersion()`: Attempts to extract the `AUTO_INCREMENT` from table status,
+    decreases it by 1 and returns that. Actually returns the maximal version number that is currently in use.
+    This can probably be replaced by a query of the likes of `SELECT MAX(id) FROM table`.
+
+Places that make use of `\Magento\ImportExport\Model\ResourceModel\Helper::getNextAutoincrement()` were identified as:
+
+* CE
+    * `\Magento\CatalogImportExport\Model\Import\Product::_saveLinks()`:
+        Uses the result to predict the id of the next product link that is being imported.
+    * `\Magento\CatalogImportExport\Model\Import\Product\Option::_importData()`:
+        Uses the result to predict the id of the next option and its value(s) being imported.
+    * `\Magento\ConfigurableImportExport\Model\Import\Product\Type\Configurable::_getNextAttrId()`:
+        Uses the result to collect the labels of of the next super attribute of a configurable that is being imported.
+        Since the `AUTO_INCREMENT` is only fetched once (in case it has not been fetched before) and is then incremented
+        with each further call, this could lead to serious off-by-N issues.
+    * `\Magento\CustomerImportExport\Model\Import\Address::_getNextEntityId()`:
+        Only called in case no address id is specified for the address that is being imported. The result is then used
+        to predict the id of the address that is being imported.
+        The `AUTO_INCREMENT` is only fetched one, in case it has not been fetched before. Every further call returns the
+        cached value and increments it (post increment) for the next call. This could lead to off-by-N issues.
+    * `\Magento\CustomerImportExport\Model\Import\Customer::_getNextEntityId()`
+        Only called in case no customer id is specified for the customer that is being imported. The result is then used
+        to predict the id of the customer that is being imported.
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/architecture?pullRequest=301) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/architecture?pullRequest=301) before we can accept your contribution.<br/>

IvanChepurnyi:
Relying on table metadata to retrieve the next auto-increment can lead to ID collisions. It is better to generate a new auto-increment via insert statement and then rollback the insert after the operation. In such a way, MySQL automatically increments the next id, preventing any race conditions of duplicate ids generated. 

Here is an example of how to achieve id allocation in batch, that I've used for my custom import implementation:

```mysql
START TRANSACTION;
 
INSERT INTO catalog_product_entity (sku)
    VALUES 
           ('sku1'), 
           ('sku2'), 
           ('sku3'), 
           ('sku4');
 
SELECT entity_id, sku
    FROM catalog_product_entity
    WHERE 
        sku IN ('sku1', 'sku2', 'sku3', 'sku4');
 
ROLLBACK;
```

If you are interested in rewriting import, here is something I am working on right now:
https://gitpitch.com/IvanChepurnyi/mm2019nyc-slides

jean-bernard-valentaten:
@IvanChepurnyi I agree, using `AUTO_INCREMENT` was a bad decision, which is why this document was written.  
What you propose is another option, though I personally am not so fond of building a stash of reserved primary keys. If your process crashes at some point before actually persisting the data, the ids are burnt since `ROLLBACK` has no effect on `AUTO_INCREMENT`. Also this technique adds unneccessary queries to the process which in turn has a negative impact on performance when importing big sets of data.

IvanChepurnyi:
@jean-bernard-valentaten, the point is that those ids must be reserved right before inserting new data, not in advance. Also, there is no performance overhead, quite the opposite as you can perform multi-statement inserts instead of insert for each new record. That's why I linked my slides from the speech in New York to explain how 1m SKUs with 20 websites were imported under 10 minutes, that would take 9 hours with standard import/export in Magento 2.

design-documents/Investigation-on-removal-of-dependency-on-Auto_Increment-Id-in-application.md
```diff
@@ -0,0 +1,443 @@
...
+the decision.
+
+### List of places where `AUTO_INCREMENT` is used and for which purpose
+
+Currently the `AUTO_INCREMENT` is read from database in the following two methods:
+* `\Magento\ImportExport\Model\ResourceModel\Helper::getNextAutoincrement()`: Attempts to extract the `AUTO_INCREMENT`
+    from table status and returns it so that it can be used for predictive purposes.
+* `\Magento\Framework\Mview\View\Changelog::getVersion()`: Attempts to extract the `AUTO_INCREMENT` from table status,
+    decreases it by 1 and returns that. Actually returns the maximal version number that is currently in use.
+    This can probably be replaced by a query of the likes of `SELECT MAX(id) FROM table`.
```
akaplya: this is not an option, because this code is not thread-safe.

larsroettig: @akaplya what is about LAST_INSERT_ID(); https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_last-insert-id  ?

jean-bernard-valentaten: @akaplya Can you elaborate what you mean by "not thread safe". Could a situation occur where two processes try to generate a new materialized view for the same entity? Why is this not thread safe?

akaplya: @jean-bernard-valentaten last_increment_id unlike max(id)  will return last generated ID no matter to transaction and it's status.
So it is fine to use last_increment_id, btw you can find an example in SalesSequence module. but never rely on transactional data to determine the next sequential ID.

akaplya: For customers import, it is possible to utilise sequences, as we have for EE and Content Staging or Sales documents numbers.
https://www.percona.com/blog/2008/04/02/stored-function-to-generate-sequences/
In this case, instead of guessing a new ID import script may generate and a unique ID for each new customer or address.

To implement this solution in a consistent and compatible (with models) manner, you can add proper identifier handling into corresponding resource models. Logic should be pretty straightforward. If the resource model has to save an object with no ID and we cannot match this object with an existing record (by the unique key) in the database. For such case, the resource model can request a new ID for the object from a sequence.
See example from Sales module https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/Model/ResourceModel/EntityAbstract.php#L136

Such a solution will be backward compatible, except the cases when somebody tries to insert customer data directly through the INSERT statement.

jean-bernard-valentaten: @akaplya As far as I understand from the MySQL documentation `last_insert_id` is kept per connection only, thus if you have multiple connections, they will get different values when selecting `last_insert_id`. Also this is not supported by all RDBMS, so it would become an issue in case other RDBMS become a thing. I don't see how this is a reliable alternative.

akaplya: This is what we need, we need to generate a value that cannot be repeated by another session/process.
Call to a function last_insert_id, is implementation details. Which could be hidden by fancy sequence interface  Please see the example https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/SalesSequence/Model/Sequence.php#L83

jean-bernard-valentaten: Agreed, I changed it.

PR #275, opened 2019-09-12
Add GraphQl schema for receiving Pickup Locations.

swnsma:
GraphQl Schema for receiving Pickup Locations (in the scope of In-Store Pickup implementation).

### Proposed Schema
```
type Query {
    pickupLocations (
        distance: DistanceFilterInput,
        filter: PickupLocationFilterInput,
        sort: PickupLocationSortInput,
        pageSize: Int = 20,
        currentPage: Int = 1
    ): PickupLocations
}
```
Proposed Schema is to be similar with `Catalog.Products` GraphQl Schema and to `GetPickupLocations` webapi endpoint.
List of arguments:
- distance - special filter of Pickup Locations by distance to the entered address;
- filter - filter by Pickup Location entity attributes;
- sort - sorting of the return collection;
- page_size - the size of the collection page;
- current_page - number of the page which will be returned;

```
input DistanceFilterInput {
    radius: Int!
    country_code: String!
    region: String
    city: String
    postcode: String
}
```
Input type for distance filter:
- radius - radius for the distance search
- country_code - country for the search
- region
- city
- postcode

```
input PickupLocationFilterInput {
    name: FilterTypeInput
    pickup_location_code: FilterTypeInput
    country_id: FilterTypeInput
    postcode: FilterTypeInput
    region: FilterTypeInput
    region_id: FilterTypeInput
    city: FilterTypeInput
    street: FilterTypeInput
}
```
Input type for set of attribute filters.
Uses `FilterTypeInput` from the app/code/Magento/GraphQl/etc/schema.graphqls
```
input PickupLocationSortInput {
    name: SortEnum
    pickup_location_code: SortEnum
    distance: SortEnum
    country_id: SortEnum
    region: SortEnum
    region_id: SortEnum
    city: SortEnum
    street: SortEnum
    postcode: SortEnum
    longitude: SortEnum
    latitude: SortEnum
    email: SortEnum
    fax: SortEnum
    phone: SortEnum
    contact_name: SortEnum
    description: SortEnum
}
```
Input for sort orders. Uses `SortEnum` from app/code/Magento/GraphQl/etc/schema.graphqls

```
type PickupLocations {
    items: [PickupLocation]
    page_info: SearchResultPageInfo
    total_count: Int
}
```
Top level object, result of the Pickup Locations Search.

```
type PickupLocation {
    pickup_location_code: String
    name: String
    email: String
    fax: String
    contact_name: String
    description: String
    latitude: Float
    longitude: Float
    country_id: String
[...]

PR #304, opened 2019-10-01
Schema for store pickup GraphQL API

paliarush:
Schema for store pickup GraphQL API.

GraphQL schema proposed in this PR is approved by PO (@mbrinton01) and current proposal should replace https://github.com/magento/architecture/pull/275 if approved.

### DIFF
--- a/design-documents/graph-ql/coverage/store-pickup.graphqls
+++ b/design-documents/graph-ql/coverage/store-pickup.graphqls
@@ -30,7 +30,8 @@ type PickupLocations {
 }
 
 type PickupLocation {
-    name: String!
+    name: String! # In the admin is called Frontend Name
+    description: String! # In the admin is called Frontend Description
     country: String!
     region: String!
     city: String!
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/store-pickup.graphqls
```diff
@@ -0,0 +1,29 @@
...
+    pickupLocations (
+        filter: PickupLocationFilterInput,
+        pageSize: Int = 20,
+        currentPage: Int = 1
+    ): PickupLocations
+}
+
+input PickupLocationFilterInput {
+    # This type is added for extensibility
+    search_term: String!
```
swnsma: For me it is not clear how `search_term` could work with current API implementation.
Do I correctly understand that:
- Search Term - search string to search by customer address
- we delegate all logic of parsing entered address to server-side php scripts, instead of client-side javascript
- radius for the search will be taken from configurations and there is no way to pass it without customizations

swnsma: The questions for now:
- it is not clear how correctly parse `search_term` into address. Offline address search does not support 'just a string'
- if `search_term` could include search by attribute values (like name of Pickup Location (PL) or some address fields) it is not clear how to extract this knowledge from search string
- probably, fulltext index could be as solution for PL attributes, but it will not solve issue with Offline search.

paliarush: Thank you, will add `radius` and ability to query radius units, google/offline search mode.

PR #264, opened 2019-09-06
Declarative schema database compatibility

melnikovi:

PR #306, opened 2019-10-02
DB Information Schema - COLUMNS

buskamuza:
## Problem

This is a subset of https://github.com/magento/architecture/pull/264 , covering only COLUMNS and TABLES interfaces.

Different RDBMS return Information Schema (metadata) information in different format. Magento framework should allow to switch between different implementations in order to support those different versions of DBs, as well as differences between versions.

## Solution

Introduce granular interfaces, based in Information Schema standard (looking into MySQL and MariaDB definitions in this scope, links are in the document). Application can switch between different implementation based on the profile.

### DIFF
--- a/design-documents/database/declarative-schema-database-compatibility.md
+++ b/design-documents/database/declarative-schema-database-compatibility.md
@@ -65,7 +65,7 @@ abstract class InformationSchema\Table\Column
     public function getIsNullable(): bool {}
     public function getExtra(): Extra {}
     public function getComment() {}
-    public function getDefaultValue(): string {}
+    public function getDefaultValue(): ?string {}
 }
 ```
 
@@ -198,8 +198,6 @@ It has default value of `mysql4`. I propose to add 2 values: `mysql` and `mariad
 ## Considerations
 
 `\Magento\Framework\DB\Adapter\AdapterInterface` has methods like `describeTable` which return raw values from database engine. This is leaky interface. We should have interfaces that return normalized data and are compatible with all databases. Methods that don't return normalized data need to be deprecated on `\Magento\Framework\DB\Adapter\AdapterInterface`.
-
-`engine` field in declarative schema need to be deprecated.
 
 ## Resources
 
--- a/design-documents/declarative-schema-database-compatibility.md
+++ b/design-documents/declarative-schema-database-compatibility.md
@@ -1,0 +1,45 @@
+# Declarative schema database compatibility
+
+## About
+
+Declarative schema relies on querying data that may have different format depending on database. For example, here is the query from `Magento\Framework\Setup\Declaration\Schema\Db\MySQL\DbSchemaReader::readColumns`
+
+```
+SELECT 
+	COLUMN_NAME as 'name',
+	COLUMN_DEFAULT as 'default',
+	DATA_TYPE as 'type',
+	IS_NULLABLE as 'nullable',
+	COLUMN_TYPE as 'definition',
+	EXTRA as 'extra',
+	COLUMN_COMMENT as 'comment'
+FROM
+    information_schema.COLUMNS
+WHERE
+    TABLE_SCHEMA = 'magento'
+AND
+    TABLE_NAME = 'store_website'
+ORDER BY
+    ORDINAL_POSITION ASC;
+```
+
+MySQL returns NULL for COLUMN_DEFAULT and MariaDB returns NULL as a string.
+
+## Design
+
+Introduce concept of database adapters to allow normalize values in declarative schema for different databases.
+
+`\Magento\Framework\Setup\Declaration\Schema\Db\SchemaBuilder` should accept `\Schema\Db\MySQL\DbSchemaReaderFactory` that would allow to create database specific implementations of `DbSchemaReaderInterface` for given connection. `\Schema\Db\MySQL\DbSchemaReaderFactory` will depend on `\Magento\Framework\App\DeploymentConfig` to get configuration for connection, `model` will contain name of database adapter.
+
+Similar approach need to be taken for clients of the following interfaces
+1. `\Magento\Framework\Setup\Declaration\Schema\Db\DDLTriggerInterface`
+1. `\Magento\Framework\Setup\Declaration\Schema\Db\DbSchemaWriterInterface`
+1. `\Magento\Framework\Setup\Declaration\Schema\Db\DbDefinitionProcessorInterface`
+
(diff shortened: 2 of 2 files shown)


larsroettig:
@buskamuza can we add a section about **backward compatibly** =?

design-documents/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,209 @@
...
+
+```
+abstract class InformationSchema\Table\Column
+{
+    public function getTableName(): string {}
+    public function getName(): string {}
+    public function getIsNullable(): bool {}
+    public function getExtra(): Extra {}
+    public function getComment() {}
+    public function getDefaultValue(): string {}
```
Stepa4man: `?string` should be used

design-documents/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,209 @@
...
+
+`model` is the name of parameter that supposed to be used to configure database specific adapters. Somehow functionality behind it is absent.
+
+It has default value of `mysql4`. I propose to add 2 values: `mysql` and `mariadb` and have `mysql4` be alias for `mysql`.
+
+## Considerations
+
+`\Magento\Framework\DB\Adapter\AdapterInterface` has methods like `describeTable` which return raw values from database engine. This is leaky interface. We should have interfaces that return normalized data and are compatible with all databases. Methods that don't return normalized data need to be deprecated on `\Magento\Framework\DB\Adapter\AdapterInterface`.
+
+`engine` field in declarative schema need to be deprecated.
```
Stepa4man: Why?

buskamuza: Removed. It's part of original document.

design-documents/declarative-schema-database-compatibility.md
buskamuza: Updated

buskamuza:
Can we make `$connectionName` optional?

jean-bernard-valentaten:
> Can we make `$connectionName` optional?

I'm not sure I understand the context of this question? Where would you like to make `$connectionName` optional?

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,207 @@
...
+1. `\Magento\Framework\Setup\Declaration\Schema\Db\DbDefinitionProcessorInterface` (see usages of `\Magento\Framework\Setup\Declaration\Schema\Db\DefinitionAggregator`)
+
+New interfaces and types for managing table schema.
+
+```
+class InformationSchema\Table
+{
+    public function getSchema(): string {}
+    public function getName(): string {}
+    public function getEngine(): string {}
```
melnikovi: Propose to either to remove it or introduce enum for engine. I think we need to introduce enums for collation and charset. What is schema, definition of the table?

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,207 @@
...
+    public function getCollation(): string {}
+    public function getCharset(): string {}
+}
+```
+Note: auto increment is omitted intentionally. Auto increment should not be used for the application logic.
+
+```
+class InformationSchema\Extra
+{
+    public function getOnUpdate(): string {}
```
melnikovi: I think it might differ between databases, I would propose to introduce enums.

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,207 @@
...
+
+```
+abstract class InformationSchema\Table\Column
+{
+    public function getTableName(): string {}
+    public function getName(): string {}
+    public function getIsNullable(): bool {}
+    public function getExtra(): Extra {}
+    public function getComment() {}
+    public function getDefaultValue(): ?string {}
```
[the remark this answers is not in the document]
melnikovi: Should we remove table_name field? I also think that default_value might be better to move to specific types

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,207 @@
...
+}
+```
+
+```
+class InformationSchema\Table\VarcharColumn extends Column
+{
+    public const TYPE = 'VARCHAR';
```
melnikovi: I would remove it from here because it differs between database engines.

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+```
+class InformationSchema\Table\IntColumn extends Column
+{
+    public const TYPE = 'INT';
+
+    public function isUnsigned(): bool {}
+    public function getPadding(): int {}
+    public function getType(): IntEnum {}
```
melnikovi: Propose to have different types instead.

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+### Profiles
+
+The above interfaces will have multiple implementations, and the framework should be switching between them depending on the RDBMS it works with.
+The framework decides which implementations to use based on declared profiles.
+The profiles are declared in `di.xml` or as a separate configuration, with the following structure:
+
+```
+information_schema_profiles:
```
melnikovi: Should it be databse_information_schema_profiles?

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: I would not use an enum for the engine. Technically most RDBMS support having storage engines other than their default but this is mostly used in MySQL / MariaDB. Introducing an enum for engines could become a maintenance issue as MySQL / MariaDB evolve (e.g. as of MariaDB 10.4 the default storage engine is AriaDB which is an evolution of MyISAM).

Schema is (more or less) the database in which the table resides. SQL defines that a table resides in the schema of a catalog, hence the columns `information_schema.tables.catalog` and `information_schema.tables.schema`. MySQL / MariaDB do not support catalogs thus the schema reflects the database (as opposed to PostgreSQL supporting multiple schemas per catalog).

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: Agreed

buskamuza:
Closing in favor of https://github.com/magento/architecture/pull/318

PR #312, opened 2019-10-05
[GraphQL] Customer Orders API

joni-jones:
## Problem

The current GraphQL customer orders API has limited functionality and is inconsistent with the Magento Luma theme.

## Solution

Introduce changes in customer orders API with better customizability and extensibility while supporting the same set of data provided for Luma theme.

## Requested Reviewers

@paliarush 
@akaplya

### DIFF
--- a/design-documents/graph-ql/coverage/customer-orders.md
+++ b/design-documents/graph-ql/coverage/customer-orders.md
@@ -1,30 +1,31 @@
 # Customer Orders API
 
-## Overview
+# Overview
 
 The GraphQL API should provide a possibility to retrieve orders, shipments, invoices, credit memos for the logged in customer. The current schema allows fetching only simple order details and doesn't provide a possibility to fetch order details by a number.
 
 The proposed solution is deprecation of `customerOrders` query and `orders` field to the `customer` query:
 
 ```graphql
-# query to return a list of all customer orders
+@doc("Query to return a list of all customer orders")
 type Customer {
     orders (
         filter: CustomerOrdersFilterInput
-        currentPage: Int = 1 # current page of the customer order list. default is 1.
-        pageSize: Int = 20 # page size for the customer orders list. default is 20.
+        currentPage: Int = 1 @doc("current page of the customer order list. default is 1")
+        pageSize: Int = 20 @doc("page size for the customer orders list. default is 20")
     ): CustomerOrders
 }
 
+@doc("Allows to extend the list of search criteria for customer orders")
 input CustomerOrdersFilterInput {
     order_number: String!
 }
 
-# collection of customer orders.
+@doc("Collection of customer orders")
 type CustomerOrders {
-    items: [CustomerOrder]! # collection of customer orders that contains individual order details.
+    items: [CustomerOrder]! @doc("collection of customer orders that contains individual order details.")
     page_info: SearchResultPageInfo
-    total_count: Int
+    total_count: Int @doc("the total count of customer orders")
 }
 ```
 
@@ -35,17 +36,17 @@ type CustomerOrders {
 The proposed type for the customer order might look like:
 
 ```graphql
-# order details
+@doc("Customer order details")
 type CustomerOrder {
-    order_date: String! # date when the order was placed.
-    status: String! # current status of the order.
-    order_number: String! # order number.
-    items: [OrderItem]! # collection of all the items purchased
-    prices: SalesPricesInterface! # prices details for the order.
-    invoices: [Invoice]! # invoice list for the order.
-    credit_memos: [CreditMemo]! # credit memo list for the order.
-    shipments: [OrderShipment]! # shipping list for the order.
-    payment_methods: [PaymentMethod]! # payment details for the order.
+    order_date: String! @doc("date when the order was placed")
+    status: String! @doc("current status of the order")
+    order_number: String! @doc("order number")
+    items: [OrderItem]! @doc("collection of all the items purchased")
+    prices: SalesPricesInterface! @doc("prices details for the order")
+    invoices: [Invoice]! @doc("invoice list for the order")
+    credit_memos: [CreditMemo]! @doc("credit memo list for the order")
+    shipments: [OrderShipment]! @doc("shipment list for the order")
+    payment_methods: [PaymentMethod]! @doc("payment details for the order")
 }
 ```
 
@@ -56,47 +57,40 @@ type CustomerOrder {
 The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
 
 ```graphql
+@doc("Interface to reprent order/invoice/shipment/credit memo items")
 interface SalesItemInterface {
-    name: String # name of the base product.
-    sku: String! # sku of the base product.
-    url: String # url of the base product.
-    sale_price: Float! # sale price for the base product including all the child products and selected options.
-    discounts: [Discount] # final discount information for the base product including discounts on options and child products.
-    selected_options: [String!] # selected options for the base product. for e.g color, size etc.
-    entered_options: [SalesItemOption] # entered option for the base product. for e.g logo image etc.
+    name: String @doc("name of the base product")
+    sku: String! @doc("sku of the base product")
+    url: String @doc("url of the base product")
+    sale_price: Float! @doc("sale price for the base product including selected options")
+    discounts: [Discount] @doc("final discount information for the base product including discounts on options")
+    parent_name: String @doc("name of parent product like configurable or bundle")
+    parent_sku: String @doc("SKU of parent product like configurable or bundle")
+    parent_url: String @doc("URL of parent product in the catalog")
+    selected_options: [SalesItemOption] @doc("selected options for the base product. for e.g color, size etc.")
+    entered_options: [SalesItemOption] @doc("entered option for the base product. for e.g logo image etc.")
 }
 
+@doc("Represents sales item options like selected or entered")
 type SalesItemOption {
-    id: String!
-    value: String!
+    id: String! @doc("name of the option")
+    value: String! @doc("value of the option")
 }
 ```
 
 The `SalesItemInterface` will be implemented by the following types:
 
 ```graphql
-# Order Product implementation of OrderProductInterface
+@doc("Order Product implementation of OrderProductInterface")
 type OrderItem implements SalesItemInterface {
-    quantity_ordered: Float! # number of items
-    quantity_shipped: Float! # number of shipped items
-    quantity_refunded: Float! # number of refunded items
-    quantity_invoiced: Float! # number of invoiced items
-    quantity_backordered: Float! # number of back ordered items
-    quantity_canceled: Float! # number of cancelled items
-    quantity_returned: Float! # number of returned items
-    status: String! # the status of order item
-    children: [OrderChildItem]
-}
-
-type OrderChildItem implements SalesItemInterface{
-    quantity_ordered: Float! # number of items items
-    quantity_shipped: Float! # number of shipped items
-    quantity_refunded: Float! # number of refunded items
-    quantity_invoiced: Float! # number of invoiced items
-    quantity_backordered: Float! # number of back ordered items
-    quantity_canceled: Float! # number of cancelled items
-    quantity_returned: Float! # number of returned items
-    status: String! # the status of order item
+    quantity_ordered: Float! @doc("number of items")
+    quantity_shipped: Float! @doc("number of shipped items")
+    quantity_refunded: Float! @doc("number of refunded items")
+    quantity_invoiced: Float! @doc("number of invoiced items")
+    quantity_backordered: Float! @doc("number of back ordered items")
+    quantity_canceled: Float! @doc("number of cancelled items")
+    quantity_returned: Float! @doc("number of returned items")
+    status: String! @doc("the status of order item")
 }
 ```
 
@@ -105,11 +99,11 @@ type OrderChildItem implements SalesItemInterface{
 To provide more customization for different payment solutions, the payment method will be represented by own type instead of simple string:
 
 ```graphql
-#  payment method used to pay for the order
+@doc("Payment method used to pay for the order")
 type PaymentMethod {
-    name: String! # payment method name for e.g Braintree, Authorize etc
-    type: String! # payment method type used to pay for the order for e.g Credit Card, PayPal etc.
-    additional_data: [KeyValue] # additional data per payment method type
+    name: String! @doc("payment method name for e.g Braintree, Authorize etc.")
+    type: String! @doc("payment method type used to pay for the order for e.g Credit Card, PayPal etc.")
+    additional_data: [KeyValue] @doc("additional data per payment method type")
 }
 ```
 
@@ -120,16 +114,18 @@ type PaymentMethod {
 As entities like order, invoice, credit memo might have complex prices type:
 
 ```graphql
+@doc("Interface to provide sales prices")
 interface SalesPricesInterface {
-    subtotal: Float! # subtotal amount excluding, shipping, discounts and tax
-    discounts: [Discount] # applied discounts
-    tax: Float! # applied taxes
-    grand_total: Float! # final total amount including shipping and taxes
-    base_grand_total: Float! # final total amount in base currency
+    subtotal: Float! @doc("subtotal amount excluding, shipping, discounts and tax")
+    discounts: [Discount] @doc("applied discounts")
+    tax: Float! @doc("applied taxes")
+    grand_total: Float! @doc("final total amount including shipping and taxes")
+    base_grand_total: Float! @doc("final total amount in base currency")
 }
 ​
+@doc("Order prices details")
 type OrderPrices implements SalesPricesInterface {
-​   shipping_handling: Float! # shipping and handling for the order
+​   shipping_handling: Float! @doc("shipping and handling for the order")
 }
 ```
 
@@ -138,21 +134,19 @@ type OrderPrices implements SalesPricesInterface {
 The invoice entity will have the similar to the order schema:
 
 ```graphql
+@doc("Invoice details")
 type Invoice {
-    number: String! # user friendly identifier for the invoice
-    prices: InvoicePrices! # invoice prices details
-    items: [InvoiceItem]! # invoiced product details
+    number: String! @doc("user friendly identifier for the invoice")
+    prices: InvoicePrices! @doc("invoice prices details")
+    items: [InvoiceItem]! @doc("invoiced product details")
 }
 
+@doc("Invoice item details")
 type InvoiceItem implements SalesItemInterface{
-    quantity_invoiced: Float! # number of invoiced items
-    children: [InvoiceChildItem]
+    quantity_invoiced: Float! @doc("number of invoiced items")
 }
 
-type InvoiceChildItem implements SalesItemInterface{
-    quantity_invoiced: Float! # number of invoiced items
-}
-
+@doc("Invoice prices details")
 type InvoicePrices implements SalesPricesInterface {
   
 }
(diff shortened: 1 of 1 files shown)


nrkapoor:
@danielrenaud @cpartica @prabhuram93 @dthampy @avattam06 - team, please review the order details schema PR.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,210 @@
...
+}
+
+type ShipmentChildProduct implements ChildOrderProductInterface{
+    shipped: Float! # umber of shipped items
+}
+```
+
+## Additional Types
+
+The `NameValue` type will provide a possibility to use key-value pairs:
```
danielrenaud: Is this supposed to be KeyValue, or are fields currently using KeyValue actually supposed to be NameValue?

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,210 @@
...
+
+# order child product interface
+interface ChildOrderProductInterface {
+    name: String # name of the child product.
+    sku: String # sku of the child product.
+    url: String # url of the child product.
+    final_price: Float! # final price for the child product including all the selected options
+    discounts: [Discount] # final discount information for the child product including discounts on options
+    selected_options: [KeyValue] # selected options for the child product. for e.g color, size etc
+    entered_options: [String!] # "entered option for the child product. for e.g logo image etc.
```
danielrenaud: What would this look like, for example, a text field labeled "Personal Note"? Since the return value is a string is this serialized somehow?

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,210 @@
...
+### Order Item
+
+The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
+
+```graphql
+interface OrderProductInterface {
+    product_name: String # name of the base product.
+    sku: String! # sku of the base product.
+    url: String # url of the base product.
+    child_products: [ChildOrderProductInterface] # child products in case of bundled products.
```
danielrenaud: Is bundle the only case? What about configurable?

danielrenaud:
note: selected_options and entered_options will base64_encoded values. 
see: https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md#add-simple-product-to-cart

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: Configurable products are presented by simple with options. Each option is a variation of a simple product.

fooman:
Order Item currently has these statuses 

```
class Item extends AbstractModel implements OrderItemInterface
{
    const STATUS_PENDING = 1;

    // No items shipped, invoiced, canceled, refunded nor backordered
    const STATUS_SHIPPED = 2;

    // When qty ordered - [qty canceled + qty returned] = qty shipped
    const STATUS_INVOICED = 9;

    // When qty ordered - [qty canceled + qty returned] = qty invoiced
    const STATUS_BACKORDERED = 3;

    // When qty ordered - [qty canceled + qty returned] = qty backordered
    const STATUS_CANCELED = 5;

    // When qty ordered = qty canceled
    const STATUS_PARTIAL = 6;

    // If [qty shipped or(max of two) qty invoiced + qty canceled + qty returned]
    // < qty ordered
    const STATUS_MIXED = 7;

    // All other combinations
    const STATUS_REFUNDED = 8;

    // When qty ordered = qty refunded
    const STATUS_RETURNED = 4;
```

so we either should provide all the underlying data to the frontend or probably better to add a status field on the order items.

We should also make a note for the implementation that not all information is to be exposed to the frontend. For example not all order statuses should be displayed and likely some information on payment method's additional data should not be taken verbatim as a lot of this might be internal/sensitive data.

fooman:
It seems the current display of order totals works off a fixed list. To keep this extensible similar to how custom totals currently work it would be good to have at least a way to communicate the sort order of OrderPricingInterface.

Currently it is possible to display price information in both the base and order currency if they are different. Any chance we can extend the schema to cover base currency too?

fooman:
It's currently possible in Luma to access information about your own order even if placed as a guest. Not sure if this proposal could be extended to that scenario and/or if this needs to be a separate effort.

joni-jones:
> Currently it is possible to display price information in both the base and order currency if they are different. Any chance we can extend the schema to cover base currency too?

Actually our vision that base amount/currency should not be provided by GraphQL API and we deprecate such fields in the existing GraphQL implementation (as an example https://github.com/magento/graphql-ce/issues/889).

@fooman could you provide scenarios when a storefront application needs amounts in the base currency?

joni-jones:
> so we either should provide all the underlying data to the frontend or probably better to add a status field on the order items.

Each item type has own set of fields, for example, `OrderItem` has `ordered:shipped:invoiced:refunded`, `InvoiceItem` has `invoiced` so on. But seems like it makes sense to have a status for the `OrderItem` type.

joni-jones:
> We should also make a note for the implementation that not all information is to be exposed to the frontend. For example not all order statuses should be displayed and likely some information on payment method's additional data should not be taken verbatim as a lot of this might be internal/sensitive data.

The mechanism will be the same as for the current storefront. Right now, order statuses are mapping to storefront statuses. The same for the payment additional data. The `privateInfoKeys` and `paymentInfoKeys` configuration allow restricting information available on storefront.

joni-jones:
> It seems the current display of order totals works off a fixed list. To keep this extensible similar to how custom totals currently work it would be good to have at least a way to communicate the sort order of OrderPricingInterface.

Each entity type has own pricing type like `OrderPricing`, `InvoicePricing`, `CreditMemoPricing` but the entity type declaration uses `OrderPricingInterface`. Maybe, it makes sense to specify exact pricing types for entities types.

@fooman could you clarify about sort order? GraphQL allows returning fields in the same order as they were requested from storefront application.

joni-jones:
> It's currently possible in Luma to access information about your own order even if placed as a guest. Not sure if this proposal could be extended to that scenario and/or if this needs to be a separate effort.

I believe it should be a separate proposal, the current describes API for the customer orders but for guest scenarios, most of the types can be re-used.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+# query to return a list of all customer orders
+type Customer {
+    orders (
+        filter: CustomerOrderFilterInput
+        currentPage: Int = 1 # current page of the customer order list. default is 1.
+        pageSize: Int = 20 # page size for the customer orders list. default is 20.
+    ): CustomerOrders
+}
+
+input CustomerOrderFilterInput {
```
paliarush:
```suggestion
input CustomerOrdersFilterInput {
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+
+> Right now, we don't introduce the filter input for such entities like invoice, credit memo, shipment as all these entities are related to the same order and GraphQL resolvers anyway at first resolve the order and after that other child entities. But, in future, such filter input can be introduced as optional argument.
+
+## Order Type Schema
+
+The proposed type for the customer order might look like:
+
+```graphql
+# order details
+type Order {
```
paliarush: Better to namespace the name to allow future extensibility. For example, there could be guest order representations. This should be done across the schema, not only for this specific type.
```suggestion
type CustomerOrder {
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    order_date: String! # date when the order was placed.
+    status: String! # current status of the order.
+    order_number: String! # order number.
+    items: [OrderItem] # collection of all the items purchased
```
paliarush: Looks like we can guarantee at least empty array for items, invoices etc.
```suggestion
    items: [OrderItem]! # collection of all the items purchased
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    order_pricing: OrderPricingInterface! # pricing details for the order.
+    invoices: [Invoice] # invoice list for the order.
+    credit_memos: [CreditMemo] # credit memo list for the order.
+    shipments: [OrderShipment] # shipping list for the order.
+    payment_method: [PaymentMethod] # payment details for the order.
```
paliarush: We should change `payment_method` to `payment_methods` OR change it to non-array type

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    additional_data: [KeyValue] # additional data per payment method type
+}
+```
+
+### Pricing Schema
+
+As entities like order, invoice, credit memo might have complex pricing type:
+
+```graphql
+interface OrderPricingInterface {
```
paliarush:
```suggestion
interface SalesPricesInterface {
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+
+The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
+
+```graphql
+interface SalesProductInterface {
+    name: String # name of the base product.
+    sku: String! # sku of the base product.
+    url: String # url of the base product.
+    final_price: Float! # final price for the base product including all the child products and selected options.
+    discounts: [Discount] # final discount information for the base product including discounts on options and child products.
```
paliarush:
```suggestion
    applied_discounts: [Discount] # final discount information for the base product including discounts on options and child products.
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+### Order Item
+
+The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
+
+```graphql
+interface SalesProductInterface {
+    name: String # name of the base product.
+    sku: String! # sku of the base product.
+    url: String # url of the base product.
+    final_price: Float! # final price for the base product including all the child products and selected options.
```
paliarush:
```suggestion
    sale_price: Float! # final price for the base product including all the child products and selected options.
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    shipped: Float! # umber of shipped items
+}
+```
+
+## Additional Types
+
+The `KeyValue` type will provide a possibility to use key-value pairs:
+
+```graphql
+type KeyValue {
```
paliarush: It would be better to introduce specific types.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    entered_options: [KeyValue] # entered option for the base product. for e.g logo image etc.
+}
+```
+
+The `SalesProductInterface` will be implemented by the following types:
+
+```graphql
+# Order Product implementation of OrderProductInterface
+type OrderItem implements SalesProductInterface {
+    ordered: Float! # number of items items
```
paliarush: Similar fix should be done for all other fields in this type.
```suggestion
    quantity_ordered: Float! # number of items items
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    selected_options: [String!] # selected options for the base product. for e.g color, size etc.
+    entered_options: [KeyValue] # entered option for the base product. for e.g logo image etc.
+}
+```
+
+The `SalesProductInterface` will be implemented by the following types:
+
+```graphql
+# Order Product implementation of OrderProductInterface
+type OrderItem implements SalesProductInterface {
```
paliarush: Does not look right because usually Item has reference to a product and quantity. At the same Item is not Product.

The same question applies to other "items" implementing `SalesProductInerface`.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+### Pricing Schema
+
+As entities like order, invoice, credit memo might have complex pricing type:
+
+```graphql
+interface OrderPricingInterface {
+    subtotal: Float! # order subtotal excluding, shipping, discounts and tax
+    discounts: [Discount] # all the discounts applied to the order
```
paliarush:
```suggestion
    applied_discounts: [Discount] # all the discounts applied to the order
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+    tax: Float! # tax applied to the order
```
paliarush:
```suggestion
    applided_tax: Float! # tax applied to the order
```

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+## Invoice Type Schema
+
+The invoice entity will have the similar to the order schema:
+
+```graphql
+type Invoice {
+    number: String! # user friendly identifier for the invoice
+    pricing: OrderPricingInterface! # invoice pricing details
```
paliarush: Should it be as follows?
```suggestion
    pricing: InvoicePricing! # invoice pricing details
```

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: What will be the difference between customer and guest orders? They will have the same set of attributes the difference only in the context and authorization rights.

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: Does it make sense to rename them? If discounts are presented in order it means they already applied. The same is true for other amounts.

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: What kind of reference do you like to see? `SalesProductInterface` provides `url` for a catalog product?

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: The idea is to have generic key => value type because we have a lot of custom data which doesn't have a strict structure, for example, payment details, each payment integration has own set of additional data and we should provide a possibility to customize it by 3rd party integrations.

fooman:
@joni-jones my own website is a good use case for multi-currency. We are New Zealand based and our payment gateway only accepts New Zealand Dollars (NZD). The base currency of our website is NZD. To make it easier to understand for international customers what is the approximate cost (the final cost depends on the credit card used and if it has a different exchange rate to the exchange rate service used and if the credit card adds any extra conversion fees) we add popular currencies like USD, EUR, AUD, etc.

During checkout we display a note like Grand total in USD is X, your credit card charge will be Y in NZD. After the order is placed we display both base and order currency on orders, invoices and creditmemos.

fooman:
> @fooman could you clarify about sort order? GraphQL allows returning fields in the same order as they were requested from storefront application.

There is [sales.xml](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/etc/sales.xml#L11) which specifies a sort_order which doesn't seem to carry to GraphQl.

fooman:
>Each item type has own set of fields, for example, OrderItem has ordered:shipped:invoiced:refunded, InvoiceItem has invoiced so on. Could you provide some cases when such statuses might be useful?

An Item can currently have 9 different statuses - see [here](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/Model/Order/Item.php#L25)
I suggest adding this status so that the frontend can display it to customers.

Also I believe qtyBackordered should get included - see its use [here](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/Model/Order/Item.php#L314).

joni-jones:
> There is [sales.xml](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/etc/sales.xml#L11) which specifies a sort_order which doesn't seem to carry to GraphQl.

The order of returned fields it's one of the GrapqhQL features, I think, it might confuse developers if fields are returned in a different order than they requested. And again, GraphQL allows us to specify a needed order via query and seems like no reason to use sales.xml.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,211 @@
...
+    value: String!
+}
+```
+
+The `SalesItemInterface` will be implemented by the following types:
+
+```graphql
+# Order Product implementation of OrderProductInterface
+type OrderItem implements SalesItemInterface {
+    qty_ordered: Float! # number of items items
```
paliarush: Please rename all `qty` to quantity across the PR. We have a task in backlog to do this for all existing fields too.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,211 @@
...
+> Right now, we don't introduce the filter input for such entities like invoice, credit memo, shipment as all these entities are related to the same order and GraphQL resolvers anyway at first resolve the order and after that other child entities. But, in future, such filter input can be introduced as optional argument.
+
+## Order Type Schema
+
+The proposed type for the customer order might look like:
+
+```graphql
+# order details
+type CustomerOrder {
+    order_date: String! # date when the order was placed.
```
paliarush: Please convert all field comments to '@doc' directives.

design-documents/graph-ql/coverage/customer-orders.md
paliarush: Please fix the same issue in other places.
```suggestion
    qty_ordered: Float! # number of items
```

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: Does it make sense to do for the proposal? It's not a schema file.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,214 @@
...
+    name: String # name of the base product.
+    sku: String! # sku of the base product.
+    url: String # url of the base product.
+    sale_price: Float! # sale price for the base product including all the child products and selected options.
+    discounts: [Discount] # final discount information for the base product including discounts on options and child products.
+    selected_options: [String!] # selected options for the base product. for e.g color, size etc.
+    entered_options: [SalesItemOption] # entered option for the base product. for e.g logo image etc.
+}
+
+type SalesItemOption {
```
akaplya: What this options about?

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,214 @@
...
+The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
+
+```graphql
+interface SalesItemInterface {
+    name: String # name of the base product.
+    sku: String! # sku of the base product.
+    url: String # url of the base product.
+    sale_price: Float! # sale price for the base product including all the child products and selected options.
+    discounts: [Discount] # final discount information for the base product including discounts on options and child products.
+    selected_options: [String!] # selected options for the base product. for e.g color, size etc.
```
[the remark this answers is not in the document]
joni-jones: It will be base64 encoded string, the same as for cart operations schema.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -0,0 +1,214 @@
...
+> The order `status` should be filtered in the same way as for Luma via `Order Status` and `Visible On Storefront` configuration 
+
+### Order Item
+
+The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
+
+```graphql
+interface SalesItemInterface {
+    name: String # name of the base product.
+    sku: String! # sku of the base product.
```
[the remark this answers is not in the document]
joni-jones: It has `selected_options`, `entered_options` and `children` on the type level.

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: It's a type for the customer entered option.

joni-jones:
@fooman I've made some changes. Do you have any ideas/questions?

PR #313, opened 2019-10-05
Update README.md

Parikshit-Hooda:
## Problem

typo (minotes)

## Solution
fix typo (minotes -> minutes)


Parikshit-Hooda:
@melnikovi @buskamuza @paliarush

PR #314, opened 2019-10-07
WIP: Frontend Template Versioning

nathanjosiah:
## Problem

The problem described here is twofold
- Magento themes that override templates creates a series of problems for the merchants, theme developers, and end-users. Some of these problems directly compromise security.
- We have no backwards compatibility rules or guidelines for templates.

### Example (from design document):

Product gallery template for 2.3.1 was large and contained all sorts of HTML and JS and JS config. https://github.com/magento/magento2/blob/2.3.1/app/code/Magento/Catalog/view/frontend/templates/product/view/gallery.phtml
Product gallery template for 2.3.2 was dramatically changed and hardly contains anything. https://github.com/magento/magento2/blob/2.3.2/app/code/Magento/Catalog/view/frontend/templates/product/view/gallery.phtml

If I have a template override based on 2.3.1 for gallery.phtml and upgrade Magento to 2.3.2 how do I know my override is compatible? How do I update the template? How would I even know this change was made to the template I'm overriding? Maybe everything looks fine in my template override but it was a security fix, how do I know if my override needs to be patched or if the changes were cosmetic? Essentially, how can I answer the basic question of "are my overridden templates up-to-date regardless of how customized they are?"

## Solution

Create a template versioning system with backwards compatibility rules for version changes that also gives the merchant visibility into the compatibility of the theme(s) they are using.

## Requested Reviewers

@DrewML

### DIFF
--- a/design-documents/template-versioning.md
+++ b/design-documents/template-versioning.md
@@ -17,13 +17,21 @@ To make this problem even worse there have no rules surrounding backwards compat
 
 These are very common scenarios and they have large security implications as well. Let's say there are some security fixes that spans across many templates. This could be the simple addition of a new hidden field, or something else like a new escapement policy. Whether it affects one or all of our template files, only a select few will actually ever be updated on merchant websites due to overrides and the difficulty of upgrading the templates.
 
-### Design
+#### Example
 
+Product gallery template for 2.3.1 was large and contained all sorts of HTML and JS and JS config. https://github.com/magento/magento2/blob/2.3.1/app/code/Magento/Catalog/view/frontend/templates/product/view/gallery.phtml
+Product gallery template for 2.3.2 was dramatically changed and hardly contains anything. https://github.com/magento/magento2/blob/2.3.2/app/code/Magento/Catalog/view/frontend/templates/product/view/gallery.phtml
+
+If I have a template override based on 2.3.1 for gallery.phtml and upgrade Magento to 2.3.2 how do I know my override is compatible? How do I update the template? How would I even know this change was made or needs to be made? Maybe everything looks fine but it was a security fix, how do I know if my override needs to be patched or if the changes were cosmetic?  
+
+### Design
 
 At a very high level this proposal aims to introduce several new things:
 
 - A template versioning system with backwards compatibility rules for version changes
 - Merchant visibility into theme compatibility
+
+In an attempt to keep the scope of this design feasible, only the file fallback mechanism is being addressed at this time. Any other template overrides through manually calling block methods or other forms of overrides are out of scope.
 
 #### Template versioning
 
@@ -33,9 +41,8 @@ A system of rules for html and phtml files using a new `@version x.x.x` annotati
 
 When a merchant installs a theme or upgrades their magento instances it would be helpful for them and for the theme developers to know that what they are using may not be compatible with the version of Magento installed. This will alert the merchant that their theme may not have the latest security fixes and that the theme needs to be updated. Not just for security fixes but other issues as well. They may be notified through some admin notification upon them install or magento upgrade and a new theme dashboard could be created to show which theme files are incompatible.
 
-Maybe this can also include a simple tool that allows theme developers to easily view what changes have been made since the last version upgrade of the templates so they can patch their templates. Seeing the difference between their template and the Magento template isn't helpful in most cases since other unrelated customizations will have been made. 
+Maybe this can also include a simple tool that allows theme developers to easily view what changes have been made since the last version upgrade of the templates so they can patch their templates. Seeing the difference between their template and the Magento template isn't helpful in most cases since other unrelated customizations will have been made. This is a stretch goal and is already available in 3rd party tools. 
   
-
 <!-- In this section provide relevant details at a high level, including the introduction of any new technologies being utilized for the design. 
 
 Hints:
--- a/design-documents/template-versioning.md
+++ b/design-documents/template-versioning.md
@@ -1,0 +1,87 @@
+## Magento Frontend Template Versioning
+
+### Terms
+
+<!-- Describe any new terms used in the document -->
+
+### Overview
+
+Let's say ModuleA has template foo.phtml and CoolTheme overrides foo.phtml. Magento makes changes to foo.phtml and other files in ModuleA. Now CoolTheme has several problems:
+
+- Hard to know if there were changes to templates it has replaced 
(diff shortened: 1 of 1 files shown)


design-documents/template-versioning.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+At a very high level this proposal aims to introduce several new things:
+
+- A template versioning system with backwards compatibility rules for version changes
+- Merchant visibility into theme compatibility
+
+#### Template versioning
+
+A system of rules for html and phtml files using a new `@version x.x.x` annotation that allows templates to be updated in a way that allows developers to not only answer the problems listed above but also to handle the updating of their overrides in a much more controlled and expected way. The rules for versioning would also aim to prevent developers from creating backwards incompatible template changes between magento versions.
```
DrewML: Likely a naive question - do we use the `@version x.x.x` annotations anywhere else, or would this be a new pattern for versioning?

nathanjosiah: We don't. We use `@since` for some `@api` related code but that doesn't solve the problem we are trying to solve here. We could definitely use `@since` as well as part of the new rules.

DrewML:
Injecting my personal opinion, even though this is still in a draft state (I'm the worst):

If the options are between:

1. More rules/constraints on template modifications
2. A tool that better surfaces changes that agencies/SIs/etc will need to address

I'd strongly prefer the route of option 2. But we should _definitely_ elicit community input on that, because we often guess wrong.

design-documents/template-versioning.md
DrewML: > We could definitely use @since as well as part of the new rules.

No idea what we use `@since` for so I won't comment there :)

Can you give an example (comment or in doc, either way) of what this would look like in a template with some modifications?

nathanjosiah: `@since` indicated when a specific class, method, parameter, etc, has been added. It says that thing has been available `@since 100.0.2` for example. It doesn't solve the problem of changes discussed but we could annotate DOM features or output like WordPress does. We still need something to indicate when and how significant the template was modified.

hostep:
Such a tool already exists, been using it for a while now, pretty happy with it, is not only for frontend changes, is not the most user friendly, but works pretty well: https://github.com/AmpersandHQ/ampersand-magento2-upgrade-patch-helper/

The upgrade from Magento 2.3.2 to 2.3.3 was especially gruesome due to sooooo many changes in `.phtml` files, almost spend 2 full days on making a theme compatible again and have all security-related changes in place.
Some tool which could help reduce this time in an automated fashion would be very awesome, but very very very hard to do.

DrewML:
@hostep that's super cool - you always seem to have a good link or two handy 😄 

Could see something like that working well with the UX of a tool like [Facebook's codemod](https://github.com/facebook/codemod), that can help walk a developer through a list of changes file-by-file.

nathanjosiah:
Thanks @hostep for this recommendation! I'm definitely not against an external tool meeting this need. This is more of a secondary need in the scope of this issue but I would love to see it addressed at the same time if possible.

fooman:
As all template files are delivered as part of a module could this be folded into https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/codebase-changes.html ?

As template customisation is the most common customisation point I think it could be added to this list: https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/codebase-changes.html#version-increase-scenarios

With the efforts underway for service isolation templates would be in a smaller module which then could be versioned on its own.

nathanjosiah:
@fooman Very nice suggestion! While this PR isn't specifically about the higher versioning, it makes sense to have some rules surrounding the cascading versioning consequences of making template changes. Especially since we don't have ANY right now.

design-documents/template-versioning.md
DrewML: > we could annotate DOM features or output like WordPress does

Would _love_ to see any references to that in this document as you flesh out ideas more, just for the sake of prior art research for the rest of us reviewing.

I have never touched WP so I'm ignorant here :)

DrewML:
Thanks for working on this 👍. Needs more concrete details, but good direction for us to be heading in.

design-documents/template-versioning.md
nathanjosiah: Sure. I'll update the doc in the near future but for now, here's an example of a woocommerce template with both a version in the template itself as well as annotated features for when they were added https://github.com/woocommerce/woocommerce/blob/master/templates/myaccount/dashboard.php#L47

design-documents/template-versioning.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Merchant visibility
+
+When a merchant installs a theme or upgrades their magento instances it would be helpful for them and for the theme developers to know that what they are using may not be compatible with the version of Magento installed. This will alert the merchant that their theme may not have the latest security fixes and that the theme needs to be updated. Not just for security fixes but other issues as well. They may be notified through some admin notification upon them install or magento upgrade and a new theme dashboard could be created to show which theme files are incompatible.
+
+Maybe this can also include a simple tool that allows theme developers to easily view what changes have been made since the last version upgrade of the templates so they can patch their templates. Seeing the difference between their template and the Magento template isn't helpful in most cases since other unrelated customizations will have been made. This is a stretch goal and is already available in 3rd party tools.
```
hostep: @nathanjosiah: There was [a presentation](https://events.bizzabo.com/213706/agenda/session/179872) at Magento Live in Amsterdam last week, where some people from Magento told us that they already started working on such a tool.
But I'm not sure if they were looking at frontend or backend specifically, that wasn't really clear yet I think.
Maybe @chrishedge can help clarify this?

Vinai:
Not sure if I can contribute meaningfully to this PR discussion,  but my process looks like this:

1. diff the theme files of the previous version with the current version.
2. read through the diff and figure out which templates contained meaningful changes (e.g. not comments or whitespace).
3. go over the templates with meaningful changes and filter the ones that are overridden in custom themes.
4. compare the diff to the custom theme and apply the changes manually.
5. use lots of comments in the customized templates to indicate what is custom and what is original.
6. test

I haven't a special purpose tool so far.

nathanjosiah:
I think this problem is largely solved by the new Safe Upgrade Tool

hostep:
@nathanjosiah: is the SUT already publicly available, if yes, can you put a link to it? Thanks!

PR #315, opened 2019-10-07
[ARC-596] Introduce Server-Side Rendering doc

jedmao:
# [ARC-596](https://jira.corp.magento.com/browse/ARC-596)

## Problem

There seems to be some confusion and/or misconceptions about server-side rendering, how it works and what's required to make it work properly.

## Solution

This document addresses the above issues, explaining what SSR is, how things worked before and how it solves some problems with modern-day JavaScript applications, specifically React.


jedmao:
@paliarush see 70518b9.
- [x] Added some _italic_ text at the top of the document that explains the context of SSR to be within the bounds of React, PWA and the BFF layer.
- [x] Clarified at which points the GraphQL API requests are made.
- [x] Added a Requirements section, which only has 1 bullet point in it, but it's helpful.
- [x] Added an Extensibility section, but didn't want to go to deep, as it warrants its own document.

design-documents/frontend/server-side-rendering.md
```diff
@@ -0,0 +1,129 @@
...
+_SSR belongs to the BFF layer of [service isolation](https://github.com/magento/architecture/tree/master/design-documents/service-isolation)._
+
+SSR can be a challenge to implement correctly, but the benefits have the potential to provide the unparalleled performance you seek to decrease bounce rate and drive more business.
+
+## Pros
+
+- Faster, first meaningful paint.
+- No need to render initial content via JavaScript on the client.
+- SEO/bot support.
+- Better SEO performance.
```
DrewML: - `SEO/bot support`
- `Better SEO performance.`

These seem duplicative, maybe just combine to 1

design-documents/frontend/server-side-rendering.md
```diff
@@ -0,0 +1,129 @@
...
+## Cons
+
+- User may see some content before it's interactive.
+- Difficult or tedious to implement.
+- May require additional configuration and refactoring.
+- Some features [not yet supported](#not-supported-yet).
+
+## How it works
+
+The basic idea is that we don't want the web browser (client) to be responsible for rendering content via JavaScript, because JavaScript is notoriously slow, especially at rendering DOM and especially on older or mobile devices. Unnecessary rendering also consumes more energy, which has the potential to suck a mobile device's battery dry. :battery:
```
DrewML: > we don't want the web browser (client) to be responsible for rendering content via JavaScript

Probably would be more clear to say

> we don't want the web browser (client) to be responsible for rendering the initial page's content via JavaScript

The first statement seems not-ideal because:

1. There will still be JS-based rendering (just not for the first bits of DOM)
2. SSR won't kick in for subsequent page navs in the same tab (history API + React client-side)

design-documents/frontend/server-side-rendering.md
```diff
@@ -0,0 +1,129 @@
...
+If we leverage the HTTP server to do as much heavy lifting as we can, some of this heavy lifting (non-sensitive pages) can be cached as well, resulting in a faster, better user experience.
+
+### Without SSR
+
+1. A web browser (client) requests a URL for a particular route (e.g., `/`).
+1. The HTTP server runs the appropriate code for that route.
+   - HTML is partially generated with empty containers that need to be filled and rendered by JavaScript after the initial page load (bad for SEO).
+1. The web browser (client) receives and processes the HTML, but the user may see some empty content that hasn't been filled in yet.
+1. JavaScript is loaded and processed, including the React library.
+1. Ideally, a single GraphQL API request is made to collect the data required to fill in the empty containers with content.
```
DrewML: Think this is a bit too specific:

> Ideally, a single GraphQL API request is made to collect the data required to fill in the empty containers with content

Can we make this description bit more general? A single API request for an entire UI is not ideal because it doesn't differentiate between high and low priority content, so critical functionality gets delayed due to network time for low priority content.

Maybe just say:

> The client-side JavaScript makes a request to one or several APIs for the initial data needed to render the page

DrewML: > Ideally, a single GraphQL API request is made to collect the data required to fill in the empty containers with content.

Worth noting that this isn't always the case. Above this document says:

> HTML is partially generated with empty containers that need to be filled and rendered by JavaScript after the initial page load (bad for SEO).

It's not uncommon to inline the data needed for the first render in a `script` tag that assigns to some known global

design-documents/frontend/server-side-rendering.md
```diff
@@ -0,0 +1,129 @@
...
+1. React renders the final HTML into the above-mentioned empty containers by combining the data received with the HTML defined in the React components.
+1. Once content is seen, it becomes immediately interactive.
+
+### With SSR
```
DrewML: We spoke a bit about this offline, but this section is missing some steps that add execution time.

There is a note that says:

> Ideally, a single GraphQL API request is made from this HTTP server, collecting all the data that's required to render the page.
>   - HTML is fully generated [via React methods](https://reactjs.org/docs/react-dom-server.html) with no empty containers. This HTML can be [streamed](https://reactjs.org/docs/react-dom-server.html#rendertonodestream) and cached

How do you know what data to fetch here yet? Unfortunately, for most apps (especially with colocated data fetching) you don't. That's why folks doing SSR today (in the large majority of cases) have to do a 2-pass render. It looks something like this

1. Mount the the initial React App with the proper routing data, based on the request
2. Walk the rendered React tree manually, to collect data requirements for each component.
3. Make an API request on the server side for that data
4. Do a second mount of the React app, this time feeding it both routing data and the API results

For some context, `next.js` uses a custom lifecycle method [`getInitialProps`](https://nextjs.org/learn/basics/fetching-data-for-pages) to handle the data collection from the first pass render.

design-documents/frontend/server-side-rendering.md
```diff
@@ -0,0 +1,129 @@
...
+   - No additional HTML needs to be rendered into containers. As a result, the user perceives faster performance.
+   - At this point, some content might appear to be interactive (e.g., buttons), but won't actually work yet.
+1. JavaScript is loaded and processed, including the React library.
+1. React [hydrates](https://reactjs.org/docs/react-dom.html#hydrate) any containers whose HTML contents were rendered by [ReactDOMServer](https://reactjs.org/docs/react-dom-server.html) in step 2 (above). Nothing needs to be re-rendered here.
+1. The page is now fully interactive.
+1. Subsequent API requests are made directly to the GraphQL API.
+
+## Requirements
+
+- A JavaScript-capable HTTP server, typically a [Node.js][] server (e.g., [Express][]).
```
DrewML: Should clarify this: you don't need a JavaScript-capable HTTP server - you just need your web server to have access to _something_ that let's it execute JavaScript. That does not need to live within the HTTP server itself. Small but important distinction

design-documents/frontend/server-side-rendering.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Implementation
+
+Some solutions provide out-of-the-box SSR, like [Next.js](https://nextjs.org/), [Razzle][] and the light-weight [Preact](https://preactjs.com/guide/v10/server-side-rendering); however, if you want more tight control of your application, you might want to set it up yourself. If you go the latter route, you would still do well to use these projects as inspiration. In fact, [Razzle][] has [a great development setup](https://github.com/jaredpalmer/razzle#how-razzle-works-the-secret-sauce).
```
DrewML: Don't think `Next.js` and `Razzle` are comparable technologies.

`Next.js` solves the whole problem of "how do I fetch my data on the server, render it to the client, then pick that back up?"

`Razzle` just runs your webpack builds for you in a way that makes it easy to work on client and server at the same time locally.

design-documents/frontend/server-side-rendering.md
DrewML: The alternative is not doing the steps mentioned above, and letting all your components render their default states without data.

But that itself has its own costs. Immediately after the first render on the client, you have to go back to the network, get your data, and then re-render the tree from the top down, which could lead to a very different UI.

design-documents/frontend/server-side-rendering.md
jedmao: Yes, but even with the data, JavaScript still needs to render it, which is what I'm saying.

design-documents/frontend/server-side-rendering.md
jedmao: I'm not concered about "for most apps" because this document is about an ideal app. There are various ways to do it, as you pointed out, but that's an implementation detail. We did something similar to Next.js at Nordstrom. The point here is that it can be done. Even if it requires 2 passes, that response is still cacheable.

That said, I'm much more interested in the [GraphQL Code Generator](https://graphql-code-generator.com) approach where you can pluck `gql` tags from the components to form a single query, [like I'm doing here](https://github.com/jedmao/tsx/blob/redux/codegen.yml). This whole topic warrants its own proposal though and I didn't want to derail the gist of this document with those implementation details.

I'm going to address the "critical content" part of your concerns though.

design-documents/frontend/server-side-rendering.md
jedmao: Yes, technically, you are correct. I really don't want to point people down that path though, so I'll just stick to my "non-golden path" and "not recommended" piece in the FAQ section.

design-documents/frontend/server-side-rendering.md
jedmao: The only comparison I made between them is that they both "provide out-of-the-box SSR." Are you claiming that this is not the case? What about [this](https://github.com/jaredpalmer/razzle/blob/25f61bc8ce90553747d2335587af653304c085c5/packages/create-razzle-app/templates/default/src/server.js#L15)?

Razzle [makes the following claim](https://github.com/jaredpalmer/razzle#readme):

> Razzle is a tool that abstracts all complex configuration needed for SSR into a single dependency.

I think that's more than "just locally." Granted, Razzle is not the most popular option, but it has some golden nuggets in its configuration.

As for data fetching, you're right. Razzle doesn't handle that, but that's intentional. I'm not really telling anyone what to use here; only, that it can and should be done.

jedmao: I'm going to add a "data fetching" section to elaborate on some of these details.

jedmao:
@DrewML, I believe I addressed all of your concerns in 9651acf.

PR #318, opened 2019-10-15
DB Information Schema - COLUMNS

jean-bernard-valentaten:
## Problem
This is a subset of #264 , covering only COLUMNS and TABLES interfaces.

Different RDBMS return Information Schema (metadata) information in different format. Magento framework should allow to switch between different implementations in order to support those different versions of DBs, as well as differences between versions.

## Solution
Introduce granular interfaces, based in Information Schema standard (looking into MySQL and MariaDB definitions in this scope, links are in the document). Application can switch between different implementation based on the profile.

### DIFF
--- a/design-documents/database/declarative-schema-database-compatibility.md
+++ b/design-documents/database/declarative-schema-database-compatibility.md
@@ -65,7 +65,7 @@ abstract class InformationSchema\Table\Column
     public function getIsNullable(): bool {}
     public function getExtra(): Extra {}
     public function getComment() {}
-    public function getDefaultValue(): string {}
+    public function getDefaultValue(): ?string {}
 }
 ```
 
@@ -198,8 +198,6 @@ It has default value of `mysql4`. I propose to add 2 values: `mysql` and `mariad
 ## Considerations
 
 `\Magento\Framework\DB\Adapter\AdapterInterface` has methods like `describeTable` which return raw values from database engine. This is leaky interface. We should have interfaces that return normalized data and are compatible with all databases. Methods that don't return normalized data need to be deprecated on `\Magento\Framework\DB\Adapter\AdapterInterface`.
-
-`engine` field in declarative schema need to be deprecated.
 
 ## Resources
 
--- a/design-documents/database/declarative-schema-database-compatibility.md
+++ b/design-documents/database/declarative-schema-database-compatibility.md
@@ -349,23 +349,6 @@ class InformationSchema\Table\CharColumn extends Column
     public function getDefaultValue(): ?string {}
 }
 ```
-```
-class InformationSchema\Table\NationalCharColumn extends Column
-{
-    /**
-     * Retturns the maximum length in characters.
-     */
-    public function getCharLength(): int {}
-
-    /**
-     * Returns the default value of the column.
-     *
-     * @return string|null
-     */
-    public function getDefaultValue(): ?string {}
-}
-```
-Note that MySQL / MariaDB maps this type to `char`.
 
 ```
 class InformationSchema\Table\VarcharColumn extends Column
@@ -387,26 +370,6 @@ class InformationSchema\Table\VarcharColumn extends Column
 ```
 
 ```
-class InformationSchema\Table\NationalVarcharColumn extends Column
-{
-    /**
-     * Returns the maximum length in characters.
-     *
-     * @return int
-     */
-    public function getCharLength(): int {}
-
-    /**
-     * Returns the default value of the column.
-     *
-     * @return string|null
-     */
-    public function getDefaultValue(): ?string {}
-}
-```
-Note that MySQL / MariaDB maps this type to `varchar`.
-
-```
 class InformationSchema\Table\ClobColumn extends Column
 {
     /**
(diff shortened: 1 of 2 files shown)


larsroettig:
@akaplya can you review it pls

larsroettig:
Kann be merged told me @akaplya

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,1093 @@
...
+     * Returns the default value of the column.
+     *
+     * @return int|null
+     */
+    public function getDefaultValue(): ?int {}
+}
+```
+
+```
+class InformationSchema\Table\VarBitColumn extends Column
```
melnikovi: Do we need to introduce types that are not used by application currently? We will have to make sure that these types support across different databases engines.

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,1093 @@
...
+
+    /**
+     * Returns the default value of the column.
+     *
+     * @return float|null
+     */
+    public function getDefaultValue(): ?float {}
+}
+```
+Note that ANSI-SQL defines the exact numerical data types `DECIMAL` and `NUMERIC` as being equivalent. MySQL / MariaDB
```
melnikovi: I would try to make columns types generic and delegate responsibility of mapping them to particular database engine to management classes.

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,1093 @@
...
+
+```
+class InformationSchema\Table\NationalClobColumn extends Column
+{
+    public const TYPE = 'NCLOB';
+
+    /**
+     * Returns the character set name.
+     *
+     * Note that not all RDBMS support this, returning `null` indicates whether this is supported.
```
melnikovi: Then we can't make type like this.

design-documents/database/declarative-schema-database-compatibility.md
```diff
@@ -0,0 +1,1093 @@
...
+### Profiles
+
+The above interfaces will have multiple implementations, and the framework should be switching between them depending on
+the RDBMS it works with.
+The framework decides which implementations to use based on declared profiles.
+The profiles are declared in `di.xml` or as a separate configuration, with the following structure:
+
+```
+information_schema_profiles:
+  mysql-5.6:
```
melnikovi: Are we going to have these values in env.php?

larsroettig:
@melnikovi in PHP we do not have **enum** design pattern to archive this we need to use a library like https://github.com/spatie/enum or https://docs.spatie.be/enum/v2/introduction/.

melnikovi:
@larsroettig I was referring to this https://www.php.net/manual/en/class.splenum.php.

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: @melnikovi No they are not necessarily needed, I just added them for the sake of being complete. Technically both `bit` and `bit varying` have been dropped as of SQL-2003, so we can stimply ommit these.

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: @melnikovi Agreed

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: @melnikovi Do you suggest to drop all non-ANSI-SQL features from column classes or do you have something else in mind?

jean-bernard-valentaten:
@melnikovi I have a couple of questions about you considerations:

> Introducing enums for collation and charset

Fine with me, but bear in mind that these vary from RDBMS to RDBMS, thus those enums would have to be located in the DBAL.

> What should be return type of getDefaultValue?

Can you elaborate what you mean exactly? Imho the return type of `getDefaultValue` should be tied to the type of column but if you want it to be a generic type, I'd suggest making it `?string`. Is that what you are trying to say?

> What should be return type of getOnUpdate?

What do you propose?

design-documents/database/declarative-schema-database-compatibility.md
melnikovi: Yes, let's omit them.

design-documents/database/declarative-schema-database-compatibility.md
melnikovi: It's ok to not include this in the design. The idea is that providers will map database specific types to generic types in application.

design-documents/database/declarative-schema-database-compatibility.md
melnikovi: Yes, let's limit column types to currently used (can check declarative schema to see what is used).

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: Ok, I removed the classes for `bit` and `bit varying` from the proposal.

design-documents/database/declarative-schema-database-compatibility.md
jean-bernard-valentaten: I added a comment in the section "Providers" stating what we have discussed.

jean-bernard-valentaten:
> Introduce enum for engine. Need to think what MySQL engines do we need to support and create enum based on that

I added that to the proposal. We may not be able to use `SplEnum` though as it is contained in a PECL package only.

> Need to remove TYPE from column classes

I removed them as requested

> Remove InformationSchema\Table\IntColumn:: getType and create different classes instead

That was already done, there are type specific classes for all ANSI-SQl variants of `integer`.

melnikovi:
Open questions will be finalized in subsequent proposals.

PR #319, opened 2019-10-15
Multi-service architecture details

paliarush:
Multi-service architecture details


melnikovi:
Have you considered splitting frontend application into services? This would allow to have ownership boundaries for the teams and should make upgrade easier.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+
+![Gateway](img/GraphQL%20BFF.png)
+
+The client (e.g. browser) is interacting with the system using GraphQL. The only component of the system exposed publicly is Storefront Gateway, which is responsible for routing, full response caching and perimeter-level authorization.
+
+Storefront Gateway routes incoming queries between PWA, GraphQL Engine and Storefront Authentication Gateway.
+
+PWA's main responsibility is to [render React application on the server-side](../frontend/server-side-rendering.md). Resulting HTML page with React application becomes cacheable on the server.
+
+GraphQL server parses GraphQL query, enriches the request (explained later) and proxies calls to the Storefront services. It also manages Customer context JWT. GraphQL schema is defined here and is extensible, query resolution is based on generic resolvers and configuration for field mapping to Storefront APIs.
```
tariqjawed83: For routing to Adobe/Magento SAAS API, lot of companies are even using GraphQL to export uniform interface and connect to other servers (routing). This will be based on use-case analysis, so we have two options

1. Either Frontend calls Storefront Gateway and Gateway provides Service Discovery to the client, and then client connect's directly to the SAAS API (specially if its Adobe/Magento hosted API). SAAS API can have different security requirements, we need to think about how we can handle this?

2. GraphQL exposes uniform interface, and routes the request to SAAS API; which means Frontend will be calling only GraphQL API. Although in this approach, we can manage the access/security in a uniform manner.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+
```
tariqjawed83: Do Magento connects to 3rd party APIs? if that is the case, we need to think about that? either here or in additional proposal

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+ - Access control list (ACL) 
+ 
+Storefront Gateway may not need to be extensible.
+
+It is important to note that services have to be redeployed after installation of extensions. It should be possible to manage ACL via admin API gateway to avoid the need of redeployment.
+
+## Open questions
+
+ 1. Should PWA be deployed in multi-service mode? Probably not because it will hurt performance and increase complexity. 
+ 1. Should SF services require customer and store scope arguments explicitly and rely on query enrichment on GraphQL Server side? Alternatively they can perform enrichment themselves, this will simplify implementation of the GraphQL Server, but may introduce additional inter-dependencies between services.
```
tariqjawed83: I believe if service can make default assumptions for those parameters its better to have that within the service; but if service requires different calls each time they are executed, its better to have those values cached at a different layer and provide those to the service. Although there will be some inter-dependencies between the services, which we cannot avoid.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+ - List of supported authentication mechanisms
+ - Access control list (ACL) 
+ 
+Storefront Gateway may not need to be extensible.
+
+It is important to note that services have to be redeployed after installation of extensions. It should be possible to manage ACL via admin API gateway to avoid the need of redeployment.
+
+## Open questions
+
+ 1. Should PWA be deployed in multi-service mode? Probably not because it will hurt performance and increase complexity.
```
tariqjawed83: We can discuss this in more details, anytime we introduce a layer in Microservices Architecture, which acts as monolithic can potentially becomes the bottle-neck for development/release cycle; we need analyze how the particular product release will impact PWA and of-course performance aspects of the application.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+
+The goal of this document is to gather requirements, understand data flow and propose architecture for multi-service Magento application.
+
+Please make sure to get familiar with [high-level service isolation vision](../service-isolation.md) before proceeding.
+
+![Gateway](img/GraphQL%20BFF.png)
+
+The client (e.g. browser) is interacting with the system using GraphQL. The only component of the system exposed publicly is Storefront Gateway, which is responsible for routing, full response caching and perimeter-level authorization.
+
+Storefront Gateway routes incoming queries between PWA, GraphQL Engine and Storefront Authentication Gateway.
```
tariqjawed83: For all internal APIs whether storefront or admin, should we need to distinguish between the two? or just say API Gateway? which could later on be the gateway for Admin APIs as well,

design-documents/service-isolation/detailed-architecture.md
buskamuza: > calls to the Storefront services
Is it important to highlight that services are "Storefront"? The document is generic, and doesn't mention that it covers SF only scenarios.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+
+GraphQL authentication gateway is an independent service for improved security purposes, although it can be merged with GraphQL server.
+
+The storefront services are communicating with each other using REST or Protobuf protocol. Storefront APIs are implemented as service contracts exposed via REST/Protobuf.
+
+## Authentication, Store and Customer contexts
+
+**Auth JWT** - authentication token, which contains encrypted customer ID and is used on GraphQL Server to authenticate the logged-in customers.
+ This token is generated by the storefront authentication server in response to the request containing valid customer credentials.
+ Customer ID is not encrypted and is visible to all parties including the client application.
```
buskamuza: Where is it not encrypted? This sentence looks opposite to "authentication token, which contains encrypted customer ID" said above.

buskamuza:
Is everything from `outdated_implementation.md` document reflected in this new one?
This one look like describes the problem from different angle.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+ - PWA
+ - Domain services
+ - List of supported authentication mechanisms
+ - Access control list (ACL) 
+ 
+Storefront Gateway may not need to be extensible.
+
+It is important to note that services have to be redeployed after installation of extensions. It should be possible to manage ACL via admin API gateway to avoid the need of redeployment.
+
+## Open questions
```
melnikovi: Should we also support replaceability? We had [this](https://github.com/magento/architecture/pull/96) proposal some time ago.

design-documents/service-isolation/detailed-architecture.md
melnikovi: Splitting PWA into multiple services would allow to have ownership boundaries for the teams and should make upgrade easier. These are the main benefits of microservice architecture.

design-documents/service-isolation/detailed-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+
+Storefront services are responsible for authorizing requests based on the Auth JWT and the list of required permissions. Actual ACL check is delegated to the the Authorization service.
+
+GraphQL authentication gateway is an independent service for improved security purposes, although it can be merged with GraphQL server.
+
+The storefront services are communicating with each other using REST or Protobuf protocol. Storefront APIs are implemented as service contracts exposed via REST/Protobuf.
+
+## Authentication, Store and Customer contexts
+
+**Auth JWT** - authentication token, which contains encrypted customer ID and is used on GraphQL Server to authenticate the logged-in customers.
```
melnikovi: I think we should pass token using cookie with HttpOnly flag (this is how PHP session cookie being passed) instead of a header. But in this case, why wouldn't we use session?

design-documents/service-isolation/detailed-architecture.md
paliarush: 1. This sounds reasonable to me. Will add as an option to the document. Are you okay with having this as an alternative option in the proposal?
2. Hiding Magento SAAS API behind GraphQL will require additional network hop. We can discuss if this is acceptable.

design-documents/service-isolation/detailed-architecture.md
paliarush: It probably does for online payments and shipment. @joni-jones could you please provide any information on this?

design-documents/service-isolation/detailed-architecture.md
paliarush: In most cases the service will not be able to make assumption about defaults. Examples are catalog and CMS functionality which depends on Store Code. Catalog also depends on customer group and customer segment.

design-documents/service-isolation/detailed-architecture.md
joni-jones: Magento is not a connector to 3rd party systems itself, at least right now. But the application has multiple integrations with different 3rd party solutions like shipping carriers, payment gateways, fraud systems, etc. In this case, each service will communicate with a 3rd party system directly as only such service has knowledge of how the 3rd API looks and how to communicate with it. Also, in most cases, a separate service will expose all needed API to API gateway to avoid direct communication between API Gateway and 3rd party integration.

design-documents/service-isolation/detailed-architecture.md
joni-jones: I believe Admin API also will be exposed because it's needed to build integration by 3rd party applications to work with Store Management application.

design-documents/service-isolation/detailed-architecture.md
joni-jones: What about other applications like mobile which don't use cookies?

melnikovi: For other applications it might be ok to use token. Problem with storing token in web applications is that it can easily stollen.

design-documents/service-isolation/detailed-architecture.md
paliarush: @jedmao what is your opinion on this?

jedmao: We would absolutely lose performance if we split it up into separate deploys, because the dependency graph would be fragmented such that webpack cannot be smart about how bundles are created.

design-documents/service-isolation/detailed-architecture.md
paliarush: It is possible to specify custom Admin base URL. In this case we will have to have two gateways, right?
If not needed on specific installation, we can combine them for simplicity.

design-documents/service-isolation/detailed-architecture.md
paliarush: Initial intent was to create proposal for storefront BFF only. 
The document is not describing any admin scenarios, only storefront.

design-documents/service-isolation/detailed-architecture.md
paliarush: Customer ID is signed but not encrypted, the first occurrence is incorrect, will fix it. Thank you

design-documents/service-isolation/detailed-architecture.md
paliarush: Replaceability should be supported based on the service API contract. Custom implementation could be provided as soon as they comply with the API.

The referenced proposal was closed as far as I can see.

design-documents/service-isolation/detailed-architecture.md
paliarush: It can be stored in the same place as cookie, please give some references which explain risks and better solutions.

design-documents/service-isolation/detailed-architecture.md
tariqjawed83: yes, I believe both options should be considered and thought out, and be available on case by case basis as per my prior experience, there will not be one size fits all.

design-documents/service-isolation/detailed-architecture.md
tariqjawed83: There are multiple patterns here, one as suggested by @joni-jones which is okay but it does spreads 3rd party integration code to multiple individual services, each handling them in different ways possibly. 

In case of observably and security, some organization routes external calls to outgoing gateway; but I believe if we use sidecar pattern / proxy like Envoy, this might solve the security and telemetry issue without a gateway.

design-documents/service-isolation/detailed-architecture.md
tariqjawed83: @jedmao @melnikovi both are valid points on their, and frankly we should not make it too complicated for the first release, but our target architecture should aim for Micro Frontends; and we need to think end to end for Microservices without introducing a monolithic frontend layer somehow.

this is interesting article 
https://medium.com/@tomsoderlund/micro-frontends-a-microservice-approach-to-front-end-web-development-f325ebdadc16

https://www.manning.com/books/micro-frontends-in-action

design-documents/service-isolation/detailed-architecture.md
tariqjawed83: yes I would suggest to combine, unless there is a solid benefit and sometimes security is that concern; but if none at the moment, then why maintaining another infrastructure layer.

design-documents/service-isolation/detailed-architecture.md
paliarush: Added section

design-documents/service-isolation/detailed-architecture.md
paliarush: Added statement about Observability

design-documents/service-isolation/detailed-architecture.md
paliarush: Will keep it as monolith at this iteration.

design-documents/service-isolation/detailed-architecture.md
paliarush: Mentioned this in the proposal.

PR #320, opened 2019-10-16
Stateless tokens

AlexMaxHorkun:
## Problem

Admin/customers API tokens are stored in DB thus forcing Magento to contact storage during authentication which is avoidable, requiring additional cron job for clean up, need extra protection

## Solution

Encode information required for authentication into tokens themselves, do not store them

## Requested Reviewers

@paliarush


design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+__Pros__:
+* Established standard
+* Existing libraries
+* Wide variety of supported encryption algorithms
+* Security concerns resolved by standard and libraries themselves
+ 
+__Cons__:
+* Will have to add new project dependency for a library and it's dependencies
+* Some libraries have [known security vulnerabilities](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/)
+* Standard requires to encode specific information into tokens which may be redundant for Magento
```
joni-jones: Could you clarify which exact required fields do you mean? RFC doesn't describe any required claims.

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+ 
+__Pros__:
+* Established standard
+* Existing libraries
+* Wide variety of supported encryption algorithms
+* Security concerns resolved by standard and libraries themselves
+ 
+__Cons__:
+* Will have to add new project dependency for a library and it's dependencies
+* Some libraries have [known security vulnerabilities](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/)
```
joni-jones: It's easy to update 3rd party library when finding a security problem in the custom implementation. Also, such tools as Github track security vulnerabilities in the packages.

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+is retrieved when authenticating users for web API via token passed in the header. Admin and customer tokens may be
+manually expired by admin users by updating `oauth_token` table data. Tokens themselves are random strings of static
+length.
+ 
+With described above circumstances in order to perform authentication Magento instances have to have access to same DB
+and perform (relatively) expensive SQL queries to DB. Te more users Magento has the more tokens it has to store
+until they expire and run an extra cron job to remove expired ones.
+ 
+ 
+### Overview of new stateless tokens implementation
```
joni-jones: More scenarios should be covered not only user authentication. We have more scenarios that require stateless tokens like 3rd party integrations, webhooks, BFF communication.

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+$userData = ['user_data' => ['user_id' => 1, 'user_type' => 3], 'issued' => time(), 'expires' => time() + (60*60)];
+$encodedData = json_encode($userData);
+$signature = $encryptor->hash($encodedData, true);
+
+return $token = 'token:' .base64_encode($encodedData) .'.' .base64_encode($signature);
+```
+ 
+__Pros__:
+* Easy to implement
+* Uses EncryptorInterface which guarantees up-to-date secure encryption
```
joni-jones: It should provide a possibility to chose JWE or JWS depending on the scenario.

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+    public function encode(AuthData $data): string;
+
+    public function decode(string $token): AuthData;
+}
+``` 
+ 
+There are 2 ways we can go about implementing this:
+ 
+#### Own implementation using EncyptorInterface
+User data returned from [the SPI](#spi-for-token-data-manager) in form of array of scalars will be serialized into string, paired with
```
joni-jones: Does this mechanism support a custom key for signature?

joni-jones:
Another proposal to support JWT tokens https://github.com/magento/architecture/pull/188.

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+
+    public function getUserData(): array;
+}
+
+
+interface TokenEncoderInterface
+{
+    public function encode(AuthData $data): string;
+
+    public function decode(string $token): AuthData;
```
joni-jones: How the verification will look like?

design-documents/stateless-api-token.md
AlexMaxHorkun: my bad, non are required but there are some reserved claims. I'll update the doc

design-documents/stateless-api-token.md
AlexMaxHorkun: it is easy, but it takes time due to our release strategy and how long it takes an issue to reach core devs

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,147 @@
...
+3rd-party developers will be able to extend this list via [the SPI](#spi-for-token-data-manager).
+ 
+After serialization this data must be signed to confirm origin. Magento instance(es) will have the same secret used
+to decode the tokens without having to validate them via DB.
+ 
+Magento will have _SPI_ that will handle encoding/decoding tokens and allow 3rd-party developers to change
+how tokens are handled:
+ 
+```php
+interface AuthData
```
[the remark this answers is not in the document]
AlexMaxHorkun: 3rd-party devs could add custom fields to userData

design-documents/stateless-api-token.md
AlexMaxHorkun: I think right now all we can do is provide API for encoding and decoding tokens. I'll add such API description to the doc

design-documents/stateless-api-token.md
AlexMaxHorkun: child interface JWTTokenEncoderInterface would allow that

design-documents/stateless-api-token.md
AlexMaxHorkun: developers would be able to replace Encryptor instance via di.xml

design-documents/stateless-api-token.md
AlexMaxHorkun: Verification will be done by an implementation of TokenEncoderInterface. Special exception will be thrown if it fails. As for the implementation itself - if we go with own implementation route then we'd just compare signatures by recreating them, otherwise we'll rely on a JWT library

design-documents/stateless-api-token.md
```diff
@@ -0,0 +1,178 @@
+## Introduce stateless web API tokens
```
kokoc: `Stateless tokens` is a confusing term by itself. In fact, the proposed token will contain a state, so it makes the naming even more confusing. My understanding is that this proposal is more about client-side sessions using JWT.

The main benefit of the solution is the possibility to avoid DB queries and save some DB capacity. However, the proposed solution doesn't eliminate them completely. Authorization part (user roles, resources, permissions) should still be located on the server side, so we are trading one DB query to CPU power spent on encryption/decryption. It's not clear if this benefit actually exists for the proposed scenario.

Regarding long-living tokens(API tokens), I clearly see benefit if we use this technology for customer sessions, because they contain more information, they are short-living(there is no need to revoke tokens, so we can get more benefit) 


The API in this proposal is highly coupled with another proposal about JWT. Probably it makes sense to merge them.


In addition to token revocation, we need to describe the rotation of secrets or asymmetric keys. This is Adobe security requirements

AlexMaxHorkun: * removed the `Stateless tokens` term
 
* there are 2 main benefits: not having to store and clean up tokens / do not compromise security by tokens being readable from DB
 
* added proposed JWT library and tried to merge API proposed in the JWT proposal with the one here
 
* asymmetric algos will be usable with JWT TokenEncoder implementations but have no practical use with JWS used for authentication - that's why I don't mention them. Rotation of secrets would be out of scope of this proposal since we use secrets for other stuff and the problem should be addressed separately

kokoc: 👍 Looks good to me. One more occurrence of "stateless tokens" in file name

AlexMaxHorkun: fixed

design-documents/decrypt-api-token.md
```diff
@@ -0,0 +1,255 @@
...
+##### API for JWT token encoder
+```php
+interface JwsTokenEncoderInterface extends TokenEncoderInterface
+{
+    /**
+     * JWS specific decoding method that returns more info regarding the token.
+     *
+     * @return JwsToken JWS DTO containing headers, payload and resolved token data.
+     */
+    public function decodeJws(string $token, JwsEncryptionDataInterface $encryption): JwsToken;
```
joni-jones: It looks strange that encoder interface has method to decode.

joni-jones: JWS has nothing in common with encryption.

AlexMaxHorkun: what is the problem with that? Makes sense to me that a class that can encode a value can decode it as well

AlexMaxHorkun: JwsSignatureDataInterface then?

joni-jones: Yes, if this class is supposed to work like management but in your case, the interface, according to its name, has a single responsibility - encode data and the additional method `decodeJws` violates it.

AlexMaxHorkun: just updated couple of class names, please take a look @joni-jones

design-documents/decrypt-api-token.md
```diff
@@ -0,0 +1,255 @@
...
+interface TokenManagerInterface
+{
+    /**
+     * Create token with encoded authentication data.
+     *
+     * @param array $data Array of scalars, authentication data to be encoded.
+     * @param TokenEncodingDataInterface $data
+     * @return string Token.
+     */
+    public function encode(array $data, TokenEncodingDataInterface $data): string;
```
joni-jones: Do you have a prototype for those interfaces? How the implementation of `TokenEncodingDataInterface` will look like for the client code?

AlexMaxHorkun: It's just a DTO which, depending on implementation used will contain secret key, algorithm name, claims list etc

AlexMaxHorkun:
Will be implemented by our team

PR #324, opened 2019-10-18
MC-21637: Adds proposal for refactoring of product import

jean-bernard-valentaten:
## Problem

Product import currently relies on MySQL `AUTO_INCREMENT` which leads to issues when using MySQL 8

## Solution

Either make use of a construct similar to sequences or change primary key to use UUIDs instead of a serial key

## Requested Reviewers

@paliarush
@akaplya

### DIFF
--- a/design-documents/import/refactor-product-import.md
+++ b/design-documents/import/refactor-product-import.md
@@ -1,0 +1,380 @@
+# Proposal to refactor product import in order to drop dependency on `AUTO_INCREMENT` and table status
+
+While implementing support for MySQL 8, we found out that some places make use of the `AUTO_INCREMENT` to predict
+the primary key of data rows that are being created. One of these places being the product import.
+
+In general no application should rely on `AUTO_INCREMENT` when processing data for insertion, as generation of serial ids 
+should be left to the RDBMS in order to be thread-safe. Unfortunately when importing relational data, such as products, 
+we need to be able to predict identity of main records (`catalog_product_option` and `catalog_product_link` in our case).
+Normally the entity models would take care of persisting the relational data involved but this is way too slow for big 
+amounts of data, thus the importers resort to generating plain SQL `INSERT` or `UPDATE` queries.  
+In order to improve the performance, as little SQL queries as possible should be made, thus inserting the main record 
+and selecting it from database would slow down the process. Instead the importer predicts the next available serial id 
+by fetching the table's `AUTO_INCREMENT` value one time and then incrementing it per new record.
+
+This will eventually lead to collisions if another process inserts data into the same table shortly after the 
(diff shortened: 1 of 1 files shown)


design-documents/import/refactor-product-import.md
```diff
@@ -0,0 +1,380 @@
...
+models. Altering the database and migrating from serial id to UUID without losing referential integrity on the other 
+hand will be a complex task that requires good planning and lots of testing.
+
+I'd estimate roughly 5 to 8 man-days for the job. 
+
+## Conclusion
+
+Although using UUIDs instead of serial ids would be a great option, migrating existing data could become troublesome and
+the overall performance of using a `char` field for primary key needs to be benchmarked before even attempting to use
+UUIDs. Additionally this is probably a BIC, thus it is not an option for a patch release.
```
buskamuza: It is important to describe expected (if any) BiC changes for each approach.

buskamuza:
MySQL article about UUIDs https://mysqlserverteam.com/storing-uuid-values-in-mysql-tables/

design-documents/import/refactor-product-import.md
```diff
@@ -0,0 +1,380 @@
...
+
+Switching to sequences on the other hand brings stability and thread-safety at minimal cost. The changes that need to be 
+done are straight forward and will definitely not raise a BIC.
+
+Considering all aspects, I'd suggest switching to generic sequences. In the long run I'd suggest to support sequences in
+the declarative schema and let the database abstraction layer decide how these are implemented for arbitrary RDBMS (e.g. 
+native sequences in Oracle, PostgreSQL and MSSQL, generic sequence table in MySQL / MariaDB etc.). The `identity` 
+attribute of columns in the declarative schema could then be used to indicate which column is fed by the sequence and
```
tariqjawed83: For distributed systems, it will not be possible to scale this kind of solution, where for generating ids we are relying on the database itself; we need to control at the program level how we want to generate the IDs. Twitter was using similar sequence for MySQL using AUTO_INCREMENT; but they realized it was not possible to scale this approach.

I would suggest looking into Twitter Snowflake 
https://github.com/twitter-archive/snowflake/tree/snowflake-2010

Another article around this subject is this
https://www.callicoder.com/distributed-unique-id-sequence-number-generator/

Although currently many of Magento customers running on single-node database, but in the cloud we would need to support large scale bulk operations in distributed style, in which case we would be able to allow multiple threads pushing the data; so an optimized service that can handle this scale would be better approach; which doesn't rely on hitting database for Ids.

Snowflake uses 8 byte unsigned integers, currently Magento using 4 byte unsigned integer, not-sure whether it will be BIC? but the idea is to centralize the ID generation and take it outside of database itself; the service should be similarly scalable as other services to support the load.

PR #327, opened 2019-10-22
Enums

melnikovi:
## Problem

There is a need to restrict values of a parameter or return value of the method.

## Solution

This is commonly solved with [enumerate types](https://en.wikipedia.org/wiki/Enumerated_type). Introduction of the enums required in order to add [support of different databases](https://github.com/magento/architecture/pull/318).

### DIFF
--- a/design-documents/enums.md
+++ b/design-documents/enums.md
@@ -1,0 +1,31 @@
+# Enums
+
+## Overview
+
+There is a need to restrict values of a parameter or return value of the method. This is commonly solved with [enumerate types](https://en.wikipedia.org/wiki/Enumerated_type). Introduction of the enums required in order to add [support of different databases](https://github.com/magento/architecture/pull/318).
+
+## Design
+
+Using enums gives the following benefits:
+* You can specify list of possible values `function setEngine(EngineEnum $engine) {`
+* You can get list of all possible values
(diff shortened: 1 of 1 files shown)


design-documents/enums.md
```diff
@@ -0,0 +1,31 @@
...
+There is a need to restrict values of a parameter or return value of the method. This is commonly solved with [enumerate types](https://en.wikipedia.org/wiki/Enumerated_type). Introduction of the enums required in order to add [support of different databases](https://github.com/magento/architecture/pull/318).
+
+## Design
+
+Using enums gives the following benefits:
+* You can specify list of possible values `function setEngine(EngineEnum $engine) {`
+* You can get list of all possible values
+* You can prohibit adding new values by making enum class final
+
+There is native PHP extension [SplEnum](http://www.php.net/manual/en/class.splenum.php). But it's not bundled with PHP by default and adding requirement on it means requiring to [install dependency](https://www.php.net/manual/en/spl-types.installation.php).
```
jean-bernard-valentaten: Additional information: `SplEnum` is not available for PHP 7 and up (see: https://stackoverflow.com/questions/45911217/install-spl-types-on-php-7-1-compiled-from-source)

PR #329, opened 2019-10-23
Discount itemization - Database Changes

prabhuram93:
## Problem

The individual itemized break down of the discounts applied at cart and cart item level are to be computed and persisted on every order.

## Solution

Existing Implementation

The itemized discounts are computed and stored with extension attributes on Magento\Quote\Api\Data\CartItemInterface Magento\Quote\Api\Data\AddressInterface. These are served via GraphQL cart query. The eventual end goal is to persist this data on every order.

```
/**
 * Rule discount Interface
 */
interface RuleDiscountInterface
{
    /**
     * Get Discount Data
     *
     * @return \Magento\SalesRule\Api\Data\DiscountDataInterface
     */
    public function getDiscountData();

    /**
     * Get Rule Label
     *
     * @return string
     */
    public function getRuleLabel();

    /**
     * Get Rule ID
     *
     * @return int
     */
    public function getRuleID();
}
```

```
 <extension_attributes for="Magento\Quote\Api\Data\CartItemInterface">
        <attribute code="discounts" type="Magento\SalesRule\Api\Data\RuleDiscountInterface[]" />
    </extension_attributes>
    <extension_attributes for="Magento\Quote\Api\Data\AddressInterface">
        <attribute code="discounts" type="Magento\SalesRule\Api\Data\RuleDiscountInterface[]" />
    </extension_attributes>
```

**Schema changes**

The existing discount related metadata are being stored in quote_address, quote_item, sales_order, sales_order_item. But these tables tend to get large and have millions of records, and is decided its not ideal to alter heavyweight tables. So new tables have to be created to hold discounts data.

Also have to be cognizant about avoiding referencing between sales and quote tables.

The data will be stored as serialized json to support extensibility.

quote_totals
```
 <table name="quote_attributes" resource="checkout" engine="innodb" comment="Quote Attributes Document">
        <column xsi:type="int" name="quote_id" padding="10" unsigned="true" nullable="false" identity="false"
                comment="Quote ID"/>
        <column xsi:type="text" name="value" nullable="true" comment="Serialized quote attributes"/>
        <constraint xsi:type="primary" referenceId="PRIMARY">
            <column name="quote_id"/>
        </constraint>
        <constraint xsi:type="foreign" referenceId="QUOTE_ATTRIBUTES_QUOTE_ID_QUOTE_ENTITY_ID"    table="quote_attributes" column="quote_id" referenceTable="quote" referenceColumn="entity_id" onDelete="CASCADE"/>
        <index referenceId="QUOTE_ATTRIBUTES_QUOTE_ID" indexType="btree">
            <column name="quote_id"/>
        </index>
 </table>
```
sales_order_totals
```
 <table name="sales_order_attributes" resource="checkout" engine="innodb" comment="Sales Order Attributes Document">
        <column xsi:type="int" name="order_id" padding="10" unsigned="true" nullable="false" identity="false"
                comment="Order ID"/>
        <column xsi:type="text" name="value" nullable="true" comment="Serialized order attributes"/>
        <constraint xsi:type="primary" referenceId="PRIMARY">
            <column name=order_id"/>
        </constraint>
        <constraint xsi:type="foreign" referenceId="SALES_ORDER_ATTRIBUTES_ORDER_ID_SALES_ORDER_ENTITY_ID"   table="sales_order_attributes" column="order_id" referenceTable="sales_order" referenceColumn="entity_id" onDelete="CASCADE"/>
        <index referenceId="SALES_ORDER_ATTRIBUTES_ORDER_ID" indexType="btree">
            <column name="order_id"/>
        </index>
 </table>
```
```
/**
 * Quote Attribute Interface
 */
interface QuoteAttributeInterface
{
    /**
     * Get Quote Id
     *
     * @return int
     */
    public function getQuoteId();

    /**
     * Get Value
     *
     * @return string
     */
    public function getValue();
}
```
**Benefits**

Need to modify current table structure to incorporate this feature. Will support extensibility in the future to store additional data pertaining to totals, without modifying the table.

## Requested Reviewers
@joni-jones @kokoc @paliarush @akaplya

### DIFF
--- a/design-documents/cart-discount-itemization.md
+++ b/design-documents/cart-discount-itemization.md
@@ -26,13 +26,13 @@ The data will be stored as serialized json to support extensibility.
  <table name="quote_address_totals" resource="checkout" engine="innodb" comment="Quote Address Totals">
         <column xsi:type="int" name="address_id" padding="10" unsigned="true" nullable="false" identity="false"
                 comment="Address ID"/>
-        <column xsi:type="text" name="totals" nullable="true" comment="Collected item totals"/>
+        <column xsi:type="text" name="totals" nullable="true" comment="Serialized address totals"/>
         <constraint xsi:type="primary" referenceId="PRIMARY">
             <column name="address_id"/>
         </constraint>
         <constraint xsi:type="foreign" referenceId="QUOTE_ADDRESS_TOTALS_ADDRESS_ID_QUOTE_ADDRESS_ADDRESS_ID    table="quote_item_totals" column="address_id" referenceTable="quote_address" referenceColumn="address_id" onDelete="CASCADE"/>
         <index referenceId="QUOTE_ADDRESS_TOTALS_ADDRESS_ID" indexType="btree">
-            <column name="item_id"/>
+            <column name="address_id"/>
         </index>
  </table>
 ```
@@ -42,7 +42,7 @@ The data will be stored as serialized json to support extensibility.
  <table name="quote_item_totals" resource="checkout" engine="innodb" comment="Quote Item Totals">
         <column xsi:type="int" name="item_id" padding="10" unsigned="true" nullable="false" identity="false"
                 comment="Item ID"/>
-        <column xsi:type="text" name="totals" nullable="true" comment="Collected item totals"/>
+        <column xsi:type="text" name="totals" nullable="true" comment="Serialized item totals"/>
         <constraint xsi:type="primary" referenceId="PRIMARY">
             <column name="item_id"/>
         </constraint>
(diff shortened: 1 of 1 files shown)


cherdman:
@joni-jones @kokoc @paliarush @akaplya @buskamuza 

Approved???

design-documents/cart-discount-itemization.md
```diff
@@ -0,0 +1,58 @@
...
+                comment="Quote ID"/>
+        <column xsi:type="text" name="value" nullable="true" comment="Serialized quote totals"/>
+        <constraint xsi:type="primary" referenceId="PRIMARY">
+            <column name="entity_id"/>
+        </constraint>
+        <constraint xsi:type="foreign" referenceId="QUOTE_TOTALS_ENITITY_ID_QUOTE_ENTITY_ID"    table="quote_totals" column="entity_id" referenceTable="quote" referenceColumn="entity_id" onDelete="CASCADE"/>
+        <index referenceId="QUOTE_TOTALS_ENTITY_ID" indexType="btree">
+            <column name="entity_id"/>
+        </index>
+ </table>
```
akaplya: how will be stored item and address levels?

design-documents/cart-discount-itemization.md
```diff
@@ -0,0 +1,58 @@
...
+
+Overview
+
+The individual itemized break down of the discounts applied at cart and cart item level are to be computed and persisted on every order.
+
+**Existing Implementation**
+
+The itemized discounts are computed and stored with extension attributes on 
+Magento\Quote\Api\Data\CartItemInterface 
+Magento\Quote\Api\Data\AddressInterface.
```
akaplya: can you please specify an interface of extension?

design-documents/cart-discount-itemization.md
```diff
@@ -0,0 +1,58 @@
...
+**Schema changes**
+
+The existing discount related metadata are being stored in quote_address, quote_item, sales_order, sales_order_item. But these tables tend to get large and have millions of records,
+and is decided its not ideal to alter heavyweight tables. So new tables have to be created to hold discounts data.
+
+Also have to be congizant about avoiding referencing between sales and quote tables.
+
+The data will be stored as serialized json to support extensibility.
+
+*quote_totals*
```
kokoc: Please pick appropriate words for table name and descriptions. The goal is to save `discounts`, but the table name and descriptions are about `totals`.
Moreover, new tables introduce schemaless storage for read-only attributes without filtering capabilities. Thus, probably it's better to give a very common name for the table and use it for all types of read-only data(not only discounts). In this case, we need to create appropriate API and describe it in document

design-documents/cart-discount-itemization.md
prabhuram93: Updated the doc.

design-documents/cart-discount-itemization.md
prabhuram93: Updated the doc with common table names.

fooman:
What data would actually be stored in the text field? The complete rule data for each sales rule? That could end up growing rather quickly. In that case we might be better off having another table like salesrule_revision_history (or make salesrule immutable and hide replaced versions) which we can reference by id.

Alternatively if we are not looking to recalculate the rule we should be able to only save the calculated results like actual discount and tax amounts and labels.

PR #330, opened 2019-10-23
Asynchronous Import 

naydav:
-- Intro
-- REST API
-- Modularity (API / Extension points)


design-documents/asynchronous-import/rest-api.md.md
```diff
@@ -0,0 +1,85 @@
...
+{
+  "source": {
+    "sourceType": "http"
+    "sourceDefinition": "http://some.domain/file.csv"
+    "sourceDataFormat": "CSV"
+  }
+  "format': {
+    "escape": "|"
+    "enclosure" : "|"
+    "delimiter" => "|"
```
nuzil: Pls. replace => with :

design-documents/asynchronous-import/rest-api.md.md
```diff
@@ -0,0 +1,85 @@
...
+- `convertingRules[]` is **optional**. Describes how to change data before Import process. Additional data could be placed in `parameters`. 
+- `import` is **required**. Describes how to import data.
+  - `importType` is **required**. Example: `stock|advanced_pricing|product`.
+  - `importBehaviour` is **required**. Example: `add|delete`
+
+**Result:**
+
+```
+{
+  "UUID": "uuid_string"
```
nuzil: "UUID" as Lowercase?

design-documents/asynchronous-import/asycnhronous-import.md
```diff
@@ -0,0 +1,43 @@
...
+
+[MVP Board](https://app.zenhub.com/workspaces/async-import-5b5f349bd6768d6255917727/reports/burndown?milestoneId=4625823)
+
+- Sources data retrieving: *HTTP, HTTPS, base64*;
+- CSV reader;
+- Data Converting Rules;
+- Import Data Exchanging;
+- Import configuration;
+- Product import;
+- Stock status import;
```
nuzil: I'm a bit confused with stock status. You mean Stock QTY and not only enabled/disabled? 
Also how about differentiate Single Stock and MSI source/stock?

ticaje:
After digging a little bit into this issue, the following assesment has come out:

It turns out that the class RetrieveSourceData contains the source validator actor that wraps all the possible validator to a specific source data retriever(strategy), this means that each strategy must comply with the validations defined by di.xml:

image

This forces me to assume a sort of data contract where a new strategy(json's in this case), must fit such a data contract. Not empty: sourceType, sourceDefinition, sourceDataFormat; and also sourceDefinition must come in valid base64 format.

This way, coming down to Json strategy, i understand that sourceDefinition field is a json encoded in base64, if this is ok then i assume that it is actually a csv translated into json, so the data strategy retriever takes this chunk and translate it into array iterator as if a magento standard csv was.

design-documents/asynchronous-import/asycnhronous-import.md
```diff
@@ -0,0 +1,43 @@
...
+- Advanced pricing import;
+- Documentation;
+
+## TODO
```
nuzil: Lets change a Priorities here

- Design of `Import configuration`;
- Design of `Get Import status`;
- Design of `Import processing - Sync / Async`
- Design of `Restart failed operations`;

design-documents/asynchronous-import/modularity/import-csv.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+```php
+/**
+ * Describes how to parse data
+ *
+ * @api
+ */
+interface CsvFormatInterface extends ExtensibleDataInterface
```
tariqjawed83: Isn't this a duplicate, you already defined this interface above?

design-documents/asynchronous-import/modularity/source-data-retrieving.md
```diff
@@ -0,0 +1,153 @@
...
+
+use Magento\AsynchronousImportSourceDataRetrievingApi\Api\Data\SourceInterface;
+use Magento\Framework\Validation\ValidationResult;
+
+/**
+ * Extension point for adding source validators
+ *
+ * @api
+ */
+interface SourceValidatorInterface
```
tariqjawed83: How & where we will going to use this extension point?

design-documents/asynchronous-import/asycnhronous-import.md
```diff
@@ -0,0 +1,43 @@
...
+- Parse source data;
+- Apply data converting rule;
+- Exchange data with Magento instance via Bulk API Service contracts / Bulk API REST (partial processing);
+- Later an user can request balk status of import and resubmit objects which were failed during the import;
+
+## REST API
+
+- [REST API](rest-api.md)
+
+## Modularity (API / Extension points)
```
tariqjawed83: 1- It will be helpful to add a quick diagram to show the flow of interaction between the modules!
2- A simple class / interface UML diagram will make it really easy to grasp the concepts

design-documents/asynchronous-import/modularity/source-data-retrieving.md
```diff
@@ -0,0 +1,153 @@
...
+namespace Magento\AsynchronousImportSourceDataRetrievingApi\Api\Data;
+
+/**
+ * Represents retrieved source data (result of retrieving operation)
+ *
+ * @api
+ */
+interface SourceDataInterface extends \IteratorAggregate
+{
+    public const ITERATOR = 'iterator';
```
tariqjawed83: how we use this?

design-documents/asynchronous-import/modularity/data-exchanging.md
```diff
@@ -0,0 +1,211 @@
...
+ * @api
+ */
+interface ImportInterface extends ExtensibleDataInterface
+{
+    /**
+     * Get import uuid
+     *
+     * @return string|null
+     */
+    public function getUuid(): ?string;
```
tariqjawed83: do we need to generalize UUID? can't we just say ID. The problem with UUID is that they are big in size and doesn't index well. When your data grows in the table, the index size also grows with it, which leads to query performance hit.

https://www.callicoder.com/distributed-unique-id-sequence-number-generator/

design-documents/asynchronous-import/modularity/data-converting.md
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * @api
+ */
+interface ApplyConvertingRulesInterface
+{
+    /**
+     * Apply converting rules to import data operation. Uses differect strategies for rules applying
+     *
+     * @param array $importData
+     * @param ConvertingRuleInterface[] $convertingRules
```
tariqjawed83: some definition of convertingRules or explanation should be there...

design-documents/asynchronous-import/modularity/data-converting.md
```diff
@@ -0,0 +1,146 @@
...
+    ): array;
+}
+```
+
+```php
+namespace Magento\AsynchronousImportDataConvertingApi\Api\Data;
+
+/**
+ * Describes how to change data before import
+ *
```
tariqjawed83: please add some example use or bit more explanation of this

design-documents/asynchronous-import/modularity/source-data-retrieving.md
nuzil: It will be used: 
https://github.com/magento/architecture/pull/330/files#diff-610ea2fd7eb8081598349a79459e9313R40
On a step we are retrieving data to validate them

design-documents/asynchronous-import/modularity/data-exchanging.md
nuzil: I do not expect really big data flow where, even if you are doing daily imports with 100+ files daily (which is not common case), we would need to have some cleaning functionality for such table. 
to use UUID was provided in Asynchronous operations and as I know idea of Arch team was to go into UUID direction and discussion is still open, or?

design-documents/asynchronous-import/modularity/data-converting.md
```diff
@@ -0,0 +1,146 @@
+# Data converting before import
```
tariqjawed83: I also would like to recommend that please don't implement your own ETL code, there are many good PHP libraries available that can do all this for you. Please have a look into those

https://php-etl.readthedocs.io/en/v1/
https://blog.panoply.io/6-of-the-best-php-etl-tools

design-documents/asynchronous-import/modularity/data-exchanging.md
tariqjawed83: there some other PRs currently opened to replace DB Sequence for IDs with some other strategy, when the ID is not externally provided; where we are having some discussions!

design-documents/asynchronous-import/modularity/data-exchanging.md
```diff
@@ -0,0 +1,211 @@
...
+     * @return string
+     */
+    public function getImportType(): string;
+
+    /**
+     * Get import behaviour
+     *
+     * @return string
+     */
+    public function getImportBehaviour(): string;
```
tariqjawed83: What is this behavior? and why it is string... trying to understand, if there could be more better way to represent it rather than String.

design-documents/asynchronous-import/modularity/source-data-retrieving.md
tariqjawed83: is it injected through DI configuration?

nuzil: right

design-documents/asynchronous-import/modularity/data-converting.md
nuzil: Thanks for info, I will check it!

nuzil: @tariqjawed83, I expect that its will take a lot of time to confirm from Architecture team of usage some library in Magento Core?

design-documents/asynchronous-import/modularity/data-exchanging.md
nuzil: Behaviour here is an "add", "update", "delete", "replace" ....

nuzil: So on import process we can add new objects, update existed, delete some or replace.
We will start with "add" and "delete" with MVP

nuzil:
Related to https://github.com/magento/architecture/pull/330#discussion_r344402912

AsyncImportUML

Its a high level diagram, hope will help a bit

PR #332, opened 2019-10-24
Introduce GraphQL Schema Development doc

jedmao:
## Problem

If we decide to move forward building a GraphQL API in Node.js, we should consider the developer experience.

## Solution

I think we have the luxury of holding off until Prisma releases some exciting tools that we would do well to leverage. They are fully dedicated to improving the developer experience.

### DIFF
--- a/design-documents/graph-ql/schema-development.md
+++ b/design-documents/graph-ql/schema-development.md
@@ -25,18 +25,16 @@ In this document, we will explore the code-first approach.
 
 ### Pros
 
-- SDL is a generated artifact, not manually maintained.
 - The code-first schema is the single source of truth.
+   - SDL is a generated artifact and not manually maintained.
    - No inconsistencies. Schema definition & resolvers are always in sync.
    - No redundancy (duplication).
    - Easier to maintain.
+   - Scalable.
 - The shape of the API is closer to the data model.
 - Better IDE support & developer experience.
 - Better code reuse and type safety.
-- Can split by features/modules.
 - Terse (less boilerplate).
-- Declarative.
-- Scalable.
 
 ### Cons
 
--- a/design-documents/graph-ql/schema-development.md
+++ b/design-documents/graph-ql/schema-development.md
@@ -1,0 +1,108 @@
+# [GraphQL][] Schema Development
+
+Before developing a [GraphQL schema](https://graphql.org/learn/schema/), we need to consider the following:
+
+- [GraphQL Ecosystem](#graphql-ecosystem)
+- [GraphQL Server](#graphql-server)
+- [Code-First vs. SDL](#code-first-vs-sdl)
+- [Code-First Frameworks](#code-first-frameworks)
+   - [TypeGraphQL](#typegraphql)
+   - [Nexus](#nexus)
+
+## GraphQL Ecosystem
+
+[GraphQL][] is [a specification](https://graphql.github.io/graphql-spec/) for which a good handful of implementations are written across various languages (e.g., [Rust](https://github.com/graphql-rust), [PHP](https://github.com/webonyx/graphql-php), [Python](https://graphene-python.org/), [.NET](https://github.com/graphql-dotnet/graphql-dotnet) and [JavaScript](https://www.npmjs.com/package/graphql)). There is no wrong choice between these implementations; however, when considering the ecosystem, there is a clear winner.
+
+In 2015, the [GraphQL][] spec was [first published](https://graphql.github.io/graphql-spec/July2015/) along with the reference implementation in JavaScript. It should be no surprise then, that today, JavaScript holds the strongest ecosystem of [GraphQL][] libraries, with [4,833 related npm packages](https://www.npmjs.com/search?q=graphql) at the time of this writing (10/2019). As such, the JavaScript ecosystem is a solid and safe choice, moving forward.
+
+## GraphQL Server
+
+Within the JavaScript ecosystem, there are [a few server choices available](https://www.npmtrends.com/apollo-server-vs-koa-graphql-vs-prisma), but the most popular and mature choice is [Apollo Server][]. Optionally, [`graphql-yoga`](https://www.npmjs.com/search?q=graphql-yoga) uses [Apollo Server][] under the hood.
+
+## Code-First vs. SDL
+
+As [previously mentioned](#graphql-ecosystem), there are myriad JavaScript libraries available to assist in the creation of a [GraphQL][] server, but before we explore them, we should consider the pros & cons of defining our schema as code-first vs. SDL. Many articles [have been written](#references) about this topic and there are excellent arguments on both sides, but code-first appears to be the direction things are headed.
+
+In this document, we will explore the code-first approach.
+
+> Instead of increasing the complexity of GraphQL server development with more tools, we should strive for a simpler development model. Ideally, one that lets developers leverage the programming language they're already using – [this is the idea of code-first](https://www.prisma.io/blog/the-problems-of-schema-first-graphql-development-x1mn4cb0tyl3?source=post_page-----cf0e50d5ccff----------------------#code-first-a-language-idiomatic-way-for-graphql-server-development).
+
+### Pros
+
+- SDL is a generated artifact, not manually maintained.
+- The code-first schema is the single source of truth.
+   - No inconsistencies. Schema definition & resolvers are always in sync.
+   - No redundancy (duplication).
+   - Easier to maintain.
+- The shape of the API is closer to the data model.
+- Better IDE support & developer experience.
+- Better code reuse and type safety.
+- Can split by features/modules.
+- Terse (less boilerplate).
+- Declarative.
+- Scalable.
+
+### Cons
+
+- Some of [the latest code-first libraries](https://www.prisma.io/blog/introducing-graphql-nexus-code-first-graphql-server-development-ll6s1yy5cxl5) are not yet mature enough.
+- Tightly coupled to a specific language (e.g., [TypeScript][]).
+
+## Code-First Frameworks
+
+Both [TypeGraphQL](#typegraphql) and [Nexus](#nexus) are the key players that provide a code-first solution, but they go about it in entirely different ways. One thing they have in common, however, is that they are designed to use [TypeScript][].
+
+_At this point, we're talking about developing a [GraphQL][] API in [Node.js][] with [TypeScript][]._
+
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/schema-development.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Code-First vs. SDL
+
+As [previously mentioned](#graphql-ecosystem), there are myriad JavaScript libraries available to assist in the creation of a [GraphQL][] server, but before we explore them, we should consider the pros & cons of defining our schema as code-first vs. SDL. Many articles [have been written](#references) about this topic and there are excellent arguments on both sides, but code-first appears to be the direction things are headed.
+
+In this document, we will explore the code-first approach.
+
+> Instead of increasing the complexity of GraphQL server development with more tools, we should strive for a simpler development model. Ideally, one that lets developers leverage the programming language they're already using – [this is the idea of code-first](https://www.prisma.io/blog/the-problems-of-schema-first-graphql-development-x1mn4cb0tyl3?source=post_page-----cf0e50d5ccff----------------------#code-first-a-language-idiomatic-way-for-graphql-server-development).
+
+### Pros
```
DrewML: I think a lot of these pros are missing information.

> SDL is a generated artifact, not manually maintained.

Can you explain what benefits this provides?

> The shape of the API is closer to the data model

Is this a pro? One of the benefits of GraphQL is that it's easier to not leak your backend's architecture into client-side code.

> Better code reuse and type safety

Probably needs some examples. SDL describes the types, so it's not clear why this could not be used to generate TypeScript (or other) types

> Declarative

SDL looks pretty declarative to me 🤔. Can you explain this?

> Can split by features/modules

Example here would be helpful, how this can be done with code-first but not with SDL

> Scalable

Seems like this is a relatively new way of doing things, might be a bit premature to throw that label on 😄

design-documents/graph-ql/schema-development.md
```diff
@@ -0,0 +1,108 @@
...
+Both [TypeGraphQL](#typegraphql) and [Nexus](#nexus) are the key players that provide a code-first solution, but they go about it in entirely different ways. One thing they have in common, however, is that they are designed to use [TypeScript][].
+
+_At this point, we're talking about developing a [GraphQL][] API in [Node.js][] with [TypeScript][]._
+
+### [TypeGraphQL][]
+
+> [TypeGraphQL][] is a library that makes this process enjoyable by defining the schema using only classes and a bit of decorator magic.
+
+```ts
+@ObjectType()
```
DrewML: Can we change this example to one without decorators for a more accurate picture of how we'd be writing code. Decorators have been through _massive_, _massive_ churn in TC39, and are only at Stage 2. 

Relevant: https://github.com/tc39/proposal-decorators#how-should-i-use-decorators-in-transpilers-today

design-documents/graph-ql/schema-development.md
```diff
@@ -0,0 +1,108 @@
...
+class Recipe {
+  @Field()
+  title: string;
+
+  @Field(type => [Rate])
```
paliarush: Is there a reason why the type cannot be deducted in this case?

design-documents/graph-ql/schema-development.md
jedmao: This is the #1 reason I'm not happy about TypeGraphQL, but it's the only great option (today) that I could find. I'm mostly holding out for whatever Prisma releases, but that's a few months away.

I don't think we can change the example. It comes straight from TypeGraphQL's documentation and you know it's going to be ugly without decorators.

design-documents/graph-ql/schema-development.md
jedmao: I agree that this is a weakness of TypeGraphQL's impementation. This is their example, pulled directly from their site, but I'm positive this can be improved. I mean, it _can_ be, but at what cost? We could intercept the AST at compile time to infer these types. There might even be open issues about this.

design-documents/graph-ql/schema-development.md
jedmao: I was trying to keep the list simple, but apparently that's too confusing. I'll try to add some clarity to these points.

jedmao: Before I do, though, let me see if you like these responses:

1. The main benefit is that you define your schema & resolvers in the same place (in code) instead of going back & forth between SDL & code (i.e., resolvers, etc.).
1. I don't see how this would be leaking anything into client-side code, as all of this "code" is written on the back end. Are you suggesting the SDL would be developed on the front end?
1. You _can_ generate TypeScript from SDL via [GraphQL Codegen](https://graphql-code-generator.com/), but because you can't define resolvers in SDL, you still have to bounce between SDL & code to complete the models.
1. SDL is definitely declarative. I realize now that this "Pros" list implies things that SDL are not. That was not my intent, but I can remove this bullet point to make that more clear?
1. Again, didn't mean to imply any of these pros could not also be done with SDL. I don't see any reason you couldn't also do this with SDL. Do you have any insights here?
1. It's scalable in that it's easier to maintain without introducing inconsistencies between SDL & code. I don't think that's premature at all.

design-documents/graph-ql/schema-development.md
DrewML: > I don't think we can change the example. It comes straight from TypeGraphQL's documentation and you know it's going to be ugly without decorators.

Right. In that case it's just not a viable solution for Magento. We _can't_ release rely on stage 2 ECMA proposals - it's a dangerous gamble.

jedmao:
@DrewML I removed the TypeGraphQL section entirely. Please see e7783ee.

fooman:
Under Cons I would expect a mention of what happens to all the work that went into creating GraphQl modules for M2 in php. It is not at all clear what is to be expected to happen with the existing implementation.

There is even tooling out there that creates M2 GraphQl endpoints
https://mage2gen.com/

Vinai:
To me it sounds like it's way to early to pick a candidate. Basically, whatever is chosen will end up being the new knockout.

DrewML:
> Under Cons I would expect a mention of what happens to all the work that went into creating GraphQl modules for M2 in php.

Yep, this is definitely missing atm, but important.

> To me it sounds like it's way to early to pick a candidate. Basically, whatever is chosen will end up being the new knockout.

It is _very_ important to me that we don't back ourselves into that corner again :D

-----

FWIW, I know this PR is missing a ton of context, due to some internal discussions we've had. To help frame the discussion, this is _super_ early exploratory work, but there are no concrete plans to ditch the current API. There are some concerns around the long-term health of the Webonyx PHP GraphQL library though, which sparked this exploratory work.

jedmao:
It's definitely too early to pick a candidate. That's why the conclusion was written such that "if you need something now then X". This document is not declaring our commitment to anything at the moment, but rather investigating what options we have and presenting the findings.

design-documents/graph-ql/schema-development.md
```diff
@@ -0,0 +1,85 @@
...
+
+In 2015, the [GraphQL][] spec was [first published](https://graphql.github.io/graphql-spec/July2015/) along with the reference implementation in JavaScript. It should be no surprise then, that today, JavaScript holds the strongest ecosystem of [GraphQL][] libraries, with [4,833 related npm packages](https://www.npmjs.com/search?q=graphql) at the time of this writing (10/2019). As such, the JavaScript ecosystem is a solid and safe choice, moving forward.
+
+## GraphQL Server
+
+Within the JavaScript ecosystem, there are [a few server choices available](https://www.npmtrends.com/apollo-server-vs-koa-graphql-vs-prisma), but the most popular and mature choice is [Apollo Server][]. Optionally, [`graphql-yoga`](https://www.npmjs.com/search?q=graphql-yoga) uses [Apollo Server][] under the hood.
```
paales: - graphql-yoga should be deprecated, it's based on Apollo Server 1.0. The things it does are already in Apollo Server 2.0
- Prisma should be considered as an ORM with deep integrations with the frontend API, has nothing to do with the actual server (especially with Prisma 2.0)

design-documents/graph-ql/schema-development.md
```diff
@@ -0,0 +1,85 @@
...
+- Better IDE support & developer experience.
+- Better code reuse and type safety.
+- Can split by features/modules.
+- Terse (less boilerplate).
+- Declarative.
+- Scalable.
+
+### Cons
+
+- Some of [the latest code-first libraries](https://www.prisma.io/blog/introducing-graphql-nexus-code-first-graphql-server-development-ll6s1yy5cxl5) are not yet mature enough.
```
paales: We find Nexus to actually be quite stable, but the issue seems to be in prisma1/2 being unstable / missing features.

jedmao: Thanks for the feedback. Prisma definitely needs more work.

jedmao: I've requested for Prisma to download and install Magento so that they can add the Magento database schema to their internal list of schemas for testing, which is exactly what they've done. In fact, they already had a pretty large list as it was, so this was just one more.

The bottom line is that Prisma is determined to fix these kinks and add more MVP features before a GA release, hopefully in Q1 2020. I do think we should wait for that, but I'm also very optimistic about what they intend to deliver, from what I've already seen.

design-documents/graph-ql/schema-development.md
jedmao: See https://github.com/magento/architecture/pull/332/commits/f3fb8d0bc26d11380d4855c0b895ebcc0151dcce

design-documents/graph-ql/schema-development.md
jedmao: - I nested the "SDL is a generated artifact" point under the single source of truth bullet point to emphasise why it's a pro.
- I also nested "Scalable" under the same bullet point.
- Removed the "can split..." bullet.
- Removed "declarative".

See https://github.com/magento/architecture/pull/332/commits/7419e6e5ddf81b47e2bd6eb087a6a4c725469cc2

jedmao: - I nested the "SDL is a generated artifact" point under the single source of truth bullet point to emphasise why it's a pro.
- I also nested "Scalable" under the same bullet point.
- Removed the "can split..." bullet.
- Removed "declarative".

See https://github.com/magento/architecture/pull/332/commits/7419e6e5ddf81b47e2bd6eb087a6a4c725469cc2

PR #333, opened 2019-10-25
MC-22136: UrlResolver identifier

cpartica:
## Problem

In reference to usage of Id in product search that became deprecated, we need an identifier to replace the ID as Int in urlResolver. This can be any unique identifier for Product, Category and CMS that can be later used to query the actual entity.

## Solution

Deprecate EntityUrl.id
Add new field to EntityUrl called identifier
Return sku for EntityUrl.identifier on product, and match id for other entities

This is the Architectural proposal needed for changes

Other idea is to add filter by id back to product filter, because we did not keep backward compatibility by removing it, problem is that Id it's not part of search index right now, and we don't want it to be.

### DIFF
--- a/design-documents/graph-ql/coverage/url-resolver-identifier.md
+++ b/design-documents/graph-ql/coverage/url-resolver-identifier.md
@@ -47,8 +47,23 @@ This ID has to be filterable in all entities, even in product
      }
    }
  ```
-As of now this feature is not possible, but it has to be introduced to unlock this flow.
+ 
+  ```graphql
+  products(filter: { id: { in: ["ID1", "ID2", "ID3"]}}) {
+      items {
+        id
+        name
+      }
+    }
+  ```
+  
+This is not currently available in current schema, but it has to be introduced to unlock this flow.
 A separate proposal for product will be requested.
+
+ #### Intended functionality
+ 
+- If only id is passed, bypass the current Search API and fetch data directly from DB
+- If id + other filters are passed, consider this an unsupported case. To be implemented later . Throw an exception in case of such query. This case is intended to be implemented as a separate task.
 
  ## Alternatives
  
--- a/design-documents/graph-ql/coverage/url-resolver-identifier.md
+++ b/design-documents/graph-ql/coverage/url-resolver-identifier.md
@@ -3,15 +3,15 @@
  
 In the client flow, for PWA, in order to load a entity/template, one must make a query to the `urlResolver` to obtain the *id* and *Entity Type* of the content that can be Product, Category or CMS.
 
-####Example:
-#####Request:
+#### Example:
+##### Request:
 ```graphql
 urlResolver(url: "/breathe-easy-tank") {
   id
   type
 }
 ```
-#####Response:
+##### Response:
 ```json
 {
   "data": {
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/url-resolver-identifier.md
```diff
@@ -0,0 +1,68 @@
...
+     items {
+       id
+       name
+     }
+   }
+ ```
+
+ ## Alternatives
+ 
+ We can directly return the entity instead of returning the identifier from UrlResolver.
```
paliarush: Please give a link to existing architecture proposal related to this topic: https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/storefront-route.md

design-documents/graph-ql/coverage/url-resolver-identifier.md
```diff
@@ -0,0 +1,68 @@
...
+
+Currently we can't use this Id in Product to filter by, because of how Search API works, also CMS has another unique identifier that can be used.
+Also we want not to expose database autoincrement ID in graphql in the future because of replacements with UUID.
+
+ ## Proposed solution
+
+Deprecate `id` field from `urlResolver` and introduce a new field called identifier, unique to each entity that's not the Id from the database.
+
+```graphql
+urlResolver(url: String) {
```
paliarush: We may need to keep `id` and introduce `urlResolverV2` for client-side caching. We are going to have a discussion today with domain architects responsible for affected areas.

design-documents/graph-ql/coverage/url-resolver-identifier.md
cpartica: added

design-documents/graph-ql/coverage/url-resolver-identifier.md
cpartica: added a comment at the bottom part of alternatives

paliarush: Here are the final list of steps proposed (based on architectural discussion):
1. introduce “ids in” to products filter, type ids:[ID!]
1.  introduce urlResolverV2, whith return type of id:ID
     1.  id is id of the entity, casted to string
     1.  if only ids are passed, bypass Search API and fetch data directly from DB
     1. if ids + other filters are passed, consider this an unsupported case. To be implemented later (create a task). Throw an exception in case of such query.

paliarush: @cpartica in the adjusted proposal I see product filter by "id", not by "ids". Is this a mistake or there is a reason to avoid "ids"?

The following guidelines should be added to the proposal in some form, they are really important:

```
- if only ids are passed, bypass Search API and fetch data directly from DB
- if ids + other filters are passed, consider this an unsupported case. To be implemented later (create a task). Throw an exception in case of such query.
```

cpartica: added those lines

design-documents/graph-ql/coverage/url-resolver-identifier.md
```diff
@@ -0,0 +1,109 @@
...
+urlResolverV2(url: String) {
+  id: ID!
+  ...
+}
+```
+
+This ID has to be filterable in all entities, even in product
+
+ ```graphql
+ products(filter: { id: { eq: "$valueOfIdentifierFromUrlResolver"}}) {
```
paliarush: Why do we need to support `eq` and not just `in`?

design-documents/graph-ql/coverage/url-resolver-identifier.md
```diff
@@ -0,0 +1,109 @@
...
+  products(filter: { id: { in: ["ID1", "ID2", "ID3"]}}) {
+      items {
+        id
+        name
+      }
+    }
+  ```
+  
+This is not currently available in current schema, but it has to be introduced to unlock this flow.
+A separate proposal for product will be requested.
```
buskamuza: Is this phrase still relevant?

cpartica: I was probably stating the obvious

design-documents/graph-ql/coverage/url-resolver-identifier.md
cpartica: removed

PR #337, opened 2019-11-04
magento/graphql-ce#196: Swatches: expose Hex color code and swatch images

lenaorobei:
## Solution
Proposed schema for adding swatch data for configurable options.
``` graphql
{
  products(filter: {sku: {eq: "configurable-product"}}) {
    items {
      ... on ConfigurableProduct{    
      configurable_options{
          values {
            label
            swatch_data{
              value
            }
          } 
        }
      }
    }
  }
}
```
## Side notes
- This approach does not align with [current implementation of product swatches](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/SwatchesGraphQl/etc/schema.graphqls), but is semantically correct.
- Current implementation of product swatches cannot be changed in order to use `SwatchDataInterface` due to BC.

## Requested Reviewers
@akaplya 
@paliarush

### DIFF
--- a/design-documents/graph-ql/coverage/Swatches.graphqls
+++ b/design-documents/graph-ql/coverage/Swatches.graphqls
@@ -1,0 +1,19 @@
+type ConfigurableProductOptionsValues {
+    swatch_data: SwatchDataInterface @doc(description: "Swatch data for configurable product option")
+}
+
+interface SwatchDataInterface {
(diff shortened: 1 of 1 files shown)


akaplya:
Approved

paales:
@lenaorobei Was just looking at this, shouldn't the swatch information also be available in the products aggregates?

Issue #269, opened 2019-09-10
Code Style Guidelines for GraphQL


buskamuza:
Document code style guidelines for GraphQL schema.
Currently, the following unwritten agreements exist:
1. `quantity` should be used, not `qty`
2. [Camel case](https://en.wikipedia.org/wiki/Camel_case) SHOULD be used for queries and mutations (root-level fields)
3. [Camel case](https://en.wikipedia.org/wiki/Camel_case) SHOULD be used for arguments
   3.1. Investigate if Snake case should be used (in case it's already more widely used than Camel case)
4. [Upper camel case](https://en.wikipedia.org/wiki/Camel_case) SHOULD be used for types
5. [Snake case](https://en.wikipedia.org/wiki/Snake_case) SHOULD be used for fields

* Investigate if the above rules make sense or other agreements should be introduced. Document the agreements as guidelines.
* Investigate if any other code style guidelines make sense and document them.

jedmao:
1. I agree that we shouldn't be shortening names.
1. Agreed.
1. Not sure what else there is to investigate. The GraphQL docs themselves [use camel case arguments](https://graphql.org/graphql-js/passing-arguments/), which is consistent with everything else.
1. Agreed.
1. Again. GraphQL docs [use camel case everywhere for fields](https://graphql.org/learn/schema/#object-types-and-fields). I think anything else would be considered bad practice.

The most common consumer of a GraphQL API is a browser (i.e., JavaScript). As such, we should conform JavaScript's patterns and use camel case throughout.
- JavaScript clients shouldn't have to change the case into something else in order to have consistent casing in their applications
   - Any such conversions cause overhead in the client, which is unacceptable.
   - A non-native function is required to perform such conversions, which adds additional footprint to the JavaScript payload.

buskamuza:
What do we do with EAV attributes - they are snake case?
* Convert them for consistency

orlangur:
@buskamuza 2-5 why SHOULD and not MUST?

stijnbernards:
> 1. I agree that we shouldn't be shortening names.
> 2. Agreed.
> 3. Not sure what else there is to investigate. The GraphQL docs themselves [use camel case arguments](https://graphql.org/graphql-js/passing-arguments/), which is consistent with everything else.
> 4. Agreed.
> 5. Again. GraphQL docs [use camel case everywhere for fields](https://graphql.org/learn/schema/#object-types-and-fields). I think anything else would be considered bad practice.
> 
[...]

PR #341, opened 2019-11-08
updateCustomerV2 code style fixes

lenaorobei:
## Problem
- Propose schema is inconsistent with existing one. `CustomerOutput` already exists and is used in `createCustomer (input: CustomerInput!): CustomerOutput` mutation.
``` graphql
type CustomerOutput {
    customer: Customer!
}
```
- Removed `updateCustomerPassword` because `changeCustomerPassword(currentPassword: String!, newPassword: String!): Customer` already exists
## Requested Reviewers

@paliarush 
@melnikovi

### DIFF
--- a/design-documents/graph-ql/coverage/customer-email-password-update.md
+++ b/design-documents/graph-ql/coverage/customer-email-password-update.md
@@ -15,7 +15,7 @@ Deprecate `updateCustomer` mutation in favor of `updateCustomerV2`. `CustomerUpd
 
  ```graphql
  mutation {
-    updateCustomerV2(input: CustomerUpdateInput!): Customer
+    updateCustomerV2(input: CustomerUpdateInput!): CustomerOutput
  }
  
  type CustomerUpdateInput {
@@ -36,7 +36,7 @@ Deprecate `updateCustomer` mutation in favor of `updateCustomerV2`. `CustomerUpd
  
  ```graphql
   mutation {
-     updateCustomerEmail(email: String!, password: String!): Customer
+     updateCustomerEmail(email: String!, password: String!): CustomerOutput
   }
  ```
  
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/customer-email-password-update.md
```diff
@@ -36,15 +36,15 @@ Deprecate `updateCustomer` mutation in favor of `updateCustomerV2`. `CustomerUpd
...
+     updateCustomerEmail(email: String!, password: String!): CustomerOutput
   }
  ```
  
  **updateCustomerPassword**
  
  ```graphql
   mutation {
-     updateCustomerPassword(password: String!, old_password: String!): Customer
+     updateCustomerPassword(currentPassword: String!, newPassword: String!): CustomerOutput
```
melnikovi: According to https://github.com/magento/architecture/issues/269, fields should be snake case. I would also propose to keep password and old_password arguments.

design-documents/graph-ql/coverage/customer-email-password-update.md
```diff
@@ -15,7 +15,7 @@ Deprecate `updateCustomer` mutation in favor of `updateCustomerV2`. `CustomerUpd
 
  ```graphql
  mutation {
-    updateCustomerV2(input: CustomerUpdateInput!): Customer
+    updateCustomerV2(input: CustomerUpdateInput!): CustomerOutput
```
melnikovi: I agree that it would not be consistent with existing operations. But I would only change it if we think that we may want to add more data in the future.

design-documents/graph-ql/coverage/customer-email-password-update.md
lenaorobei: But this is argument of the mutation.
Regarding naming, I think we need more thoughts here.
@buskamuza @paliarush

PR #345, opened 2019-12-02
Sequence functionality for entities in Magento

fascinosum:
## Problem

Due to specific install/upgrade flow, when we use Declarative Schema, most of Staging foreign keys recreates on each upgrade.

## Solution

We have 3 ways to explicitly declare new Staging sequence foreign keys:

1. Add 10 new dependencies into 4 Staging modules
2. Introduce about 10 new modules to solve all new dependencies
3. Move database changes from Staging modules into corresponding CE modules

The simplest way to do without backward-incompatible changes which can be implemented in a patch release is the option one - add 10 new dependencies into 4 Staging modules.
This may be a temporary solution of the existing issue.

## Requested Reviewers

@akaplya 
@buskamuza 
@melnikovi


design-documents/database/sequence-functionality-in-magento.md
```diff
@@ -0,0 +1,132 @@
...
+* app/code/Magento/VisualMerchandiser/Setup/Recurring.php
+
+The solution is implemented declaratively using DI, but it still causes some issues with upgrade in EE. 
+
+#### Declarative Schema issue
+Due to specific install/upgrade flow, when we use Declarative Schema, most of Staging foreign keys recreates on each upgrade.
+
+We have 3 ways to explicitly declare new Staging sequence foreign keys:
+
+1. Add 10 new dependencies into 4 Staging modules
```
buskamuza: It would be helpful to cover what exact changes do we need in the modules, that require adding dependencies. Same for all solutions. It's not very clear what exactly should be done in the modules. Dependencies are just consequences.

fascinosum:
It was implemented

PR #347, opened 2019-12-06
Storefront: pre-requisites for split deployment

mslabko:
## Pre-requisites for split deployment (Monolith and Storefront App).

Storeftont application (SF App) has to satisfy two requirements:
- should be performance efficient 
- should be decoupled from Monolith

Problem Statement:
- 1st requirement is addressed using plain SQL queries via Resource Connection object (framework) instead of re-using Magento Service contracts. But this leads to duplication of existing business logic and customizations. Taking into account that SF App supposed to be our prefered approach for GraphQL coverage - we should introduce a robust way which would avoid duplication of Business Logic and preserve existing customization by 3rd party developers.
- 2nd requirement addressed partially as new SF App implementation is not depending (there is still known tech debt here) on Magento Modules (code decoupling), but relies on the same Data storage (data decoupling).

Goal of current proposal is to address both of the issues above:
- Prevent duplication of business logic, preserving existing customizations
- Decouple data storage for Monolith and SF App, describing how to deliver data from the Magento Monolith to the Storefront services.


design-documents/storefront/delivery-data-to-sf.md
```diff
@@ -0,0 +1,36 @@
+## Pre-requisites for split deployment (Monolith and Storefront App).
+
+
+Storeftont application (SF App) has to satisfy two requirements:
```
real34:
```suggestion
Storefront application (SF App) has to satisfy two requirements:
```

paliarush:
Closing per request from @mslabko, since this PR is not relevant anymore.

PR #348, opened 2019-12-19
MQE-1932: MFTF tests deprecation process

jilu1:
## Solution
Propose MFTF test deprecation syntax and process

### DIFF
--- a/design-documents/testing/functional/mftf-entities-deprecation-process.md
+++ b/design-documents/testing/functional/mftf-entities-deprecation-process.md
@@ -1,0 +1,182 @@
+## MFTF Test Entities Deprecation Process
+
+### Terminologies
+
+`MFTF` - Magento Functional Testing Framework
+
+`MFTF Test Entities` - Variously typed building blocks which can be used to construct `MFTF Tests`. Types of `MFTF Test Entities` are `Test`, `Action Group`, `Page`, `Section`, `Data`, `Meta Data`, and `Suite`
+
+`MFTF Tests` - Tests written to verify Magento functionality using `MFTF`
+
+`BC` - Backward Compatible
+
+`BIC` - Backward Incompatible
+
+`Deprecation Script` - Script which removes all `MFTF Test Entities` with _deprecated_ annotation from Magento 2 Core code before a Magento Major Release
+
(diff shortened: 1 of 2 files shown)


KevinBKozan:
Just repeating what I've said from previous reviews, but I'd prefer only using syntax 2 for absolute consistency in deprecation. This process will already be cumbersome, so keeping everything lean where possible is a must in my opinion.

jilu1:
Personally I prefer syntax 1 for its consistency.

@KevinBKozan Do you mean syntax 1 or syntax 2?

KevinBKozan:
@jilu1 apologies, I did mean Syntax 1 :+1:

design-documents/testing/functional/mftf-entities-deprecation-process.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Applicable Scope for Deprecation
+
+Besides deprecations at the entity level,  action group argument and section element are independent themselves and can be deprecated without deprecating the entire entity. The following table shows full scope of where deprecation can occur.
+
+|**Entity Types**|**Entity Elements**|**Deprecation Applicable?**|
+|---|---|---|
+|ActionGroup|`<actionGroup>`|**Yes**
+| |`<actionGroup>` `<action>`|No
+| |`<actionGroup>` `<argument>`|**Yes**
```
tomreece: What does it look like if an `argument` if deprecated?

For example, if an action group has an `argument` and 5 different `actions` in that action group reference that `argument`, then how can you deprecated it without also affecting those 5 `actions`?

tomreece:
I prefer syntax 1 because it is consistent.

design-documents/testing/functional/mftf-entities-deprecation-process.md
jilu1: @tomreece Good catch! Not all action group argument deprecation can be completely isolated. I think it's good to just remove it from the list. Fixed.

jilu1:
@paliarush @okolesnyk It looks like the formal Architectural Discussions meeting process has changed. Can you please review this document and I can setup a informal meeting if further discussion is needed? Thanks!

jilu1:
I updated with more generic version number and make sure major release ends with 0.

PR #351, opened 2020-01-14
Initial architecture proposal for using a meta package for require-dev

dverkade:
## Problem

No longer need to update the require-dev packages in the root composer.json

## Solution

Create a meta package referencing the needed packages to be loaded in the require-dev section. 

## Requested Reviewers

- @joni-jones
- @buskamuza
- @paliarush

### DIFF
--- a/design-documents/composer-require-dev-metapackage.md
+++ b/design-documents/composer-require-dev-metapackage.md
@@ -1,0 +1,67 @@
+# Composer Require-dev Metapackage
+
+This document describes the architectural change to make it easier to update the development packages with composer loaded by Magento.
+
+## Situation
+
+Magento uses some packages for development purposes which are added in the `require-dev` section of the root composer.json file.
+
+## The problem
+
+The root composer.json file is not updated when you update to the latest version of Magento. However sometimes the packages used in the `require-dev` section of the root composer.json file need to be updated to more recent version, or packages need to be added or removed from this section.
+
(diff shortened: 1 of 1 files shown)


milindsingh:
Yes, this would be very nice. and also, I would suggest that dev dependencies are not added by default.

fascinosum:
hi @pdohogne-magento,
could you confirm that `magento/composer-root-update-plugin` fulfills this requirement?

pdohogne-magento:
@fascinosum The plugin makes updating the root composer.json file automatic so the manual steps of updating the `require-dev` section are no longer necessary.  Moving the requirements to a metapackage is still potentially a good idea, though, as it provides a clean way for users to remove all dev dependencies if desired and potentially makes it easier for us to make dev dependencies optional in the future if we decide to go that route.

dverkade:
@paliarush why is this being closed?

paliarush:
@dverkade mainly because updates to root composer are now automated by the `magento/composer-root-update-plugin` plugin (as confirmed by @pdohogne-magento above), which satisfies the use case described in "The Problem" section of the proposal.

Moving dev dependencies to a separate metapackage would be nice to have. Unfortunately, such change would require modifications of Magento release publication process and related automation. In addition, this may cause backward incompatibility for the CI infrastructure of system integrators and merchants. So there must be a very good reason to justify such risks.

PR #352, opened 2020-03-04
Add Playwright to tech vision

jedmao:
Playwright is an exciting, new library that supports testing across Windows, Mac & Linux as well as Chromium, Webkit and Firefox!

https://github.com/microsoft/playwright#playwright


jedmao:
Just realized I can't sign the CLA.

PR #355, opened 2020-03-11
GraphQL Schema for Product Compare

akaplya:


design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
...
+    compareList: CompareList
+}
+
+input CompareListInput {
+    id: ID!
+    items: [ID!]
+}
+
+type Mutation {
+    addItemsToCompareList(list: CompareListInput): CompareList
```
paliarush: Please introduce operation-specific output wrappers for mutation.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
...
+type Customer{
+    compareList: CompareList
+}
+
+input CompareListInput {
+    id: ID!
+    items: [ID!]
+}
+
+type Mutation {
```
paliarush: Please combine these documents with existing proposal https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/product-compare.graphqls

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
...
+    addItemsToCompareList(list: CompareListInput): CompareList
+    removeItemsFromCompareList(list: CompareListInput): CompareList
+    clearList(id: String!): Boolean
```
paliarush: Why it is `String` instead of `ID`?

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
paliarush: It is better to explicitly list all arguments without wrapper.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
...
+    indicators: [ComparableIndicator]
+}
+
+type Customer{
+    compareList: CompareList
+}
+
+input CompareListInput {
+    id: ID!
+    items: [ID!]
```
paliarush: Why not product SKU? Everywhere else in GraphQL we operate based on product SKU.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
...
+    title: String!
+}
+
+type CompareList {
+    items: [ComparableItem]
+    indicators: [ComparableIndicator]
+}
+
+type Customer{
+    compareList: CompareList
```
paliarush: Looks like current schema does not support compare lists for guests. See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/product-compare.graphqls for possible solution.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
...
+    images: [ProductImage]
+    values: [ProductAttribute]
+}
+
+type ComparableIndicator {
+    code: String!
+    title: String!
+}
+
+type CompareList {
```
paliarush: Please provide usage examples (preferably with screenshots of the corresponding pages).

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,38 @@
+type ComparableItem {
+    productId: ID!
```
paliarush: Why not SKU?

design-documents/graph-ql/coverage/catalog/catalog-product-common-types.graphqls
```diff
@@ -0,0 +1,16 @@
+
+type ProductUrl {
```
paliarush: Where these types are planned to be reused outside of products compare?

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
```diff
@@ -0,0 +1,21 @@
+type ProductPriceRange {
```
paliarush: These schema changes should go in scope of separate story with separate scenarios.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: Fixed

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: removed old file

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: Fixed

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: Fixed

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
akaplya: this is not a change, this schema already in the mainline

design-documents/graph-ql/coverage/catalog/catalog-product-common-types.graphqls
akaplya: It should be aligned with the product schema

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: SKU is not a unique guaranteed identifier for the catalog.
Along with this product exposes ID as a part of product type.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: fixed

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: SKU is not a unique guaranteed identifier for the catalog.
Along with this product exposes ID as a part of product type.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
paliarush: It may be confusing. We need to have only one document because they are competing.

design-documents/graph-ql/coverage/catalog/catalog-product-common-types.graphqls
paliarush: Product schema already exists. So looks like we should not introduce new types here.

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
paliarush: Please do not include parts of the schema that already exist in scope of this proposal, it is confusing and will cause problems in the future because of duplication. Just reference any existing types from the new schema.

design-documents/graph-ql/coverage/catalog/catalog-product-common-types.graphqls
zetlen: As a consumer of this schema, I agree with @paliarush and I'd also urge you to consider adding additional interfaces where necessary, instead of special-purpose wrapper types.

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
zetlen: Why is a minimum price required but no maximum price is required? I don't see why either should be required.

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
```diff
@@ -0,0 +1,21 @@
...
+    minimum_price: ProductPrice!
+    maximum_price: ProductPrice
+}
+
+type ProductPrice {
+    regular_price: Money!
```
zetlen: I just want to add that `Money!` is a really funny name for a type.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,46 @@
...
+    items: [ComparableItem]
+    indicators: [ComparableIndicator]
+}
+
+type Customer{
+    compareList: CompareList
+}
+
+type Query {
+    getCompareListForCustomer(customerId: ID): CompareList
```
zetlen: In another comment, I've asked why compare lists belong to customers, but if they must belong to customers, then It looks like a customer has only one compare list at any time. If that is correct, then it might be better to add `compare_list` as a property on `Customer`.

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
akaplya: slightly enterprise-ish https://martinfowler.com/eaaCatalog/money.html

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: good point, fixed.

akaplya: The compare list is always specific to the buyer but a buyer not always a registered customer. Here is a small difference between guests and registered customers. The customer compare list is permanent when a guest list should not.

Regarding `compare/{sku1}+{sku2}...+{skun}` we can not do because of respect current functionality. Actually Magento stores list state at server, what probably not ideal, but such change will cause a significant change in behavior that I know we do not want to do now.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: added scenario

design-documents/graph-ql/coverage/catalog/catalog-product-common-types.graphqls
akaplya: item in compare list not a product in an understanding of a product query that we have, actually it has a different much more limited set of fields. The compare list rather report then entity. it has some product attributes true. Another question is performance, I do not want to use the product as supertype everywhere where this makes sense or does not make.  we had this with Magento, as a result huge performance degradation each time as customization added or one additional available field requested. My strong belief that API should follow to business scenario requirement and provide exactly what needed for the particular scenario on a page.

akaplya: this is a good question, right now product schema has a canonical_url which is just a string, no attributes can be added. which potentially could be a problem, are we good with URL as a string, or we can add a new object field in a backward-compatible manner?

zetlen: A conformant URL is as good as an object. If we need additional properties on the URL string, there are ways to add them within the URI spec.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,49 @@
...
+type Customer{
+    compareList: CompareList
+}
+
+type Customer {
+    compare_list: CompareList
+}
+
+type Query {
+    getCompareListById(id: ID): CompareList
```
paliarush:
```suggestion
    сompareList(id: ID): CompareList
```

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
```diff
@@ -0,0 +1,49 @@
...
+type CompareList {
+    items: [ComparableItem]
+    indicators: [ComparableIndicator]
+}
+
+type Customer{
+    compareList: CompareList
+}
+
+type Customer {
```
paliarush: Why `Customer` type is duplicated in the document?

design-documents/graph-ql/coverage/catalog/catalog-product-price.graphqls
akaplya: this not a part of a change that we are doing right now just a fixation mainline schema that we use.

design-documents/graph-ql/coverage/catalog/compare-list.graphqls
akaplya: removed old document

design-documents/graph-ql/coverage/catalog/catalog-product-common-types.graphqls
akaplya: fixed

design-documents/graph-ql/coverage/catalog/compare-list.md
```diff
@@ -4,6 +4,12 @@
 * Loginned buyer can access the latest compare list if the list has not been cleared with the last session. 
 * The buyer can request a compare list which has to return as a bare minimum price, link to a product page, name, and SKU.
 
+
+* A Compare list ID is a client-side generated ID.
+* A new compare list could be created by calling `addItemsToCompareList`.
+* If a shopper calls  `addItemsToCompareList` with the new id the previous cart will be retired and replaced with a new one.
```
paliarush: This does not sound right:
1. Why we are talking about cart?
2. If we assume it should be compare list, why/how do we retire another one?

design-documents/graph-ql/coverage/catalog/compare-list.md
```diff
@@ -1,19 +1,85 @@
...
+                }
+            }
+        }
+    }
+}
+```
+* If the registered customer does not have an active list then null will be returned.
+This means the client has to generate and send a new ID if the compare list functionality requested.
 
+* If the buyer calls addItemsToCompareList with a new ID the previous list will be disabled.
```
paliarush: It cannot be done for guests because there is not relation between the old and the new list.

PR #368, opened 2020-04-09
ECP-469: Finalize new search schema

kokoc:
# Overview

This PR contains graphql schema changes required to introduce new search functionality such as 
* highlights - parts of product attribute values which contain matched words. The words might be slightly different compared to the input, so they can't be calculated on the client side.
* multi-facets - possibility to check multiple values per facet. This means that in addition to filtered results we need to get the facets without any filters. 
* advanced range facets (prices) - support of slider UX element. Basically, we need to extract additional information about price ranges, like min value; max value, etc.
* suggestions, "did you mean?", etc  - not a part of the proposed schema, but we need to make sure these parts are fit well in the future.
* preselected variations - most relevant variation is preselected and rendered instead of configurable

The main challenge is to find a proper place in existing/new queries for new fields. 

# Inject into one of the existing types/queries

## `products` query
```graphql
type Query {
    products (
        search: String,
        filter: ProductAttributeFilterInput,
        pageSize: Int = 20,
        currentPage: Int = 1,
        sort: ProductAttributeSortInput
    ): Products
}
type Products {
    items: [ProductInterface]
    page_info: SearchResultPageInfo
    total_count: Int
    filters: [LayerFilter]
    aggregations: [Aggregation]
    sort_fields: SortFields
}
```

`products` query covers two main use cases: search by search phrase and catalog browsing with filtering by category.
This query is the most obvious choice for search-related customizations.
However, this query returns a `Products` type with `ProductInterface` inside. The `ProductInterface` is designed to
represent a product entity with complex product types support.

Adding fields to `ProductInterface` or `Products` will
make those fields optional and useless in "catalog browsing" scenario. Other words, our query is handling two different
use cases with two different results. Mixing those results in very common `ProductInterface` is messy,
inconvenient and just strange.

`Products` type is a little bit better place for new fields. We can put request-specific fields here, but we also need to introduce one more items conainer for
search results(will contain highlights + products). This approach is still messy, but at least it doesn't affect other product usecases (shopping cart item, wishlist item, pdp, etc).
Example for this option:
```graphql
type Query {
    products (
        search: String,
        filter: ProductAttributeFilterInput,
        pageSize: Int = 20,
        currentPage: Int = 1,
        sort: ProductAttributeSortInput
    ): Products
}
type Products {
    items: [ProductInterface] @deprecated(???????)
    search_items: [ProductSearchItem]
    suggestions: [String]
    page_info: SearchResultPageInfo
    total_count: Int
    filters: [LayerFilter]
    aggregations: [Aggregation]
    sort_fields: SortFields
}
type ProductSearchItem {
    product: ProductInterface!
    highlights: [HighlightType]
}
```

## other queries with product results
There are two more queries which return products and could be considered as a potential candidate for new fields:
category and categoryList. Both return `CategoryInterface` interface with `products` field inside:
```graphql
interface CategoryInterface
    id: Int
    ...
    default_sort_by: String 
    products(
        pageSize: Int = 20 
        currentPage: Int = 1 
        sort: ProductAttributeSortInput 
    ): CategoryProducts 
    breadcrumbs: [Breadcrumb] 
}

type CategoryProducts 
    items: [ProductInterface] 
    page_info: SearchResultPageInfo 
    total_count: Int 
}

```

This interface also returns `ProductInterface` and also those queries do not receive a search phrase. Thus it's they are not a good candidates for new search functionality.

# Proper place
In order to find the proper place we may take a look on the use cases we have and want to support:

* Browsing by category - includes layered navigation (facets). Supports further filtering.
* Searching by phrase - includes layered navigation (facets). Supports further filtering. Additional ordering by relevancy, highlights, variation preselect, suggestions
* Advanced search - filtering by some values & full-text search by another
* Product details page
* Various product representation - wishlist item, shopping cart item
* Category navigation

I think the following table is the desired assigment of queries per use cases:

| Scenario | Current Query | Suggested Query | Required changes |
| ------------- | ---- | ---- | ---- |
| Browsing by category | products<br/>category<br/>categoryList | category | Add layered navigation results<br/>Add filters to product sub-request |
| Searching by phrase  | products | new query | deprecate search related fields in `products` query |
| Advanced search      | n/a | TBD | TBD<br/>Probably can be omitted |
| PDP                  | products | products | deprecate filtering except product id |
| Category Navigation  | categoryList | categoryList | deprecate all non-category fields |

You can find a possible shape of new query in PR content.

From the table above we can see that `products` query will be used mostly for retrieving the product by their ids, other functionality will be transferred to `productSearch` query.
Thus we may also deprecate entire `products` query entirely and create a new query designed to get products by ids.

Here is a possible shape of the api if we decide to deprecated old query:
```graphql
type Query {
    "This query is deprecated by still works for backwards-compatibilty purposes"
    products (
            search: String,
            filter: ProductAttributeFilterInput,
            pageSize: Int = 20,
            currentPage: Int = 1,
            sort: ProductAttributeSortInput
        ): Products @deprecated("See product or productSearch queries")

     productSearch(
         phrase: String!,
         "Desired size of the search result page"
         pageSize: Int = 20,
         currentPage: Int = 1,
         filter: [SearchClauseInput],
         sort: [ProductSearchSortInput]
     ): ProductSearchResponse!

     product(id: ID!): ProductInterface
}
```

Here are few possible examples per use case:

_Category Listing_
```graphql
category(id: 17) {
    products {
        items {
            id
            name
            sku
            thumbnail {
                url
            }
        }
    }
}
```

_Search by phrase_
```graphql
productSearch(phrase: "Bags") {
    items {
       product {
            id
            name
            sku
            thumbnail {
                url
            }
       }
       hightlights {
           attribute_code
           value
       }
    }
}
```

_Product details page_

```graphql
product(id: 18) {
    id
    name
    sku
    media_gallery_entries {
        url
        media_type
        content
    }
}
```

_Category navigation_
```graphql
categoryList {
    id
    name
    url_path
    children {
        id
        name
        url_path
        children {
            id
            name
            url_path
        }
    }
}
```

### DIFF
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -4,7 +4,14 @@
 type Query {
     multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
     #Filter supports multiple clauses which will be wrapped in logical AND operator
-    productSearch(phrase: String!, filter: [SearchClauseInput], sort: [ProductSearchSortInput]): ProductSearchResponse!
+    productSearch(
+        phrase: String!,
+        "Desired size of the search result page"
+        pageSize: Int = 20,
+        currentPage: Int = 1,
+        filter: [SearchClauseInput],
+        sort: [ProductSearchSortInput]
+    ): ProductSearchResponse!
 }
 
 input ProductSearchSortInput
@@ -15,12 +22,12 @@ input ProductSearchSortInput
 
 # If from or to fields are omitted, $gte or $lte filter will be applied
 input SearchRangeInput {
-    from: Int
-    to: Int
+    from: Float
+    to: Float
 }
 
 input SearchClauseInput {
-    attribute: String!
+    attribute_code: String!
     in: [String]
     eq: String
     range: SearchRangeInput
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -35,6 +35,7 @@ type ProductSearchResponse {
 
 type MultiSearchResponse {
     products: ProductSearchResponse
+    suggestions: [String]
 }
 
 type ProductSearchItem {
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -4,18 +4,17 @@
 type Query {
     multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
     #Filter supports multiple clauses which will be wrapped in logical AND operator
-    productSearch(phrase: String!, filter: [SearchClauseInput], sort: ProductSearchSortInput): ProductSearchResponse!
+    productSearch(phrase: String!, filter: [SearchClauseInput], sort: [ProductSearchSortInput]): ProductSearchResponse!
 }
 
 input ProductSearchSortInput
 {
-    relevance: SortEnum
-    position: SortEnum
-    name: SortEnum
+    attribute: String!
+    direction: SortEnum
 }
 
 # If from or to fields are omitted, $gte or $lte filter will be applied
-type SearchRangeInput {
+input SearchRangeInput {
     from: Int
     to: Int
 }
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -52,7 +52,6 @@ type Highlight {
 interface Bucket {
     #Human readable bucket title
     title: String!
-    attribute: String!
 }
 
 type StatsBucket implements Bucket {
@@ -73,7 +72,7 @@ type RangeBucket implements Bucket {
 }
 
 interface Aggregation {
-    name: String!
+    attribute_code: String!
     buckets: [Bucket]!
 }
 ```
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -2,7 +2,6 @@
 
 ```graphql
 type Query {
-    multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
     #Filter supports multiple clauses which will be wrapped in logical AND operator
     productSearch(
         phrase: String!,
@@ -41,11 +40,6 @@ type ProductSearchResponse {
     related_terms: [String]
     page_info: SearchResultPageInfo
     total_count: Int
-}
-
-type MultiSearchResponse {
-    products: ProductSearchResponse
-    suggestions: [String]
 }
 
 type ProductSearchItem {
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -52,16 +52,16 @@ type Highlight {
 interface Bucket {
     #Human readable bucket title
     title: String!
+    attribute: String!
 }
 
 type StatsBucket implements Bucket {
-    #Could be used for filtering and may contain non-human readable data
-    id: ID!
     min: Int!
     max: Int!
 }
 
 type ScalarBucket implements Bucket {
+    #Could be used for filtering and may contain non-human readable data
     id: ID!
     count: Int!
 }
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -4,7 +4,7 @@
 type Query {
     multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
     #Filter supports multiple clauses which will be wrapped in logical AND operator
-    productSearch(phrase: String!, filter: [Clause], sort: ProductSearchSortInput): ProductSearchResponse!
+    productSearch(phrase: String!, filter: [SearchClauseInput], sort: ProductSearchSortInput): ProductSearchResponse!
 }
 
 input ProductSearchSortInput
@@ -14,10 +14,17 @@ input ProductSearchSortInput
     name: SortEnum
 }
 
-input Clause {
+# If from or to fields are omitted, $gte or $lte filter will be applied
+type SearchRangeInput {
+    from: Int
+    to: Int
+}
+
+input SearchClauseInput {
     attribute: String!
     in: [String]
     eq: String
+    range: SearchRangeInput
 }
 
 type ProductSearchResponse {
@@ -42,21 +49,31 @@ type Highlight {
     matchedWords: [String]!
 }
 
-interface Aggregation {
-    name: String!
-    buckets: [Bucket]!
+interface Bucket {
+    #Human readable bucket title
+    title: String!
 }
 
-interface Bucket {
-    count: Int!
+type StatsBucket implements Bucket {
+    #Could be used for filtering and may contain non-human readable data
+    id: ID!
+    min: Int!
+    max: Int!
 }
 
 type ScalarBucket implements Bucket {
-    title: String!
+    id: ID!
+    count: Int!
 }
 
 type RangeBucket implements Bucket {
     from: Float!
     to: Float!
+    count: Int!
+}
+
+interface Aggregation {
+    name: String!
+    buckets: [Bucket]!
 }
 ```
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -1,0 +1,62 @@
+# Queries
+
+```graphql
+type Query {
+    multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
+    #Filter supports multiple clauses which will be wrapped in logical AND operator
+    productSearch(phrase: String!, filter: [Clause], sort: ProductSearchSortInput): ProductSearchResponse!
+}
+
+input ProductSearchSortInput
+{
+    relevance: SortEnum
+    position: SortEnum
+    name: SortEnum
+}
+
+input Clause {
+    attribute: String!
+    in: [String]
+    eq: String
+}
+
+type ProductSearchResponse {
+    items: [ProductSearchItem]
+    hits: Int!
+    facets: [Aggregation]
+    facetsValues: [Aggregation]
+}
+
+type MultiSearchResponse {
+    products: ProductSearchResponse
+}
+
+type ProductSearchItem {
+    product: ProductInterface!
+    highlights: [Highlight]
+}
+
+type Highlight {
+    attribute: String!
+    value: String!
+    matchedWords: [String]!
+}
+
+interface Aggregation {
+    name: String!
+    buckets: [Bucket]!
+}
+
+interface Bucket {
+    count: Int!
+}
+
+type ScalarBucket implements Bucket {
+    title: String!
+}
+
+type RangeBucket implements Bucket {
+    from: Float!
+    to: Float!
+}
+```
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -38,6 +38,7 @@ type ProductSearchResponse {
     hits: Int!
     facets: [Aggregation]
     facets_values: [Aggregation]
+    suggestions: [String]
 }
 
 type MultiSearchResponse {
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -10,7 +10,7 @@ type Query {
 input ProductSearchSortInput
 {
     attribute: String!
-    direction: SortEnum
+    direction: SortEnum!
 }
 
 # If from or to fields are omitted, $gte or $lte filter will be applied
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -30,7 +30,7 @@ type ProductSearchResponse {
     items: [ProductSearchItem]
     hits: Int!
     facets: [Aggregation]
-    facetsValues: [Aggregation]
+    facets_values: [Aggregation]
 }
 
 type MultiSearchResponse {
@@ -46,7 +46,7 @@ type ProductSearchItem {
 type Highlight {
     attribute: String!
     value: String!
-    matchedWords: [String]!
+    matched_words: [String]!
 }
 
 interface Bucket {
--- a/design-documents/graph-ql/coverage/search.md
+++ b/design-documents/graph-ql/coverage/search.md
@@ -35,10 +35,12 @@ input SearchClauseInput {
 
 type ProductSearchResponse {
     items: [ProductSearchItem]
-    hits: Int!
     facets: [Aggregation]
     facets_values: [Aggregation]
     suggestions: [String]
+    related_terms: [String]
+    page_info: SearchResultPageInfo
+    total_count: Int
 }
 
 type MultiSearchResponse {
@@ -63,8 +65,8 @@ interface Bucket {
 }
 
 type StatsBucket implements Bucket {
-    min: Int!
-    max: Int!
+    min: Float!
+    max: Float!
 }
 
 type ScalarBucket implements Bucket {


design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
+# Queries
+
+```graphql
+type Query {
+    multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
+    #Filter supports multiple clauses which will be wrapped in logical AND operator
+    productSearch(phrase: String!, filter: [SearchClauseInput], sort: ProductSearchSortInput): ProductSearchResponse!
+}
+
+input ProductSearchSortInput
```
kandy, line 10: What about sort by price or other attributes?

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+{
+    relevance: SortEnum
```
kandy, line 12: How will sort by relevance and position in same time will work?

kokoc: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-sort.html#search-request-sort - most of engines support sorting by multiple fields. Some combinations may not have any business value, but there is no technical issues

design-documents/graph-ql/coverage/search.md
kokoc: fixed

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+    in: [String]
+    eq: String
+    range: SearchRangeInput
+}
+
+type ProductSearchResponse {
+    items: [ProductSearchItem]
+    hits: Int!
+    facets: [Aggregation]
+    facetsValues: [Aggregation]
```
kandy, line 33: What the difference between facets and facet Values?

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+    count: Int!
+}
+
+type RangeBucket implements Bucket {
+    from: Float!
+    to: Float!
+    count: Int!
+}
+
+interface Aggregation {
```
kandy, line 75: I don't understand the relation between `SearchClauseInput.attribute` and Aggregation
Can you describe how to achieve scenario:
- search by "Iphpne 11" -> result with color attribute facets
- drill down by color facets

kandy:
Can I ask you to describe scenario that will be covered by this API?

design-documents/graph-ql/coverage/search.md
kokoc: FacetValues provide information about all possible facet values for current search phrase. E.g. you may filter by rating=4, but we need information about all possible rating values 1-5 in order to render the UI

design-documents/graph-ql/coverage/search.md
kokoc: `SearchClauseInput` is a filter, you may filter by any attributes and values which were retrieved from aggregations. Thus the main flow will be:
- Perform a search request with some phrase
- Extract aggregations from the response and render them in layered navigation
- If user clicks on some filter in layer navigation section, apply this filter in new query + apply the old search phrase if needed

kandy: Actually question is how to map Aggregation.? (what the field I need to use) into SearchClauseInput.attribute?

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+    eq: String
+    range: SearchRangeInput
+}
+
+type ProductSearchResponse {
+    items: [ProductSearchItem]
+    hits: Int!
+    facets: [Aggregation]
+    facetsValues: [Aggregation]
+}
```
maghamed, line 34: for search response, it will probably make sense to have a request phrase as well.
Especially if we would support "Did You Mean?" functionality where we would automatically change the initial query, for example, fixing smelling there. 

Also, It might make sense to show the "real query" which was used for search.
For example, "IPhane" search request to google - https://www.google.com/search?q=iphane&rlz=1C5CHFA_enUA813UA813&oq=iphane&aqs=chrome..69i57j0l7.4027j0j8&sourceid=chrome&ie=UTF-8

Makes automatic search for corrected version "iphone", but along with that provides possibility to make a search by "iphane"

```
Showing results for iphone
Search instead for iphane
```
And if a user wants to see results for `IPhane` we also need to specify a way to not modify initial search phrase

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+interface Bucket {
+    #Human readable bucket title
+    title: String!
+    attribute: String!
+}
+
+type StatsBucket implements Bucket {
+    min: Int!
+    max: Int!
+}
```
maghamed, line 61: what would we have the stats bucket for? Should we return stats for the whole query or for each particular bucket? 

If it contains stats I think we should return also median/average/stddev etc.
Like an example here 

```
curl http://localhost:8983/solr/techproducts/query -d '
q=memory&
fq=inStock:true&
json.facet={
  "avg_price" : "avg(price)",
  "num_suppliers" : "unique(manu_exact)",
  "median_weight" : "percentile(weight,50)"
}'

The response to the facet request above will start with documents matching the root domain (docs containing "memory" with inStock:true) then calculate and return the requested metrics:

 "facets" : {
    "count" : 4,
    "avg_price" : 109.9950008392334,
    "num_suppliers" : 3,
    "median_weight" : 352.0
  }
```
https://lucene.apache.org/solr/guide/7_5/json-facet-api.html

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
+# Queries
+
+```graphql
+type Query {
+    multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
+    #Filter supports multiple clauses which will be wrapped in logical AND operator
+    productSearch(phrase: String!, filter: [SearchClauseInput], sort: [ProductSearchSortInput]): ProductSearchResponse!
```
maghamed, line 7: right now advanced search does not provide facets. 
I think faceting should be optional and specified at request time

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+}
+
+input ProductSearchSortInput
+{
+    attribute: String!
+    direction: SortEnum!
```
maghamed, line 13: why should we have direction as a mandatory attribute? 
Most of RDBMS just use a default value of ASC / DESC

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
+# Queries
+
+```graphql
+type Query {
+    multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
```
DrewML, line 5: I don't know that `multiSearch` and `productSearch` have the clearest names - it's hard to tell without digging through their types what their purposes are.

Can we rename them or include a description of their intended uses?

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+
+input ProductSearchSortInput
+{
+    attribute: String!
+    direction: SortEnum!
+}
+
+# If from or to fields are omitted, $gte or $lte filter will be applied
+input SearchRangeInput {
+    from: Int
```
DrewML, line 18: `from` and `to` aren't the most descriptive names here, and their value is a scalar which doesn't help a developer infer what value to pass.

Is there a name that would make this more clear? I'm guessing `from` and `to` are index offsets based on the ordered collection of results. In that case, maybe `startIndex` or `startOffset` and an equivalent for `to`.

Also not clear as a developer consuming this schema whether these offsets are inclusive or not

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,79 @@
...
+input SearchClauseInput {
+    attribute: String!
+    in: [String]
+    eq: String
+    range: SearchRangeInput
+}
+
+type ProductSearchResponse {
+    items: [ProductSearchItem]
+    hits: Int!
```
DrewML, line 31: I noticed we use `hits` here, but in other parts of the GraphQL schema we use `total_count` as the int field representing the length of a collection. Any reason not to continue using that name?

DrewML:
> @kandy: Can I ask you to describe scenario that will be covered by this API?

Would also like to see this, with a few example queries showing how the front-end would cover the scenarios

design-documents/graph-ql/coverage/search.md
DrewML: Is range == price range? https://devdocs.magento.com/guides/v2.3/rest/search-endpoint.html#advanced-searches

design-documents/graph-ql/coverage/search.md
kokoc: Everything is optional in graphql, you need to explicitly request what you need. Also advanced search functionality is not currently covered by grapqhl and I doubt if we want to cover it.

design-documents/graph-ql/coverage/search.md
kokoc: Price range + potentially range of any numerical values

design-documents/graph-ql/coverage/search.md
kokoc: They share the same attribute_code

design-documents/graph-ql/coverage/search.md
kokoc: Everything in response is optional in graphql. Moreover, advanced search is not currently covered and I didn't plan to cover it in this iteration, because functionality itself is very questionable and repeats most of the regular search

design-documents/graph-ql/coverage/search.md
kokoc: Here is official elastic search documentation:
```
The order defaults to desc when sorting on the _score, and defaults to asc when sorting on anything else.
```
I don't like this uncertainty

`ProductSearchSortInput` is optional so we can apply some default order, but if developer asks for custom ordering I think it's better to ask for direction as well.

design-documents/graph-ql/coverage/search.md
kokoc: Added suggestions to the response

kokoc: I think we don't have search phrase auto swap today, so skipping this points. We can add it later in search response

design-documents/graph-ql/coverage/search.md
kokoc: Aligned with other queries

design-documents/graph-ql/coverage/search.md
kokoc: I think we have only these use cases for now. We can extend stats bucket later.

design-documents/graph-ql/coverage/search.md
zetlen: I suggest `siteSearch`, assuming that search results may aggregate catalog, CMS and syndicated content.

design-documents/graph-ql/coverage/search.md
DrewML: Seems like a price range filter would need `from` and `to` to be our `Money` type, otherwise I can't filter between $1.00 and $1.99

design-documents/graph-ql/coverage/search.md
DrewML: > I suggest siteSearch, assuming that search results may aggregate catalog, CMS and syndicated content.

I think that's getting closer, but 2 concerns:

1. I don't think either of these queries are designed to be a generic entity search (they both return data about products). Did I read the types wrong, or is the suggestion that we expand the design of this along with the name?
2. `siteSearch` could lead folks to believe, based on the name, that it lets you search across _all_ different ["sites"](https://devdocs.magento.com/guides/v2.3/config-guide/multi-site/ms_over.html) within

kokoc: We will add categories there later

DrewML: I think my preference would be to avoid taking `search` right now, since this search is only for products. I think we should leave names like `search` for a time when we might have some global search feature (one search for catalogs/products/cms/etc)

design-documents/graph-ql/coverage/search.md
DrewML: I like @maghamed's idea of having a default.

> I don't like this uncertainty

I don't see much uncertainty from that quote, if we're only talking about ES.

But, are we planning to only support Elastic Search? If we're ever going to swap search providers, I think it would make sense for the application to define the default so it's consistent.

> ProductSearchSortInput is optional so we can apply some default order

I think an application-defined default sort order would be great 👍

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,88 @@
+# Queries
+
+```graphql
+type Query {
+    multiSearch(phrase: String!, productSize: Int = 10): MultiSearchResponse!
```
DrewML, line 5: Can an example be provided where `multiSearch` is used, where `productSearch` couldn't have been used instead?

I'm still not quite clear on what the difference are.

design-documents/graph-ql/coverage/search.md
DrewML: > I still want to see reusable numerical input/filter.

What's the motivation for needing a reusable filter?

My concern is that the docs are always going to need to explain the various uses of `from` and `to` since they're so generic. Ideally the names of items in the schema are descriptive enough that you don't need to consult docs for common operations like a search.

IMO, we'd be taking a trade-off: less schema to maintain, but more docs and reading for API consumers.

design-documents/graph-ql/coverage/search.md
DrewML: Starting to agree with you more here @kokoc. I took a look again at the usage of `ProductSearchSortInput`, and I guess it will always be specified right next to `attribute_code`, so generic `to`/`from` `float` will probably cover that 👍

design-documents/graph-ql/coverage/search.md
DrewML: Starting to agree with you more here @kokoc. I took a look again at the usage of `ProductSearchSortInput`, and I guess it will always be specified right next to `attribute_code`, so generic `to`/`from` `float` will probably cover that 👍

design-documents/graph-ql/coverage/search.md
kokoc: This was the exact purpose of the query. I removed it for now as we don't plan to use it in the nearest future. We can always add it.

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
...
+type Query {
+    #Filter supports multiple clauses which will be wrapped in logical AND operator
+    productSearch(
+        phrase: String!,
+        "Desired size of the search result page"
+        pageSize: Int = 20,
+        currentPage: Int = 1,
+        filter: [SearchClauseInput],
+        sort: [ProductSearchSortInput]
+    ): ProductSearchResponse!
```
DrewML, line 13: `productSearch` should be nullable, or all other operations under `Query` in a request will fail. See #372

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
+# Queries
+
+```graphql
+type Query {
+    #Filter supports multiple clauses which will be wrapped in logical AND operator
+    productSearch(
```
paliarush, line 6: How do we implement search across multiple entity types, like CMS, Category, Product? Don't we want to have unified search API?

DrewML: We discussed this in some other comments and on a few calls. The consensus IIRC was that we want it at some point in the future, but it's not an immediate need now. 

We thought that, for now, we'll be fine with type-specific searches, since those can still be batched into a single query.

```graphql
query SearchProductsAndCategories($query: String) {
   searchProducts(phrase: $query) {
      name
   }

   searchCategories(phrase: "$query" {
	  name
   }
}
```

This will not be suitable, however, in the case where we want the scoring/ranking of results to happen across types. But talking with @kokoc it sounded like this wasn't something we were ready to support in the near future.

design-documents/graph-ql/coverage/search.md
kokoc: removed

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
...
+}
+
+# If from or to fields are omitted, $gte or $lte filter will be applied
+input SearchRangeInput {
+    from: Float
+    to: Float
+}
+
+input SearchClauseInput {
+    attribute_code: String!
```
ckatili, line 29: Inconsistent field name, `attribute_code` here vs. `attribute` on other type.

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
...
+
+type Highlight {
+    attribute: String!
+    value: String!
+    matched_words: [String]!
+}
+
+interface Bucket {
+    #Human readable bucket title
+    title: String!
```
ckatili, line 58: All fields under the interface (e.g. Bucket) must be listed under any type that implements the interface. e.g. I don't see `title` listed under StatsBucket, ScalarBucket and RangeBucket.

ckatili: Add `attribute` under Bucket interface. This is so the client can use it on the subsequent requests (when user filters by facets). Without it, the client will not have an easy way to find the attribute value to use to begin with.

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
...
+
+type ProductSearchItem {
+    product: ProductInterface!
+    highlights: [Highlight]
+}
+
+type Highlight {
+    attribute: String!
+    value: String!
+    matched_words: [String]!
```
ckatili, line 53: It is a common convention to use camelCase for field name  (e.g.  matchedWords instead of matched_words). Not sure if this is Magento convention or else.

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
...
+    count: Int!
+}
+
+type RangeBucket implements Bucket {
+    from: Float!
+    to: Float!
+    count: Int!
+}
+
+interface Aggregation {
```
ckatili, line 78: Is this supposed to be a `type` instead of an `interface`?

design-documents/graph-ql/coverage/search.md
```diff
@@ -0,0 +1,82 @@
...
+    attribute_code: String!
+    in: [String]
+    eq: String
+    range: SearchRangeInput
+}
+
+type ProductSearchResponse {
+    items: [ProductSearchItem]
+    facets: [Aggregation]
+    facets_values: [Aggregation]
```
ckatili, line 38: As discussed with @kokoc verbally, I think `facets` is sufficient for storing facets aggregation and I am not clear the use case for `facet_values`.

design-documents/graph-ql/coverage/search.md
ckatili: I think we should add `title` field for storing aggregation friendly name. This is so the client (e.g. UI) can use it for display, instead of displaying attribute code.

paales:
Posted the following question earlier in Slack:

> Question/feature request about product aggregations:
>
> To get the layered navigation working properly pwa-studio is/we are relying on introspection queries because the aggregations don’t offer all the information required to build a proper layered navigation. This makes building frontends rather complex.
>
> Wouldn’t it be better to actually have an `AggregationInterface` that will be implemented with:
> - `AggregationEquals` must implement the options { label, value }. Each option must also implement swatch information?
> - `AggregationMatch` must not implement options
> - `AggregationNumberRange` must imlement options: { label, from, to } instead of the concatenated string `100_*` which requires custom handling, Can I just pass that value back to `ProductAttributeFilterInput`? The values there should be Int or Float, not String?
> - `AggregationPriceRange` can be implement to indicate a field should be rendered with a price in mind, else its the same as AggregationNumberRange
> - `AggregationBoolean` must not implement options

And @kokoc responded to me that this should be covered by this new search api and gave me the possibility to actually check it out! 🎉   So I've been trying out the proof-of-concept graphql endpoint (https://search-api-dev.magento-ds.com/graphql) and have some feedback / questions:

### [`SearchClauseInput`](https://github.com/magento/architecture/pull/368/files#diff-42297184b0f68484e29d653d30dd8930R28-R33)

Compared to the older `products` query we're loosing some functionality here: We are now unable to discover the attributes and possible input types (equals, range, match(?)).

As a developer I'd like to be able to discover which attributes are available for filtering like it was possible with `ProductAttributeFilterInput` so that I don't need to 'guess' which ones are available.

A great alternative might be something we see in our GraphQL CMS ([GraphCMS](https://graphcms.com/)):

```graphql
input ProductSearchFilterInput {
  category: ID
  category_not: ID
  category_in: [ID!]
  category_not_in: [ID!]
  createdAt: DateTime
  createdAt_not: DateTime
  createdAt_in: [DateTime!]
  createdAt_not_in: [DateTime!]
  createdAt_lt: DateTime
  createdAt_lte: DateTime
  createdAt_gt: DateTime
  createdAt_gte: DateTime
  handle: String
  handle_not: String
  handle_in: [String!]
  handle_not_in: [String!]
  price: Float
  price_not: Float
  price_in: [Float!]
  price_not_in: [Float!]
  price_lt: Float
  price_lte: Float
  price_gt: Float
  price_gte: Float
  size: Float
}
```

This keeps the query arguments relatively flat, which is nice.

### [`Aggregation`](https://github.com/magento/architecture/pull/368/files#diff-42297184b0f68484e29d653d30dd8930R78-R81) and [`Bucket`](https://github.com/magento/architecture/pull/368/files#diff-42297184b0f68484e29d653d30dd8930R56-R76)

When rendering an aggregation (should it be renamed to `Facet`?) on the frontend we should be able to see which type of facet we're actually rendering. Wouldn't something like the following be a better idea:


- Rename `Aggregation` to `Facet`
- Inline `StatsBucket` since there can never be more than one.
- `Bucket` seems like an odd name to have? Maybe should be renamed to something different?

```graphql
interface Bucket {
    title: String!
}
type ScalarBucket implements Bucket {
    id: ID!
    count: Int!
}
type RangeBucket implements Bucket {
    from: Float!
    to: Float!
    count: Int!
}

interface Facet {
    attribute_code: String!
    buckets: [Bucket!]!
}
type StatsFacet implements Facet {
    attribute_code: String!
    min: Float!
    max: Float!
    buckets: [ScalarBucket!]
}
type ScalarFacet implements Facet {
    attribute_code: String!
    buckets: [ScalarBucket!]!
}
type RangeFacet implements Facet {
    attribute_code: String!
    buckets: [RangeBucket!]!
}
```

#### Demo implementation

There are some inconsistencies on the demo endpoint that I think will/should be resolved at some point. So not sure how relevant they are:

- `HighLight.matched_words` is empty
- `HighLight.value` contains HTML, probably shouldn't
- `ProductItem` is something that isn't available in the written spec here, will become `ProductInterface` in the end?

I hope you find this valuable!

paales:
Oh, I forgot something, Shouldn't ScalarBucket be have additional implementations like ImageBucket, ColorBucket and TextBucket?

DrewML:
@kokoc @nrkapoor what's the status with this PR? Seems like @paales might have some useful input

design-documents/graph-ql/coverage/search.md
kokoc: I think there is no need for any changes. We have the optional `ProductSearchSortInput`, so we can apply default sort. However, the user needs to specify direction for every custom sort. Does it work?

PR #373, opened 2020-04-27
Integration test data fixture change proposal

bubasuma:
## Problem

Actually there is no formal way to customize existing data fixture directly from the test case which leads to duplicated data fixtures.

## Solution

Extend the format of data fixture annotation to support a second parameter which will be injected to the data fixture file to customize the fixture for a specific test scenario.

## Requested Reviewers
@buskamuza

### DIFF
--- a/design-documents/testing/integration/data-fixtures.md
+++ b/design-documents/testing/integration/data-fixtures.md
@@ -27,52 +27,30 @@ to support a second parameter which will be injected to the data fixture file as
 The string after comma following the fixture file name is indeed the data that needs to be injected into the fixture file for customization. The format is well known JSON format that gives a flexible way to pass any type of data to the fixture (string, int, float and array).
 
 ```php
-use Magento\Catalog\Api\Data\ProductInterface;
-use Magento\Catalog\Api\Data\ProductInterfaceFactory;
-use Magento\Catalog\Api\ProductRepositoryInterface;
-use Magento\Catalog\Model\Product\Attribute\Source\Status;
-use Magento\Catalog\Model\Product\Visibility;
-use Magento\Framework\DataObject;
+use Magento\TestFramework\Catalog\Model\ProductFixtureFactory;
 use Magento\TestFramework\Helper\Bootstrap;
 
-$productData = $productData ?? [];
-$defaultProductData = [
-    'id' => 21,
-    'sku' => 'virtual-product',
-    'name' => 'Virtual Product',
-    'attribute_set_id' => 'sku',
-    'website_ids' => 'sku',
-    'price' => 'sku',
-    'visibility' => Visibility::VISIBILITY_BOTH,
-    'status' => Status::STATUS_ENABLED,
-    'stock_data' => Status::STATUS_ENABLED,
-];
-
-$productData = array_merge($productData, $defaultProductData);
 $objectManager = Bootstrap::getObjectManager();
-$productFactory = $objectManager->get(ProductInterfaceFactory::class);
-/** @var ProductRepositoryInterface $productResource */
-$productRepository = $objectManager->get(ProductRepositoryInterface::class);
-/** @var ProductInterface $product */
-$product = $productFactory->create();
-$product = data_object_hydrate_from_array($product, $productData);
-$productResource->save($product);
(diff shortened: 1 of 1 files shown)


design-documents/testing/integration/data-fixtures.md
```diff
@@ -0,0 +1,154 @@
...
+- Reuse one of these data fixtures in a new data fixture file and edit the quantity there.
+- Reuse one of these data fixtures in your test case and edit the quantity in the test directly.
+ 
+## Solution
+
+We could certainly redesign data fixtures to be object oriented. But if we just want to tackle this issue as simpler as we can, we could consist we could extend the format of data fixture annotation
+to support a second parameter which will be injected to the data fixture file as following.
+```php
+/**
+ * @magentoDataFixture Magento/Catalog/_files/product_virtual.php, {"productData":{"stock_data": {"qty":1}}}
```
buskamuza, line 23: As I understand, there will be only one argument always - a JSON string, and it may include multiple keys in case the fixture creates multiple different instances? Just wanted to clarify.

Should we not use "," to be consistent with other annotations? Like the one for `@magentoConfigFixture`.

mykhailomatiola:
Interesting idea, but it has several problems:
1. Works well with a simple fixture (simple product, cms page), doesn't work well with complex fixtures(quote, order, bundle product)
For example if we create some basic factory for quote some other basic types should be passed into it(product as quote item, customer if this is not a guest quote and other...). But if we need some specific quote (with bundle product with specific options) but we have a factory for simple bundle product, json for such quote became huge because wee need to create such product inside annotation via json.
2. It is difficult to transfer data from other fixtures. You have to do everything inside (duplication of creation) or transfer static data from the outside, which is not always possible, since most entity relations work through the IDs.
 For example product with category. You have already created category in another fixture and you need to path category id into your fixture.
3. In case of a big fixture, json becomes unreadable, the name of the fixture isn't informative, and we just transfer static data from the fixture to test annotation.
For example If you analyze mentioned above the simple quote constructor should be name something like 'simple_quote'. But if we need some specific quote the name of basic quote will be the same. We just need to add huge json to annotation. After this we need analyze all json to understand difference between this and simple quote.  Currently, to understand a quote type simply enough read a name.
4. This fixtures is untyped, out of the box it is not clear how to use it.
5. Most of all simple fixtures already created.

bubasuma:
@mykhailomatiola 
This proposal does not focus on how you create fixtures data or how you create complex entities like quote, bundle product ...
The main idea is to make fixtures data customizable from the test.
You probably focused more on the factory class in the example which is not the main purpose of this proposal but simply a code isolation to make the data fixture file more readable.

I agree with you on the issue-3. But most of the time you will not need to customize all data. That's a rare situation and probably would be better in that case to create a new fixture file. I believe the data fixture should have already default data like in the example.
>In case of a big fixture, json becomes unreadable

dhorytskyi:
There is an ability to use a class method as fixture. Could you add information about how it will work in such cases?

Example of usage:
https://github.com/magento/magento2/blob/2.4-develop/dev/tests/integration/testsuite/Magento/User/Model/UserTest.php#L147
https://github.com/magento/magento2/blob/2.4-develop/dev/tests/integration/testsuite/Magento/User/Model/UserTest.php#L110

bubasuma:
@dhorytskyi 
I believe in this case the JSON will be passed as the function parameter like the following
```php
/**
     * @magentoDataFixture roleDataFixture {"role":{"name":"admin_role"}, "var1": "value2"}
     */
    public function testGetRole()
    {
    }

    public static function roleDataFixture(array $data)
    {
    }
```
Or we could unpack the data array across function parameters.  
```php
    public static function roleDataFixture(array $role, string $value)
    {
    }
```

But you will rarely need to customize function data fixture as they're written for your test only.

bubasuma:
@buskamuza @dhorytskyi 
I have updated the proposal with the following changes:

- Replace JSON format with a data provider method that returns an array of data which will be used to customize fixtures and resolve dependencies between fixtures
- Add ability to define and resolve dependencies between fixtures
- Extend _overrides.xml_ (Integration tests extensibility) to support fixture data customization
- Add an alternative solution to replace file based fixtures

PR #375, opened 2020-05-04
GraphQL B2B Company schema updates

DrewML:
# Changes

1. In `Company.id`, the `id` parameter has been changed to be non-nullable. This query only needs to do one thing well: lookup a company-associated user by ID. The schema already exposes a way to obtain the currently logged in `Customer` in other places.

2. Updated the descriptions of the root `Query` types to be a bit less verbose

3. Removed non-nullability from the "Check" types' response fields. If we add any other fields to those types later, the current field encountering an error would propagate up and infect the whole type. Can always make it non-nullable later without a breaking change

### DIFF
--- a/design-documents/graph-ql/coverage/company.md
+++ b/design-documents/graph-ql/coverage/company.md
@@ -6,11 +6,11 @@
 
 ```graphql
 type Query {
-    company: Company  @doc(description: "Returns all information about the current Company.")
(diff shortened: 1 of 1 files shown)


DrewML:
Note: @kandy suggested opening a separate PR to change naming of `check` fields

DrewML:
Consensus reached in meeting last week on the changes in this PR. We can follow up with changes to the `check` name fields (@kandy prefers naming that makes it clear that a field is a predicate query)

PR #366, opened 2020-04-01
ECP-471: Finalize B2B company schema

kokoc:
## Overview

This is the initial version of schema.graphqls that we designed to implement B2B Company management via GraphQL. This schema covers:

Company entity management: creating, getting and updating the Company information.
Managing Company Users: create, update, delete, getting the list of users with possibility of filtration by status.
Managing Company Roles: create, update, delete, getting the role data by ID or the list of all roles for the roles grid. Also the separate query for getting the list of ACL resources for Company role in nested tree format is available.
Managing Company Hierarchy and Team: create, update, delete the team, get team by ID, get the whole Company Hierarchy in nested tree format.
Also there are three queries for validating Company email, Admin email and User email for use during Company creation and User creation processes.

PR #376, opened 2020-05-05
Refactor B2B Company Structure query to remove recursive type and add more edges

DrewML:
## Problems

The schema for the [B2B Company Structure](https://docs.magento.com/m2/b2b/user_guide/customers/account-company-structure.html) has a few problems in my opinion:

1. `isEditable` seems out of place. There are other APIs in this company schema that a UI can use to determine if a user has permissions to edit the company structure.

2. The description of the `CompanyHierarchyElement.type` field is `Hierarchy element type: 'customer' or a 'team'`, but is typed as a `String`, meaning the client-side and server-side have implicit shared constants not defined in the schema

3. `CompanyHierarchyElement` does not provide any edges for a query to get access to the `Customer` or `CompanyTeam`. Instead, it decides what the UI has to show via a `text` and `description` field. If more info about the customer/company needs to be shown in the UI, it would require a second round-trip.

4. The `CompanyHierarchyElement` type is infinitely recursive via the `children` property. These are a major pain to query because the client has to dynamically build up a query, and size of the query itself grows with the tree depth. [I brought this up in the original review](https://github.com/magento/architecture/pull/366#discussion_r403141125) but it wasn't quite addressed. This is a design we've already had problems with in practice. See the linked prior discussion for details

5. No support for lazy-loading of large trees. For users on slow connections, the client-side will want the option of fetching only the first level or two of items, then load them in later (either during network idle time or when a user expands the tree).

6. Because of the recursive types, there is no foolproof way to get the entire company structure in one query. The current schema has a `max_depth` field that can be requested, but that would require 1 round trip to get the depth, and then a second once a query with that many levels has been constructed.

7. In our B2B docs we describe this as a `Company Structure`, but the field is named `hierarchy` here, which makes it more difficult to quickly Google for.

## Solutions

1. Remove `isEditable` field. Can be added back later if it's deemed necessary

2. Instead of an implicit constant via `CompanyHierarchyElement.type`, we're using a `union`, so the known values are expressed via the type system (introspectable via `__typename`).

3. Instead of `CompanyHierarchyElement.description` and `CompanyHierarchyElement.text`, the `CompanyStructureItem.entity` field will provide access to the `Customer` or `CompanyTeam` types.

4. No more recursive types for company structure

5. Renamed `Company.hierarchy` to `Company.structure`

## Requested Reviewers

- @kokoc
- @paliarush

### DIFF
--- a/design-documents/graph-ql/coverage/company.md
+++ b/design-documents/graph-ql/coverage/company.md
@@ -36,7 +36,10 @@ type Company @doc(description: "Company entity output data schema.") {
     ): CompanyRoles!  @doc(description: "Returns the list of defined roles at Company.")
     role(id: ID!): CompanyRole  @doc(description: "Returns company role by id.")
     acl_resources: [CompanyAclResource]  @doc(description: "Returns the list of all permission resources.")
-    hierarchy: CompanyHierarchyOutput  @doc(description: "Returns the complete data about company structure.")
+    structure(
+        rootID: ID = 0 @doc(description: "Tree depth to begin query")
+        depth: Int = 10 @doc(description: "Specifies how deeply results are fetched")
+    ): CompanyStructure @doc(description: "Company structure of teams and customers in depth-first order")
     team(id: ID!): CompanyTeam  @doc(description: "Returns company team data by id.")
 }
 
@@ -106,19 +109,16 @@ type CompanyEmailCheckResponse @doc(description: "Response object schema for a C
     isEmailValid: Boolean! @doc(description: "Email validation result")
 }
 
-type CompanyHierarchyOutput @doc(description: "Response object schema for a Company Hierarchy query.") {
-    structure: CompanyHierarchyElement @doc(description: "An array of Company structure elements.")
-    isEditable: Boolean @doc(description: "Flag that defines whether Company Hierarchy can be changed by current User or not.")
-    max_nesting: Int @doc(description: "Indicator of maximun nesting of elements within a whole Company Hierarchy.")
+union CompanyStructureEntity = CompanyTeam | Customer
+
+type CompanyStructureItem @doc(description: "Company Team and Customer structure") {
+    id: ID! @doc(description: "ID of the item in the hierarchy")
+    parentID: ID @doc(description: "ID of the parent item in the hierarchy")
+    entity: CompanyStructureEntity
 }
 
-type CompanyHierarchyElement @doc(description: "Company Hierarchy element output data schema.") {
-    id: ID! @doc(description: "Hierarchy element id.")
-    tree_id: ID @doc(description: "The hierarchical id of the element within a structure. Used for changing element's position in hierarchy.")
-    type: String @doc(description: "Hierarchy element type: 'customer' or a 'team'.")
-    text: String @doc(description: "Hierarchy element name.")
-    description: String @doc(description: "Hierarchy element description.")
-    children: [CompanyHierarchyElement!] @doc(description: "An array of child elements.")
+type CompanyStructure {
+    items: CompanyStructureItem[]
 }
 
 type CompanyTeam @doc(description: "Company Team entity output data schema.") {


design-documents/graph-ql/coverage/company.md
```diff
@@ -36,7 +36,10 @@ type Company @doc(description: "Company entity output data schema.") {
     ): CompanyRoles!  @doc(description: "Returns the list of defined roles at Company.")
     role(id: ID!): CompanyRole  @doc(description: "Returns company role by id.")
     acl_resources: [CompanyAclResource]  @doc(description: "Returns the list of all permission resources.")
-    hierarchy: CompanyHierarchyOutput  @doc(description: "Returns the complete data about company structure.")
+    structure(
+        rootID: ID = 0 @doc(description: "Tree depth to begin query")
```
DrewML, line 40: `rootID` and `depth` together allow fetching results to any depth, starting at any level of the tree. Useful for lazy loading

DrewML:
# Example

Provided is a sample query, a sample of the response payload, and some JS that will render a space-delimited tree. This is provided to show an example of how a front-end might render the tree in one-shot. Just imagine the string result in the JS is instead React components like `<TreeView />` and `<TreeItem depth={depth} item={item} />`:

## Example Query
```gql
    {
        company {
            structure {
                id
                parentID
                entity {
                    __typename
                    ... on Customer {
                        firstname
                        lastname
                    }
                    ... on CompanyTeam {
                        name
                    }
                }
            }
        }
    }
```
## Sample Response
```json
{
    "company": {
        "structure": [{
            "id": 3,
            "parentID": 0,
            "entity": {
                "__typename": "Customer",
                "firstname": "Hayden",
                "lastname": "Doe"
            }
        }, {
            "id": 10,
            "parentID": 3,
            "entity": {
                "__typename": "CompanyTeam",
                "name": "Company Name"
            }
        }, {
            "id": 14,
            "parentID": 10,
            "entity": {
                "__typename:" "Customer",
                "firstname": "Alex",
                "lastname": "Smith"
            }
        }, {
            "id": 12,
            "parentID": 10,
            "entity": {
                "__typename:" "Customer",
                "firstname": "Justice",
                "lastname": "Jones"
            }
        }, {
            "id": 8,
            "parentID":  3,
            "entity": {
                "__typename:" "CompanyTeam",
                "name": "Another Company Name"
            }
        }, {
            "id": 1,
            "parentID":  8,
            "entity": {
                "__typename:" "Customer",
                "firstname": "Charlie",
                "lastname": "Davis"
            }
        }]
    }
}
```

## Sample Rendering of tree UI
```js
function printCompanyStructure(structure) {
    const nodes = [];
    let lastDepth = 0;
    let lastItem;

    for (const item of structure) {
        let nextDepth = lastDepth;
        const isFirstItem = !lastItem;
        const sameLevelAsLast = lastItem && lastItem.parentID === item.parentID;
        const downALevel = lastItem && lastItem.id === item.parentID;
        const upALevel = !(isFirstItem || sameLevelAsLast || downALevel)

        if (downALevel) ++nextDepth;
        if (upALevel) --nextDepth;

        const label = item.entity.__typename === 'Customer' ?
            `${item.entity.firstname} ${item.entity.lastname}` :
            item.entity.name;

        nodes.push(''.padStart(nextDepth * 2, ' ') + label);

        lastDepth = nextDepth;
        lastItem = item;
    }

    console.log(nodes.join('\n'));
}
```

design-documents/graph-ql/coverage/company.md
DrewML: There was another suggestion from someone offline to support `rootIDs`instead of `rootID`, so you could get multiple subtrees in a single query.

Any input on that welcome. I don't _think_ it's necessary - can't think of a compelling use-case where you need 2 separate subtrees but not the parent node and other siblings.

paliarush, line 40:
```suggestion
        root_id: ID = 0 @doc(description: "Tree depth to begin query")
```

design-documents/graph-ql/coverage/company.md
```diff
@@ -106,19 +109,16 @@ type CompanyEmailCheckResponse @doc(description: "Response object schema for a C
...
 
-type CompanyHierarchyOutput @doc(description: "Response object schema for a Company Hierarchy query.") {
-    structure: CompanyHierarchyElement @doc(description: "An array of Company structure elements.")
-    isEditable: Boolean @doc(description: "Flag that defines whether Company Hierarchy can be changed by current User or not.")
-    max_nesting: Int @doc(description: "Indicator of maximun nesting of elements within a whole Company Hierarchy.")
+union CompanyStructureEntity = CompanyTeam | Customer
+
+type CompanyStructureItem @doc(description: "Company Team and Customer structure") {
+    id: ID! @doc(description: "ID of the item in the hierarchy")
+    parentID: ID @doc(description: "ID of the parent item in the hierarchy")
```
paliarush, line 116:
```suggestion
    parent_id: ID @doc(description: "ID of the parent item in the hierarchy")
```

DrewML:
per @kokoc, when removing `isEditing` field we need to circle-back and address the use-case it was intended to solve (i.e. how does the UI know which levels of the hierarchy can be mutated by the current user?).

DrewML:
> @akaplya mentioned that 2.4 introduced a new query to get the hierarchical list of categories as a flat list. @DrewML taking the TODO to try and make this PR consistent with how that query + types are defined

Looked into it, and this query doesn't really solve any of the problems we've outlined. Looks like it's for a different purpose.

See internal wiki document `Representing Trees in GraphQL` for  more info.

design-documents/graph-ql/coverage/company.md
kokoc, line 40: Seems like this field has wrong description

design-documents/graph-ql/coverage/company.md
```diff
@@ -36,7 +36,10 @@ type Company @doc(description: "Company entity output data schema.") {
     ): CompanyRoles!  @doc(description: "Returns the list of defined roles at Company.")
     role(id: ID!): CompanyRole  @doc(description: "Returns company role by id.")
     acl_resources: [CompanyAclResource]  @doc(description: "Returns the list of all permission resources.")
-    hierarchy: CompanyHierarchyOutput  @doc(description: "Returns the complete data about company structure.")
+    structure(
```
kokoc, line 39: Not something to fix for company. Do we want to have pagination in tree queries in other places like categories?

DrewML: Ideally, but think it's still open for discussion internally. See wiki doc titled `Representing Trees in GraphQL`

design-documents/graph-ql/coverage/company.md
DrewML: That's the argument description

PR #377, opened 2020-05-05
Change CompanyRole.permissions from [String] to [CompanyAclResource]

DrewML:
## Problem

When selecting from `CompanyRole.permissions`, you get a `[String]` of permission IDs. A UI may wish to capture the permission's label text in the same request.

## Solution

Change permissions to point to the appropriate type (`CompanyAclResource`)

## Requested Reviewers

@kokoc 
@paliarush


DrewML:
Consensus reached during meeting last week

PR #378, opened 2020-05-13
b2b requisition list schema

dani97:
## Problem
Coverage for B2B Requisition List

## Solution
Schema  for B2B Requisition List 

Queries
- All requisition list are available under customer query
- A single requisition list can be queried by Id
- Export Requisition List

Mutations
- Create Requisition List
- Rename Requisition List
- Add Items
- Remove Items
- Update Items
- Copy Items to cart
- Copy Items to another list
- Move Items to another list

## Requested Reviewers
@paliarush 
@lenaorobei

### DIFF
--- a/design-documents/graph-ql/coverage/requisitionList.graphqls
+++ b/design-documents/graph-ql/coverage/requisitionList.graphqls
@@ -1,11 +1,34 @@
-## Gift Card product type support will be added after the Product Implementation is developed
-type Query {
-    "The requisitionList query returns the contents of customer's  Requisition List"
-    requisitionList(
-        id: ID!
-    ): RequisitionList
+# Requistion List Flow User Flow
+#
+# 1.1. Create a requistion List at Customer's My Account section using `createRequistionList` Mutation
+#
+# 1.2. From catalog pages search for products and add products to already available list with `addProductsToRequisitionList`
+#  mutation or create a list with `createRequistionList` and on success add products with `addProductsToRequisitionList`
+#
+# 1.3. Select a requistion list from My Account and  customer query `requisitionLists` with id filter will fetch requistion
+# list details
+#
+# 1.4 Select a list of items from the requisition List and perform `addRequisitionListItemToCart`, so the items are copied
+# to cart
+#
+# 1.5 perform regular checkout flow
+#
+# 2. In requistion list view, `exportRequisitionList` query will generate a CSV file with particular requistion list data
+#
+# 3. In requistion list view, `renameRequistionList` mutatation can be used to rename the list
+#
+# 4. In requistion list view, one can edit items quantity, options etc or remove items with `removeRequisitionListItems` and
+# `updateRequisitionListItems` mutation
+#
+# 5. In requistion list view, select requistion list items and move them , copy them to different requistion list with
+# `moveItemsFromRequisitionList` and `copyItemsFromRequisitionList` mutations respectively
+#
 
-    "Export Requisition list"
+
+type Query {
+    """
+    Export Requisition list in Csv format
+    """
     exportRequisitionList(
         "unqiue Id of Requisition list"
         id: ID!
@@ -16,21 +39,26 @@ type Customer {
     "Get Requisition Lists of customer"
     requisitionLists(
         pageSize: Int = 20,
-        currentPage: Int = 1
+        currentPage: Int = 1,
+        filter: RequisitionListFilterInput
     ): RequisitionLists
 }
 
-"Provides Requisition Lists of customer"
+"""
+Provides Customer's Requisition Lists
+"""
 type RequisitionLists  {
     "List of Requisition Lists"
     items: [RequisitionList]
     "Page Information for pagination"
-    page_info: PageInfo
+    page_info: SearchResultPageInfo
     "Total count of Requisition Lists"
     total_count: Int
 }
 
-"Requisition List Type"
+"""
+Requisition List Type
+"""
 type RequisitionList {
     "Unique Identifier of Requisition List"
     id: ID!
@@ -46,17 +74,21 @@ type RequisitionList {
     updated_at: String
 }
 
-"Interface type for Requisition List Item"
+"""
+Interface type for Requisition List Item
+"""
 interface RequisitionListItemInterface {
     "Unique Identifier of Requisition List Item"
     id: ID!
-    "Product"
     product: ProductInterface!
     "Quantity added"
     qty: Float!
 }
 
-type SimpleRequisitionListItem implements RequisitionListItemInterface {
+"""
+Requisition List Item Implementation that for Simple and Virtual Products
+"""
+type DefaultRequisitionListItem implements RequisitionListItemInterface {
     "Unique Identifier of Requisition List Item"
     id: ID!
     "Product"
@@ -67,21 +99,12 @@ type SimpleRequisitionListItem implements RequisitionListItemInterface {
     customizable_options: [SelectedCustomizableOption]
 }
 
-type VirtualRequisitionListItem implements RequisitionListItemInterface {
-    "Unique Identifier of Requisition List Item"
-    id: ID!
-    "Product"
-    product: ProductInterface!
-    "Quantity added"
-    qty: Float!
-    "custom Option selected"
-    customizable_options: [SelectedCustomizableOption]
-}
-
+"""
+Requisition List Item Implementation that for Downloadable Products
+"""
 type DownloadableRequisitionListItem implements RequisitionListItemInterface {
     "Unique Identifier of Requisition List Item"
     id: ID!
-    "Product"
     product: ProductInterface!
     "Quantity added"
     qty: Float!
@@ -93,6 +116,9 @@ type DownloadableRequisitionListItem implements RequisitionListItemInterface {
     samples: [DownloadableProductSamples]
 }
 
+"""
+Requisition List Item Implementation that for Bundle Products
+"""
 type BundleRequisitionListItem implements RequisitionListItemInterface {
     "Unique Identifier of Requisition List Item"
     id: ID!
@@ -106,6 +132,9 @@ type BundleRequisitionListItem implements RequisitionListItemInterface {
     bundle_options: [SelectedBundleOption]!
 }
 
+"""
+Requisition List Item Implementation that for Configurable Products
+"""
 type ConfigurableRequisitionListItem implements RequisitionListItemInterface {
     "Unique Identifier of Requisition List Item"
     id: ID!
@@ -120,30 +149,40 @@ type ConfigurableRequisitionListItem implements RequisitionListItemInterface {
 }
 
 type Mutation {
-    "Create Empty Requisition List"
+
+    """
+    Create Empty Requisition List
+    """
     createRequisitionList(
         "name for the list"
         name: String!,
         "description For the list"
         description: String
-    ): RequisitionList
+    ): CreateRequisitionListOutput
 
-    "Rename a requisition list"
+    """
+    Rename a requisition list and change description"
+    """
     renameRequisitionList(
         "unique Id of requisition list"
         id: ID!,
         "new name for list"
         name: String!,
         "new description For the List"
-        description: String!
-    ): RequisitionList
+        description: String
+    ): RenameRequisitionListOutput
 
-    DeleteRequisitionList(
+    """
+    Delete a requisition list with Id
+    """
+    deleteRequisitionList(
         "unique Id of requisition list"
         id: ID!
-    ): RequisitionList
+    ): Boolean
 
-    "Add items to requisition list"
+    """
+    Add items to requisition list
+    """
     addProductsToRequisitionList(
         "unique Id of requisition list"
         id: ID!,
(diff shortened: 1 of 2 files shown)


mbrinton01:
Also needs to cover:
* Show 'Latest Activity' date for list. This currently shows on the Requisition List grid for each user
* Rename must include both title and description
* Delete Requisition List action
* Export list to .csv

dani97:
@mbrinton01 

- updated_at field provides Latest Activity data.
- I have made both name and description mandatory.
- Added delete list mutation

Export to CSV feature, in magento storefront is implemented with `text\csv` header, graphql doesn't support this header. To achieve this, we can do either of this ways
- generate csv file and store in pub/media folder and share the url (have to enforce security, restricted file access here)
- provide csv output in output wrapper as json encoded, the client should take care of csv file generation.
eg:-> 
type CsvOutput {
  data: String!
}

Please advise.

mbrinton01:
@dani97, please proceed with

`provide csv output in output wrapper as json encoded, the client should take care of csv file generation.`

dani97:
Thanks @mbrinton01. Will work on this and update the PR.

dani97:
@mbrinton01 added export functionality

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
+## Gift Card product type support will be added after the Product Implementation is developed
+type Query {
+    "The requisitionList query returns the contents of customer's  Requisition List"
+    requisitionList(
```
paliarush, line 4: Should any requisition-related queries be accessible by the guest users? If not, they should be moved under customer.

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
+## Gift Card product type support will be added after the Product Implementation is developed
+type Query {
+    "The requisitionList query returns the contents of customer's  Requisition List"
```
paliarush, line 3: Please use `@doc(description: "")` format for documenting the fields and types. 
Is there a benefit of using current approach, will the description be rendered properly in tools like GraphiQL?

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+    renameRequisitionList(
+        "unique Id of requisition list"
+        id: ID!,
+        "new name for list"
+        name: String!,
+        "new description For the List"
+        description: String!
+    ): RequisitionList
+
+    DeleteRequisitionList(
```
paliarush, line 141: It does not make sense to return removed requisition list in the response

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+    product: ProductInterface!
+    "Quantity added"
+    qty: Float!
+    "custom Option selected"
+    customizable_options: [SelectedCustomizableOption]
+    "Configurable options selected"
+    configurable_options: [SelectedConfigurableOption]
+}
+
+type Mutation {
```
paliarush, line 122: Please wrap all mutation responses. See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/extensibility.md

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+        "unique Id of requisition list"
+        id: ID!,
+        "Items to be updated from requisition list"
+        items: [UpdateRequisitionListItemsInput!]!
+    ): UpdateRequisitionListItemsOutput
+
+    "Add Requisition List Items To Cart"
+    addRequisitionListItemToCart(
+        "Cart Id to which requisition list needs to copied"
+        cart_id: String!,
```
paliarush, line 173: If requisition lists are only supported by registered customers, `cart_id` can be calculated based on the auth token.

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+        "unique Id of requisition list"
+        requisition_list_id: ID!,
+        "selected requisition list items that are to be added"
+        item_ids: [ID!]!
+    ): AddRequisitionListItemToCartOutput
+
+    "Copy Items from Requisition List"
+    copyItemsFromRequisitionList(
```
paliarush, line 181: From the name of the mutation it is not clear where the items will be copied to.

paliarush, line 181: Does it make sense to have copy/move functionality on the server side? It should be possible to implement it on the client side using other mutations and queries.

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
+## Gift Card product type support will be added after the Product Implementation is developed
+type Query {
```
paliarush, line 2: Would be nice to have sequence of queries/mutations for the specific scenarios documented. That would help to ensure completeness and compatibility of all operations. Additionally that would serve as a good starting point for new automated tests.

design-documents/graph-ql/coverage/requisitionList.graphqls
mbrinton01: Requisition lists are only available for customer accounts, not guests

design-documents/graph-ql/coverage/requisitionList.graphqls
DrewML: > Is there a benefit of using current approach

It's the way documentation is spec'd to be written in the IDL:

image

http://spec.graphql.org/June2018/#sec-Descriptions

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+    "The requisitionList query returns the contents of customer's  Requisition List"
+    requisitionList(
+        id: ID!
+    ): RequisitionList
+
+    "Export Requisition list"
+    exportRequisitionList(
+        "unqiue Id of Requisition list"
+        id: ID!
+    ): CsvOutput
```
DrewML, line 12: re: `CsvOutput`, I'm not quite convinced that it makes sense to stick a file download inside of what's basically a `String` type.

Doing some quick googling around, it seems like a common pattern for file downloads in GraphQL is to return a URL that's valid with the current session for 1 download.

Then the UI can use `<a href=${downloadLink} download>Download CSV</a>`, or `window.open(downloadLink)`

design-documents/graph-ql/coverage/requisitionList.graphqls
DrewML, line 4: What's the difference between `Query.requisitionList` and `Customer.requisitionList`?

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+    qty: Float!
+}
+
+type SimpleRequisitionListItem implements RequisitionListItemInterface {
+    "Unique Identifier of Requisition List Item"
+    id: ID!
+    "Product"
+    product: ProductInterface!
+    "Quantity added"
+    qty: Float!
```
DrewML, line 65: Why `Float` for quantity, as opposed to `Int`?

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+interface RequisitionListItemInterface {
+    "Unique Identifier of Requisition List Item"
+    id: ID!
+    "Product"
+    product: ProductInterface!
+    "Quantity added"
+    qty: Float!
+}
+
+type SimpleRequisitionListItem implements RequisitionListItemInterface {
```
DrewML, line 59: `SimpleRequisitionListItem` and `VirtualRequisitionListItem` look to have identical structures - is there a need for the UI to know that there's a different between a "Simple" and a "Virtual"?

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+    "Quantity added"
+    qty: Float!
+    "custom Option selected"
+    customizable_options: [SelectedCustomizableOption]
+}
+
+type VirtualRequisitionListItem implements RequisitionListItemInterface {
+    "Unique Identifier of Requisition List Item"
+    id: ID!
+    "Product"
```
DrewML, line 73: For cases like this where the description is identical to the name of the field, we can just remove the description

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+
+type MoveItemsFromRequisitionListOutput {
+    "Source Requisition List"
+    source : RequisitionList
+    "Destination Requisition List"
+    destination : RequisitionList
+}
+
+"PageInfo provides navigation for the query response"
+type PageInfo {
```
DrewML, line 254: There's already a type for representing paging with an identical structure, see `SearchResultPageInfo`

design-documents/graph-ql/coverage/requisitionList.graphqls
DrewML: Looks like this was discussed in https://github.com/magento/architecture/pull/378#issuecomment-629267899

@mbrinton01 would prefer not to inline file downloads into query responses. If we don't have the functionality to support file downloads, my preference would be to defer implementing export until we have a better solution.

design-documents/graph-ql/coverage/requisitionList.graphqls
```diff
@@ -0,0 +1,261 @@
...
+    ): RequisitionList
+
+    "Rename a requisition list"
+    renameRequisitionList(
+        "unique Id of requisition list"
+        id: ID!,
+        "new name for list"
+        name: String!,
+        "new description For the List"
+        description: String!
```
DrewML, line 138: `createRequisitionList` has `description` as optional, but `renameRequisitionList` has it as a required arg.

These should be consistent

design-documents/graph-ql/coverage/requisitionList.graphqls
DrewML: @paliarush it would be helpful to show what the alternative mutation would look like for comparison

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: @DrewML I created that with Query.requisitionList for fetching a single req list entity and Customer.requisitionLists as list of all req list. Will refactor this with filters added to Customer.requisitionLists

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: @DrewML  to support fractional quantities feature of magento I made that as float,it is already designed it cart that way too.

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: sure I will use that.

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: Checked with luma storefront, description is optional, shall I make it optional. @mbrinton01

design-documents/graph-ql/coverage/requisitionList.graphqls
DrewML: Didn't know that was a feature 😂 Thanks for the tip!

nrkapoor:
All - Can we try to merge the PR as soon as possible?

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: sure I will rename it as DefaultRequisitionListItem and merge the types with proper documentaion added

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: @paliarush requisition list items are just like cart items, along with product they can have custom options filled etc, and that data would be sometimes available only on server side. So moving / copying on server side would be the ideal solution, even if its possible.

design-documents/graph-ql/coverage/requisitionList.graphqls
mbrinton01: We can defer the export. It's not the most critical part of this feature and can be added in the future without breaking changes.

design-documents/graph-ql/coverage/requisitionList.graphqls
mbrinton01: It is optional and should be that way in all actions/places.

design-documents/graph-ql/coverage/requisitionList.graphqls
dani97: removed export query from the schema document.

design-documents/graph-ql/coverage/requisitionList.graphqls
paliarush: Unfortunately this syntax is not supported by our PHP GraphQL Server and for now we must use `@doc(description: "")` syntax. 
@DrewML feel free to create a ticket to implement support for `" "` and `""" """` syntax. We can then discuss and prioritize it.

@dani97 please change the doc syntax to the supported format.

dani97: @paliarush updated to `@doc`  syntax. :+1:

PR #379, opened 2020-05-18
MC-34395: auth annotation

cpartica:
## Problem

PWA client needs to know what queries require authentication

## Solution

Add @auth annotation to schema

## Requested Reviewers

@paliarush @zetlen

### DIFF
--- a/design-documents/graph-ql/framework/auth-annotation.md
+++ b/design-documents/graph-ql/framework/auth-annotation.md
@@ -27,7 +27,7 @@ enum UserRole {
 type Query {
     cart: Cart! @auth(requires:ANONYMOUS_USER)
     customerCart: Cart! @auth(requires:REGISTERED_CUSTOMER)
-    customer: Customer @auth(requires:[REGISTERED_CUSTOMER, COMPANY_USER])
+    company: Company @auth(requires:COMPANY_USER)
 }
 ```
 This can be hardcoded in the schema, or we could introduce an annotation to resolver level classes that could be read automatically
--- a/design-documents/graph-ql/framework/auth-annotation.md
+++ b/design-documents/graph-ql/framework/auth-annotation.md
@@ -1,6 +1,6 @@
 # Overview
 
-In the scope of adding features to the the GraphQL schema, a client like PWA needs to know what queries need authentication.
+In the scope of adding features to the GraphQL schema, a client like PWA needs to know what queries need authentication.
 This is needed to know to refresh a token each time it expires, so this logic or what queries need a token doesn't get hardcoded into the client side.
 It is also useful to implement the "soft" auth mode for cart, when a token expires, and the client would still store some info from the last sesson client side.
 
--- a/design-documents/graph-ql/framework/auth-annotation.md
+++ b/design-documents/graph-ql/framework/auth-annotation.md
@@ -1,0 +1,59 @@
+# Overview
+
+In the scope of adding features to the the GraphQL schema, a client like PWA needs to know what queries need authentication.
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/framework/auth-annotation.md
```diff
@@ -0,0 +1,62 @@
+# Overview
+
+In the scope of adding features to the GraphQL schema, a client like PWA needs to know what queries need authentication.
+This is needed to know to refresh a token each time it expires, so this logic or what queries need a token doesn't get hardcoded into the client side.
+It is also useful to implement the "soft" auth mode for cart, when a token expires, and the client would still store some info from the last sesson client side.
+
+## Annotation
+
+We need add an anotatoin like `@auth` to support metadata that would be needed by the client:
```
paliarush, line 9: How about something that will allow extension for B2B use cases, and potentially admin use cases like staging preview:

```
directive @auth(
  requires: UserRole = ANONYMOUS_USER,
) on OBJECT | FIELD_DEFINITION

enum UserRole {
  REGISTERED_CUSTOMER
  ANONYMOUS_USER
  COMPANY_USER
  ADMIN_USER
}
```
With such schema we can also add additional argument to the `@auth` directive listing required B2B or Admin ACL permissions.

cpartica: ok, I'll update the document

DrewML:
> PWA client needs to know what queries require authentication

Can we include an example use-case?

DrewML:
Would love to see prior art, i.e examples of how other platforms are handling a mix of authenticated and anonymous queries

design-documents/graph-ql/framework/auth-annotation.md
```diff
@@ -0,0 +1,62 @@
...
+
+```graphql
+type Query {
+    customerCart: Cart! @auth(required:true)
```
DrewML, line 13: How do you envision the client-side using this directive? Is it primarily there as documentation, or is the client expected to do some magic with it?

zetlen:
We originally asked for this because the PWA UX guidelines require that login state behave much like traditional sessions:

1. Session timeout resets on every request, so it never expires while the user is actively using the site
2. An expired session still connects to a user (warm auth) and allows segmentation

Since the PWA's thin backend doesn't create a session, it uses customer auth tokens alone to manage login. Those tokens always expire after a certain TTL, whether they're still being used or not. This has the unfortunate effect of unexpectedly logging a customer out, after an hour of using the site.

Our first idea was to let sessions expire and solve (2) above by using cached user data to "simulate" a logged-in state. However, that meant removing the expired token and popping a login only when the user tries to access a privileged value in the graph. Therefore, we wanted to optimize our fake-warm-auth experience by knowing ahead of time which queries would require login. That's why we proposed the @auth schema directive.

**However**, after speaking with @cpartica and @nrkapoor a bit more, we really need to solve (1) above by adding refresh tokens and readable TTLs. Right now, users get unexpectedly logged out; it's a high-priority, high-severity issue, and once it's solved, we don't really need the auth directive anymore.

DrewML:
> However, after speaking with @cpartica and @nrkapoor a bit more, we really need to solve (1) above by adding refresh tokens and readable TTLs. Right now, users get unexpectedly logged out; it's a high-priority, high-severity issue, and once it's solved, we don't really need the auth directive anymore.

Can we get clarification about what we're looking to proceed with? Not totally clear whether the ask is for `@auth` until token refresh exists, or if the ask if to build the token refresh functionality.

DrewML:
Any updates @zetlen @cpartica?

cpartica:
yes, we close this out, it's not needed for the time beng

PR #380, opened 2020-05-19
Add configured_variant to ConfigurableCartItem

danielrenaud:
## Problem
When querying cart_items on a cart, the parent product is returned for configurable cart items. This is because ConfigurableCartItem.product corresponds to the parent product. In some cases it is desired to have the configured child product.
One use case is that the product's image is the parent product's image, so it doesn't match the configured selection.

## Solution
The proposed solution is to add an additional field to ConfigurableCartItem, that will contain the simple product that is the configured variant.

## Requested Reviewers
@paliarush 
@akaplya

### DIFF
--- a/design-documents/graph-ql/coverage/Cart.graphqls
+++ b/design-documents/graph-ql/coverage/Cart.graphqls
@@ -50,7 +50,7 @@ type VirtualCartItem implements CartItemInterface @doc(description: "Virtual Car
 }
 
 type ConfigurableCartItem implements CartItemInterface {
-    child_sku: String! @doc(description: "SKU of the simple product corresponding to a set of selected configurable options.")
+    configured_variant: SimpleProduct! @doc(description: "Simple product corresponding to configured product.")
     configurable_options: [SelectedConfigurableOption!]
 }
 
--- a/design-documents/graph-ql/coverage/Cart.graphqls
+++ b/design-documents/graph-ql/coverage/Cart.graphqls
@@ -51,7 +51,8 @@ type VirtualCartItem implements CartItemInterface @doc(description: "Virtual Car
 
 type ConfigurableCartItem implements CartItemInterface {
     configured_variant: SimpleProduct! @doc(description: "Simple product corresponding to configured product.")
-    configurable_options: [SelectedConfigurableOption!]
+    configurable_options: [SelectedConfigurableOption!] @deprecated(reason: "use configured_options")
+    configured_options: [SelectedConfigurableOption!] @doc(description: "Configured options for product")
 }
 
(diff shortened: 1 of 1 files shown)


danielrenaud:
@paliarush  and @akaplya  Can you please review this PR?
It is necessary for addressing this issue: https://github.com/magento/magento2/issues/28573

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -50,7 +50,7 @@ type VirtualCartItem implements CartItemInterface @doc(description: "Virtual Car
 }
 
 type ConfigurableCartItem implements CartItemInterface {
-    child_sku: String! @doc(description: "SKU of the simple product corresponding to a set of selected configurable options.")
```
DrewML, line 53: Is `child_sku` in any released version of Magento?

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -50,7 +50,7 @@ type VirtualCartItem implements CartItemInterface @doc(description: "Virtual Car
...
+    configured_variant: SimpleProduct! @doc(description: "Simple product corresponding to configured product.")
```
DrewML, line 53: I think we can unify the field name prefixes in `ConfigurableCartItem` a bit more. Having both `configured_` and `configurable_` prefixes representing some user state doesn't seem ideal.

design-documents/graph-ql/coverage/Cart.graphqls
danielrenaud: `child_sku` is not currently available in the schema. I suggest we return the product itself instead of just the sku. With only the child_sku, it will require an additional query, and likely it won't be possible to fetch the child product anyway

DrewML: > child_sku is not currently available in the schema. I suggest we return the product itself instead of just the sku. With only the child_sku, it will require an additional query, and likely it won't be possible to fetch the child product anyway

Oh I 100% agree - was just trying to figure out if we were introducing a breaking change or not

DrewML: > I suggest we return the product itself instead of just the sku.

We should pretty much always try to do this - anytime we return an ID referencing an entity we're baking in a 2-request flow for clients

DrewML:
Besides the nitpicky comments on field naming, the proposed change looks good to me 👍

design-documents/graph-ql/coverage/Cart.graphqls
danielrenaud: Agreed. I think `configurable_options` should be `configured_options`. In this we will have to deprecate and add a replacement field. I've updated the schema

PR #382, opened 2020-06-01
Customer order schema updates

danielrenaud:
## Problem
1. SalesItemInterface is not being used properly as an interface
2. Unclear how to return parent and child items in Orders/Invoices for bundle and configurable products
3. Some shipment fields did not align with available data

## Solution
1. Remove SalesItemInterface and instead use unique types for each item type (OrderItem, InvoiceItem, etc) that have only relevant fields and an ID link to the associated order item
2. For items that distinguish parent and child items (e.g. OrderItem and InvoiceItem) for complex products we introduce an item interface (e.g. OrderItemInterface) so that complex product types can add fields (e.g. BundleOrderItem)
3. Remove ShipmentTracking.link; change ShipmentTracking.method to ShipmentTracking.title because available data does not align with shipping method

Other changes:
Add `comments` field to Order and Invoice

## Requested Reviewers
@joni-jones 
@akaplya 
@paliarush

### DIFF
--- a/design-documents/graph-ql/coverage/customer-orders.md
+++ b/design-documents/graph-ql/coverage/customer-orders.md
@@ -153,9 +153,10 @@ type OrderTotal implements SalesTotalAmountInterface {
 @doc("Shipping handling details")
 type ShippingHandling {
     total_amount: Money! @doc("shipping total amount")
-    amount_inc_tax: Money @doc("shipping amount including tax")
-    amount_exc_tax: Money @doc("shipping amount excluding tax")
+    amount_including_tax: Money @doc("shipping amount including tax")
+    amount_excluding_tax: Money @doc("shipping amount excluding tax")
     taxes: [TaxItem] @doc("shipping taxes details")
+    discounts: [Discount] @doc(description: "The applied discounts to the shipping)
 }
 
 @doc("Tax item details")
--- a/design-documents/graph-ql/coverage/customer-orders.md
+++ b/design-documents/graph-ql/coverage/customer-orders.md
@@ -268,7 +268,7 @@ type ShipmentTracking {
     number: String @doc("tracking number of the order shipment")
 }
 ```
-### Comments
+## CommentItem type
 ```graphql
 type CommentItem {
     timestamp: String! @doc("The timestamp of the comment")
--- a/design-documents/graph-ql/coverage/customer-orders.md
+++ b/design-documents/graph-ql/coverage/customer-orders.md
@@ -176,11 +176,12 @@ type Invoice {
     id: ID! @doc("the ID of the invoice, used for API purposes")
     number: String! @doc("sequential invoice number")
     total: InvoiceTotal! @doc("invoice total amount details")
-    items: [InvoiceItem]! @doc("invoiced product details")
+    items: [InvoiceItemInterface]! @doc("invoiced product details")
+    comments: [InvoiceComment]
 }
 
 @doc("Invoice item details")
-type InvoiceItem {
+type InvoiceItemInterface {
     id: ID! @doc("invoice item unique identifier") #base64encode(invoiceItemId)
     order_item_id: String @doc("link to order item")
     product_name: String @doc("name of the base product")
@@ -189,6 +190,18 @@ type InvoiceItem {
     product_sale_price: Money! @doc("sale price for the base product including selected options")
     discounts: [Discount] @doc("final discount information for the base product including discounts on options")
     quantity_invoiced: Float @doc("number of invoiced items")
+}
+
+type InvoiceItem implements InvoiceItemInterface {
+}
+
+type BundledInvoiceItem implements InvoiceItemInterface {
+    child_items: [InvoiceItemInterface]
+}
+
+type InvoiceComment {
+    timestamp: String!
+    message: String
 }
 
 @doc("Invoice total amount details")
--- a/design-documents/graph-ql/coverage/customer-orders.md
+++ b/design-documents/graph-ql/coverage/customer-orders.md
@@ -60,7 +60,7 @@ type CustomerOrder {
     order_date: String! @doc("date when the order was placed")
     status: String! @doc("current status of the order")
     number: String! @doc("sequential order number")
-    items: [OrderItem]! @doc("collection of all the items purchased")
+    items: [OrderItemInterface]! @doc("collection of all the items purchased")
     total: OrderTotal! @doc("total amount details for the order")
     invoices: [Invoice]! @doc("invoice list for the order")
     credit_memos: [CreditMemo]! @doc("credit memo list for the order")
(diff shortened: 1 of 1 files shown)


joni-jones:
I would suggest adding `CommentItem` also for credit memo and shipment types.

danielrenaud:
> I would suggest adding `CommentItem` also for credit memo and shipment types.

For CreditMemo and Shipment, there is ability to add comment on backend but no option to make it visible on storefront. So seems like we shouldn't show it on storefront

danielrenaud:
Made additional changes.
Added comments to CreditMemo and OrderShipment types
Change CustomerOrder.carrier, method, and shipping_address to be nullable because it is possible for these to be empty for an order (e.g. order with only virtual products)

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -60,16 +60,17 @@ type CustomerOrder {
...
     shipments: [OrderShipment]! @doc("shipment list for the order")
     payment_methods: [PaymentMethod]! @doc("payment details for the order")
-    shipping_address: CustomerAddress! @doc("shipping address for the order")
+    shipping_address: CustomerAddress @doc("shipping address for the order")
     billing_address: CustomerAddress! @doc("billing address for the order")
-    carrier: String! @doc("shipping carrier for the order delivery")
-    method: String! @doc("shipping method for the order")
+    carrier: String @doc("shipping carrier for the order delivery")
+    method: String @doc("shipping method for the order")
+    comments: [CommentItem]! @doc("comments on the order")
```
paliarush, line 73: Let's not mark comments with `!`. Our latest vision is to not mark complex fields that may have their own resolver with `!` even if they are technically always present.
The reason is that it will allow for better exception handling (exception will not bubble up to the nearest nullable field), and we can always add `!` in the future to the output types (cannot remove though). 

Please check all complex fields in output types across this new schema.

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -79,55 +80,40 @@ The `id` will be a `base64_encode(increment_id)` which in future can be replaced
...
     quantity_shipped: Float @doc("number of shipped items")
     quantity_refunded: Float @doc("number of refunded items")
     quantity_invoiced: Float @doc("number of invoiced items")
     quantity_canceled: Float @doc("number of cancelled items")
     quantity_returned: Float @doc("number of returned items")
-    status: String @doc("the status of order item")
 }
-```
 
+type OrderItem implements OrderItemInterface {
```
DrewML, line 104:
```
type OrderItem implements OrderItemInterface {
}
```

This is invalid according to the GraphQL spec

image

Source: http://spec.graphql.org/June2018/#sec-Interfaces

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -189,19 +175,34 @@ The invoice entity will have the similar to the order schema:
...
+    items: [InvoiceItemInterface] @doc("invoiced product details")
+    comments: [CommentItem] @doc("comments on the invoice")
 }
 
 @doc("Invoice item details")
-type InvoiceItem implements SalesItemInterface{
-    quantity_invoiced: Float! @doc("number of invoiced items")
+type InvoiceItemInterface {
+    id: ID! @doc("invoice item unique identifier") #base64encode(invoiceItemId)
+    order_item_id: String @doc("link to order item")
```
DrewML, line 186: Is there a reason to prefer `order_item_id` here and not something like `OrderItemInterface`.

Typically in a GraphQL schema, anywhere you _would_ put a foreign key to an entity, it would make more sense to allow selecting that entity. Otherwise, it can force the client-side into needing multiple round trips for the data

DrewML:
Do you have any screenshots of what the UI for this data looks like (or plans to look like)? Would make review easier

design-documents/graph-ql/coverage/customer-orders.md
danielrenaud: The interface does define one or more fields. The type that implements the interface does not add any additional fields. That statement isn't 100% clear to me so maybe I got it wrong

danielrenaud:
@DrewML In Luma if you login as a customer and go to "My Orders" (place some orders first), that is the UI that this is roughly based on.

design-documents/graph-ql/coverage/customer-orders.md
danielrenaud: It seems preferable to return an id for use as a reference rather than returning the object here. We don't know exactly how this will be used, but it seems like it would not be preferable to query the order items on the order level and then also on each level of InvoiceItems, ShipmentItems, etc. However it does seem likely that having this reference would be useful, though I don't have concrete examples.

design-documents/graph-ql/coverage/customer-orders.md
DrewML: Sorry, meant to link the validation for Object types.

image

http://spec.graphql.org/June2018/#sec-Objects

Have you tested this works with webonyx? If so it sounds like there's a spec-compliancy issue.

DrewML: Here's how the GraphQL reference implementation from the GraphQL foundation does it:

image

Note that validation is done in the same order as the spec. So checks for fields are done _before_ checking if the interface is respected.

DrewML:
> @DrewML In Luma if you login as a customer and go to "My Orders" (place some orders first), that is the UI that this is roughly based on.

Sure! I'm just asking that we include some screenshots since there will be a number of reviewers

design-documents/graph-ql/coverage/customer-orders.md
DrewML: If someone didn't want the whole object, wouldn't they just not query for it though?

design-documents/graph-ql/coverage/customer-orders.md
DrewML: (I bring this up because we won't be able to implement this schema in the new GraphQL server)

design-documents/graph-ql/coverage/customer-orders.md
danielrenaud: We can return the whole item.
@joni-jones is that alright with you? I know you had concern with that complicating the schema

design-documents/graph-ql/coverage/customer-orders.md
danielrenaud: The schema stitching will copy the interface fields into the type so we don't need to redeclare all interface fields on the type. This is the approach used in other schema.

danielrenaud:
Here is a collection of screenshots from the UI we are referencing for this schema.

Screen Shot 2020-06-02 at 2 47 59 PM
Screen Shot 2020-06-02 at 2 48 13 PM
Screen Shot 2020-06-02 at 2 48 22 PM
Screen Shot 2020-06-02 at 2 48 31 PM
Screen Shot 2020-06-02 at 2 48 41 PM

design-documents/graph-ql/coverage/customer-orders.md
joni-jones: I don't know the real use case where order item data is needed in the scope of invoice/shipment/credit memo item. The `order_item_id` would help to referer to the original order item if it needed (some custom logic to show which order items shipped/invoiced/refunded), I don't think the whole data would be useful.

DrewML:
Thanks @danielrenaud - that's super helpful!

design-documents/graph-ql/coverage/customer-orders.md
DrewML: @paliarush also pointed this out to me in another PR 👍. Thanks for clarifying

design-documents/graph-ql/coverage/customer-orders.md
DrewML: What about if a UI dev wants to make an `a` tag pointing to the URL of the product? Wouldn't a second round trip be needed for `product_url_key`?

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -216,13 +217,20 @@ The credit memo entity will have the similar to the order and invoice schema:
 type CreditMemo {
     id: ID! @doc("the ID of the credit memo, used for API purposes")
     number: String! @doc("sequential credit memo number")
-    items: [CreditMemoItem]! @doc("items refunded")
-    total: CredtiMemoTotal! @doc("refund total amount details")
+    items: [CreditMemoItem] @doc("items refunded")
+    total: CredtiMemoTotal @doc("refund total amount details")
```
DrewML, line 221:
```suggestion
    total: CreditMemoTotal @doc("refund total amount details")
```

design-documents/graph-ql/coverage/customer-orders.md
prabhuram93: @DrewML @joni-jones I do see a need to refer back to the Order Item data for getting the selected and entered options. based on the above screenshot as well(" 1 X Sprite..") which the InvoiceItemInterface doesn't offer.

danielrenaud: Discussed pros and cons of this offline. We decided to return order_item instead of order_item_id, it may introduce complexity and performance implications on frontend. But we prioritize our API's alignment with standard GraphQL approach.

danielrenaud:
changed `order_item_id: String` to `order_item: OrderItemInterface` for InvoiceItem /CreditMemoItem / ShipmentItem

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -79,55 +80,40 @@ The `id` will be a `base64_encode(increment_id)` which in future can be replaced
...
-@doc("Interface to reprent order/invoice/shipment/credit memo items")
-interface SalesItemInterface {
+@doc("Order item")
+type OrderItemInterface {
+    id: ID! @doc("Order item unique identifier") #base64encode(orderItemId)
     product_name: String @doc("name of the base product")
     product_sku: String! @doc("SKU of the base product")
-    product_url: String @doc("URL of the base product")
+    product_url_key: String @doc("URL key of the base product")
+    product_type: String @doc("Type of product (e.g. simple, configurable, bundle)")
```
DrewML, line 90: Discussed offline, recording here for posterity:

@danielrenaud and I discussed a few things about `product_type` offline:

1. We talked about a similar field being deprecated recently in core, [`ProductInterface.type_id`](https://github.com/magento/magento2/blob/7f2272dcbaccde66519f7d1b3c81a633ff539b3c/app/code/Magento/CatalogGraphQl/etc/schema.graphqls#L108), with the recommendation to use `__typename` at runtime. We're not returning the full entity type here though (since order items are a limited, historical snapshot of the item), so that doesn't make much sense.

1. We discussed the possibility of making this an `enum`, rather than the wider `String` type. That would help a client understand what values to handle, but it sounds like Magento doesn't have a static list of product types (is extensible).

design-documents/graph-ql/coverage/customer-orders.md
```diff
@@ -79,55 +80,40 @@ The `id` will be a `base64_encode(increment_id)` which in future can be replaced
 
 ### Order Item
 
-The order items will be presented as separate interface which will have multiple implementations for invoice, shipment and credit memo types.
-
 ```graphql
-@doc("Interface to reprent order/invoice/shipment/credit memo items")
-interface SalesItemInterface {
+@doc("Order item")
+type OrderItemInterface {
```
DrewML, line 85: The name here should probably just be `OrderItem` since it's a `type`, right? Think it's just left over from a refactor

danielrenaud: Oh no, it should be interface instead of type, OrderItem type is defined below

DrewML: Ah right! Thanks

DrewML:
We have approvals across the board on this, so I'm going to go ahead and merge.

@cpartica @danielrenaud if we're still iterating please feel free to open new PRs and we'll try to make it through them fast.

PR #383, opened 2020-06-01
GraphQL gift registry schema

paliarush:
In addition to creation of gift registry schema, some changes in cart were required to allow moving items from cart to gift registry (https://github.com/magento/architecture/pull/383/commits/a6404a455b34f8a616a050cfed5cff3586d9a018, https://github.com/magento/architecture/pull/383/commits/e3f8ea6514825d76c22ddbb0882bc805e1ddb044):
- Added customizable_options to CartItemInterface
- Added id_v2: ID to DownloadableProductLinks to identify specific link and be able to add it to gift registry
- Added child_sku to ConfigurableCartItem, need to be able to add to gift registry
- Making sure client does not have to generate hash of the option. It should be available to move items from cart to registry and wishlist
 - Fixed code style for the query/mutation arguments (https://github.com/magento/architecture/pull/383/commits/17ef7fe6831f92d494ad8a614879b9ce18d2fc05)
- Moved child_sku from bundle option to bundle option value (https://github.com/magento/architecture/pull/383/commits/fde0da37edea8bc79ee451d376ee530c9b2a7933)

Extended wishlist schema (https://github.com/magento/architecture/pull/383/commits/93eead7fbea6e98491e0534e1f7eff5d692f8ca4):
 - Extended Wishlist schema to provide information about the selected options
 - Fixed code style for the query/mutation arguments (https://github.com/magento/architecture/pull/383/commits/17ef7fe6831f92d494ad8a614879b9ce18d2fc05)

https://jira.corp.magento.com/browse/ECP-711


design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,278 @@
+type Customer {
+    gitft_registry_list: [GiftRegistry]
+    gitft_registry(id: ID!): GiftRegistry
```
DrewML, line 3: ```suggestion```
    gift_registry(id: ID!): GiftRegistry
```

File: design-documents/graph-ql/coverage/gift-registry.graphqls [hunk 2]
```diff
@@ -0,0 +1,278 @@
+type Customer {
+    gitft_registry_list: [GiftRegistry]
```
DrewML, line 2:
```suggestion
    gift_registry_list: [GiftRegistry]
```

DrewML:
Can you include screenshots of UI this should be providing coverage for?

design-documents/graph-ql/coverage/gift-registry.graphqls
DrewML, line 2: Do we typically use hungarian notation for type names? `list` seems fairly clear from the type def

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,278 @@
...
+    giftRegistrySearch(
+        registrant_firstname: String!,
+        registrant_lastname: String!,
+        gift_registry_type_id: ID,
+        searchable_dynamic_attributes: [GiftRegistryDynamicAttributeInput] @doc(description: "For select attributes ID should be provided expected. For range search, '_from' and '_to' suffixes can be added to the attribute code. For text attributes provide value for exact matching.")
+    ): [GiftRegistry] @doc(description: "Gift registry search by registrant name and additional searchable attributes.")
+}
+
+type Mutation {
+    # All mutations below should only be accessible to the registry owner
```
DrewML, line 17: > All mutations below should only be accessible to the registry owner

Shouldn't they be accessible to everyone, but throw for folks without correct authorization?

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,278 @@
...
+    createGiftRegistry(gift_registry: CreateGiftRegistryInput): CreateGiftRegistryOutput
+    updateGiftRegistry(id: ID!, gift_registry: UpdateGiftRegistryInput): UpdateGiftRegistryOutput
+    addProductsToGiftRegistry(gift_registry_id: ID!, items: [GiftRegistryItemInput!]!): AddProductsToGiftRegistryOutput
+}
+
+input GiftRegistryItemInput {
+    sku: String!
```
DrewML, line 24: Any reason for `sku` and `parent_sku` to return a `String` and not the actual concrete entity?

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,278 @@
...
+interface GiftRegistryDynamicAttributeMetadataInterface {
+    code: ID!
+    input_type: String!
+    attribute_group: String!
+    label: String!
+    is_required: Boolean!
+    sort_order: Int
+}
+
+type GiftRegistryTextAttributeMetadata implements GiftRegistryDynamicAttributeMetadataInterface {
```
DrewML, line 108: image

http://spec.graphql.org/June2018/#sec-Interfaces

DrewML, line 108: image

http://spec.graphql.org/June2018/#sec-Objects

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,278 @@
...
+    type_id: String!
+    message: String!
+    privacy_settings: GiftRegistryPrivacySettings!
+    status: GiftRegistryStatus!
+    registrants: [CreateGiftRegistryRegistrantInput!]!
+    shipping_address: GiftRegistryShippingAddressInput
+    dynamic_attributes: [GiftRegistryDynamicAttributeInput]
+}
+
+type GiftRegistryShippingAddressInput @doc(description: "Either address data or address ID should be provided") {
```
DrewML, line 56: The comment says `Either address data or address ID should be provided`, but this schema allows both to be passed. Seems like undefined behavior we should be covering

DrewML: The "should be provided" comment is also a bit confusing, because consumers are _selecting_ from this type

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,296 @@
+type Customer {
+    gitft_registry_list: [GiftRegistry]
+    gitft_registry(id: ID!): GiftRegistry
+}
+
+type Query {
+    giftRegistryTypes: [GiftRegistryType] @doc(description: "Get a list of available gift registry types")
+    giftRegistrySearch(
+        registrant_firstname: String!,
```
akaplya, line 9: Is this secure to allow searching for the gift registry by name or last name?

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,296 @@
...
+}
+
+type GiftRegistrySelectAttributeOptionMetadata {
+    code: ID!
+    label: String!
+    is_default: Boolean
+}
+
+type GiftRegistry {
+    id: ID!
```
akaplya, line 176: Is there a possibility to share a gift registry through the shareable token or URL?

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,296 @@
...
+    customizable_options: [SelectedCustomizableOption]
+}
+
+type BundleGiftRegistryItem implements GiftRegistryItemInterface {
+    customizable_options: [SelectedCustomizableOption]
+    bundle_options: [SelectedBundleOption!]
+}
+
+type ConfigurableGiftRegistryItem implements GiftRegistryItemInterface {
+    customizable_options: [SelectedCustomizableOption]
```
akaplya, line 209: in case if configurable options were set up should we point the particular SKU?

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,296 @@
...
+        gift_registry_type_id: ID,
+        searchable_dynamic_attributes: [GiftRegistryDynamicAttributeInput] @doc(description: "For select attributes ID should be provided expected. For range search, '_from' and '_to' suffixes can be added to the attribute code. For text attributes provide value for exact matching.")
+    ): [GiftRegistry] @doc(description: "Gift registry search by registrant name and additional searchable attributes.")
+}
+
+type Mutation {
+    # All mutations below should only be accessible to the registry owner
+    createGiftRegistry(gift_registry: CreateGiftRegistryInput): CreateGiftRegistryOutput
+    updateGiftRegistry(id: ID!, gift_registry: UpdateGiftRegistryInput): UpdateGiftRegistryOutput
+    addGiftRegistryItems(gift_registry_id: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput
```
akaplya, line 20: Is it possible to put  unpreconfigured bundle or configurable into the gift registry?

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: This is to differentiate from a query returning single gift registry.
Does `gift_registries` sound better to you?

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,278 @@
+type Customer {
+    gitft_registry_list: [GiftRegistry]
+    gitft_registry(id: ID!): GiftRegistry
+}
+
+type Query {
+    giftRegistryTypes: [GiftRegistryType] @doc(description: "Get a list of available gift registry types")
+    giftRegistrySearch(
```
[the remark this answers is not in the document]
paliarush: That's our current agreement, we have already discussed recently whether it makes sense to change, but seems like because backward compatibility better to keep this style. Feel free to create a separate proposal/discussion on this topic.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: This flow is currently supported by Magento. We also agreed with Nishant to not support search by email and gift registry ID for now.

@nrkapoor do you have any concerns in regards to security?

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: Exactly, they should give authorization errors, that's what I meant here. We don't have a way to hide fields from specific users.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: Gift registry is like cart (and not like wishlist), it requires all options to be preconfigured.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: This is input type, not output. In the output the actual product entity is returned.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: How would you improve it?

Also, the way it is done now is consistent with current cart schema.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: That is correct, however we have custom schema stitching in Magento GraphQL webonyx implementation, which copies all fields from interfaces to the implementation, before producing the final merged schema. This allows us to avoid duplication. So we also follow the same style in designs, then implementors can just take schema from the design document and it should work.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: Yes, see scenarios I'm aware of in another document (same PR). Schema is still in progress, will add support for this flow soon.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: See definition of `GiftRegistryItemInterface`, it has reference to product. Do we need something else?

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: Clarified the comment.

design-documents/graph-ql/coverage/gift-registry.graphqls
nrkapoor: @paliarush - we should not allow search by registry ID and email id. It does not make sense from a storefront standpoint.

design-documents/graph-ql/coverage/gift-registry.graphqls
DrewML: > Does gift_registries sound better to you?

Yeah definitely - I like that!

design-documents/graph-ql/coverage/gift-registry.graphqls
DrewML: Thanks for pointing that out!

design-documents/graph-ql/coverage/gift-registry.graphqls
DrewML: Ah gotcha - @danielrenaud had to explain this to me a few hours ago.

design-documents/graph-ql/coverage/gift-registry.graphqls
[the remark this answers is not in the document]
DrewML: Sounds good - if it's already a thing no need to discuss in this PR 👍

design-documents/graph-ql/coverage/gift-registry.graphqls
DrewML: > How would you improve it?

Very fair question! Will follow-up

design-documents/graph-ql/coverage/gift-registry.graphqls
akaplya: @nrkapoor what will be the storefront scenario we would like to support? Search by name cannot guarantee us the correctness. Can you please help with the storefront scenario we would like to support?

nrkapoor: @akaplya there is no use case for storefront for searching by ID or email. Name search may not be accurate but that is the best practice.

design-documents/graph-ql/coverage/gift-registry.graphqls
akaplya: @paliarush this is slightly unclear to me. For bundles and configurable, this query should return the selected simples, because these products assuming to be added to cart, right? But as far I understood a product in the gift registry interface could be bundle or configurable. 

What will be the flow of adding the item from the registry to the shopping cart?

My thoughts were that gift registry should provide explicit arguments that our add to cart mutation requires?

design-documents/graph-ql/coverage/gift-registry.graphqls
DrewML: (Side-note: Think you meant for `GiftRegistryShippingAddressInput` to be an `input` rather than an ObjectType definition.

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,316 @@
...
+    event_name: String!
+    type_id: String!
+    message: String!
+    privacy_settings: GiftRegistryPrivacySettings!
+    status: GiftRegistryStatus!
+    shipping_address: GiftRegistryShippingAddressInput
+    dynamic_attributes: [GiftRegistryDynamicAttributeInput]
+}
+
+type GiftRegistryShippingAddressInput @doc(description: "Either address data or address ID should be provided") {
```
DrewML, line 70:
```suggestion
input GiftRegistryShippingAddressInput @doc(description: "Either address data or address ID should be provided") {
```

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,316 @@
...
+    address_data: CustomerAddressInput
+    address_id: Int
+}
+
+type UpdateGiftRegistryRegistrantInput {
```
DrewML, line 75:
```suggestion
input UpdateGiftRegistryRegistrantInput {
```

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,316 @@
...
+    id: ID!
+    first_name: String
+    last_name: String
+    email: String
+    dynamic_attributes: [GiftRegistryDynamicAttributeInput] @doc(description: "As a result of the update, only the values of provided attributes will be affected. If the attribute is missing in the request, its value will not be changed")
+}
+
+type AddGiftRegistryRegistrantInput {
```
DrewML:
```suggestion
input AddGiftRegistryRegistrantInput {
```

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,316 @@
...
+
+type GiftRegistryRegistrant {
+    id: ID!
+    first_name: String!
+    last_name: String!
+    email: String! @doc(description: "Accessible to the registry owner only")
+    dynamic_attributes: [GiftRegistryRegistrantDynamicAttribute]
+}
+
+enum GiftRegistryStatus {
```
DrewML, line 284: I really like that we're using an enum for status instead of something like an `is_active` boolean 👍

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: I think only one should be provided. Validation error should happen if both are provided. Will improve documentation and mention about validation exception.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: @akaplya name search will give you a list of records found. It is also possible to adjust search by providing additional search criteria, like date, location etc. The list of additional searchable attribute is configured by the admin.

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: There should be a way to navigate from the registry to the configurable/bundle product page. How can this work if we only return simple product data?

Also, existing behavior is aligned with cart behavior. And there should be enough information in gift registry item data to add it to cart. The exact queries will be added soon in the use cases section of this PR.

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,321 @@
...
+    DETAILED_INFORMATION
+    SHIPPING_ADDRESS
+}
+
+type GiftRegistryRegistrant {
+    id: ID!
+    first_name: String!
+    last_name: String!
+    email: String! @doc(description: "Accessible to the registry owner only")
+    dynamic_attributes: [GiftRegistryRegistrantDynamicAttribute]
```
eduard13, line 305: Hi @paliarush, do you mind if I get some clarification here. What exactly `dynamic_attributes` are supposed to be for registrant? Or it's something that has to be extended later on?
Thank you.

design-documents/graph-ql/coverage/Wishlist.graphqls
```diff
@@ -59,10 +59,32 @@ type WishlistOutput @deprecated(reason: "Deprecated: `Wishlist` type should be u
     updated_at: String @deprecated
 }
 
-type WishlistItem {
+type WishlistItemInterface {
```
DrewML, line 62: Think you meant for this to be an `interface` right?

paliarush:
@DrewML regarding screenshots, gift registry UI is mostly described here https://docs.magento.com/user-guide/marketing/gift-registries.html

design-documents/graph-ql/coverage/gift-registry.graphqls
paliarush: Please take a look at screenshots here or play with gift registry in the admin: https://docs.magento.com/user-guide/marketing/gift-registry-create.html

Gift registry has its own system of dynamic attributes, which is not EAV. 

Please let me know if this does not answer your question.

eduard13: I didn't pay attention to the `Attribute Group`.
Thank you.

design-documents/graph-ql/coverage/gift-registry.graphqls
```diff
@@ -0,0 +1,324 @@
...
+type GiftRegistryRegistrantDynamicAttribute implements GiftRegistryDynamicAttributeInterface {
+
+}
+
+enum GiftRegistryDynamicAttributeGroup {
+    GENERAL_INFORMATION
+    EVENT_INFORMATION
+    PRIVACY_SETTINGS
+    DETAILED_INFORMATION
+    SHIPPING_ADDRESS
```
eduard13, line 300: Hi @paliarush one more question regarding the Attribute Groups. Currently we have the:
- Event Information
- Gift Registry Details
- Privacy Settings
- Shipping Address

So, I'm not sure what's the correct mapping here:
- Event Information -> `EVENT_INFORMATION`
- Gift Registry Details -> `DETAILED_INFORMATION` or `GENERAL_INFORMATION`
- Privacy Settings -> `PRIVACY_SETTINGS`
- Shipping Address -> `SHIPPING_ADDRESS`

paliarush: @eduard13 thanks for the question, please check the screenshot here https://docs.magento.com/user-guide/marketing/gift-registries.html
Try to configure all possible attributes with different attribute groups and see how they are displayed on the storefront, it is possible that current UX is not optimal and need to be improved, feel free to suggest how to improve it.

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -11,22 +11,134 @@ type CartQueryOutput {
...
-    gift_cards_amount_used: Money
-    applied_gift_cards: [CartGiftCard]
+type BundleCartItem implements CartItemInterface {
+    bundle_options: [SelectedBundleOption!]
+}
 
-    is_multishipping: Boolean!
-    is_virtual: Boolean!
+type GiftCardCartItem implements CartItemInterface {
+    sender_name: String!
```
rogyar, line 68: There might be an issue from the implementation perspective. Since this type implement interface with type resolver, most likely we will not be able to create resolver for `GiftCardCartItem` to return all flat values like `sender_name`, `recepient_name` and `message`. As a result, we might need to create a separate resolver for every mentioned field. This fact might be an issue from the performance perspective. 

Maybe we should consider introducing some intermediate entity like

```graphql
type GiftCardCartItemOptions (@resolver ...) {
  sender_name: String!
  recepient_name: String!
  message: String
}

...

type GiftCardCartItem implements CartItemInterface { 
  gift_cart_options: GiftCardCartItemoptions
  amount: SelectedGiftCardAmount
}
```

In that way, we may create just one resolver for `GiftCardCartItemoptions` instead of resolving every field separately. What do you think? 

Thank you.

paliarush: I would really prefer to keep schema-first approach. How about introducing some data provider with internal state and reusing it from all resolvers to improve performance?

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -11,22 +11,134 @@ type CartQueryOutput {
...
+    recepient_name: String!
+    amount: SelectedGiftCardAmount
+    message: String
+}
+
+type SelectedConfigurableOption {
+    id_v2: ID! @doc(description: "Hash which includes option ID, option type and slected values. Can be used to move configurable product to wishlist or gift registry")
```
DrewML, line 75: Can you explain the reason for both `id` and `id_v2` here? Not super clear to me

DrewML, line 75: Tiny nitpick on the description here: the fact that it's a hash as an implementation detail the client doesn't/shouldn't know about. IDs should be opaque values to the client

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: Discussed with the team working on the implementation, agreed not to introduce wrapper.

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: `id` was there before and is actual integer value of the option ID.

`id_v2` is a hash which contains enough information to specify this option selection when adding product to cart/wishlist/gift registry. Data which may be contained here (depending on the option):
- Option type: custom/bundle/configurable/gift card amount etc
- Option ID
- Selected option value
- Maybe, entered value for the gift card (this is not finalized yet)

paliarush: I would prefer to keep it because this is important for the use case: read from here and use when adding items to wishlist or gift registry.
It was also supposed to serve as a hint for implementors.

DrewML: Why does the client need to know what the ID consists of? It sounds like something that needs to be known by whoever is implementing the schema on the server

DrewML: > I would prefer to keep it because this is important for the use case: read from here and use when adding items to wishlist or gift registry.

I think that can be solved with a more descriptive field name. What do you think?

> It was also supposed to serve as a hint for implementors.

I think that would make more sense as an inline comment - the description is encoded into the schema and rendered as docs in tools like GraphiQL

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -11,22 +11,134 @@ type CartQueryOutput {
 }
 
 type Cart {
+    id: ID! @doc(description: "The ID of the cart.")
+    items: [CartItemInterface]
+    applied_coupons: [AppliedCoupon] @doc(description:"An array of `AppliedCoupon` objects. Each object contains the `code` text attribute, which specifies the coupon code")
```
akaplya, line 16: right now Magento supports only one coupon per cart, it is better to explicitly describe it in a comment.

design-documents/graph-ql/coverage/Wishlist.graphqls
```diff
@@ -17,11 +13,11 @@ type Customer {
 
 type Wishlist {
     id: ID
-    items: [WishlistItem]
+    items: [WishlistItemInterface]
     items_count: Int
     sharing_code: String
     updated_at: String
-    #name: String should be added in Commerce edition
+    name: String @doc(description: "Avaialble in Commerce edition only")
```
akaplya, line 20: Can we add this field thought the dedicated module?

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: The client does not, but implementors do. It is much easier to describe it here for whoever will be working on this schema. We can skip this description in the schema committed to the Magento repo.

design-documents/graph-ql/coverage/Cart.graphqls
DrewML: > right now Magento supports only one coupon per cart, it is better to explicitly describe it in a comment.


If we don't support this functionality, what's the value in having the functionality reflected in the schema at all>

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: Okay, will convert to inline comments. This is supposed to be an ID, what name do you suggest? I think this has been discussed previously and we have not found better name so far.

DrewML: Sure, I'm just curious why we wouldn't use an inline comment? That's what comments are part of SDL for :)

DrewML: Would also still like to see a name that better reflects what `id_v2` means for a client - you've specified intent for the field, but that doesn't show through in the schema design yet

DrewML: (nevermind, we're discussing this in another thread - too many schema reviews!)

design-documents/graph-ql/coverage/Wishlist.graphqls
paliarush: This is just a design, we don't have modules here. What do you mean?

In general, yes, fields can be added from separate modules.

design-documents/graph-ql/coverage/Cart.graphqls
```diff
@@ -11,22 +11,138 @@ type CartQueryOutput {
...
+    type: PriceTypeEnum!
+}
+
+type SelectedGiftCardAmount {
+    # Hash from the type of the option and value
+    id: ID!
+    value: Money!
+}
+
+enum PriceTypeEnum @doc(description: "This enumeration the price type.") {
```
DrewML, line 142: Description on this one isn't super clear, but probably doesn't need one (just repeats what the name of the type already says)

design-documents/graph-ql/coverage/Cart.graphqls
DrewML: Want to make sure I really understand this type's purpose before making a suggestion, because I still don't think I'm fully grasping it.

```graphql
type SelectedConfigurableOption {
    # Hash which includes option ID, option type and slected values. Can be used to move configurable product to wishlist or gift registry
    id_v2: ID!
    id: Int!
    option_label: String!
    value_id: Int!
    value_label: String!
}
```

Is it possible we're combining 2 logically separate types into one here? I'm having a hard time understanding how an entity could conceptually have 2 separate identifiers.

Is this type holding both the information about the option (from DB), and also the state from the user's cart? Maybe a screen-shot of what UI this is modeling would help

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: @DrewML This is already released schema and done this way to allow for future or 3rd party extensions. When support for multiple coupons is added clients will not have to be changed.

design-documents/graph-ql/coverage/Wishlist.graphqls
```diff
@@ -47,22 +43,44 @@ input WishlistItemInput {
...
+    samples: [DownloadableProductSamples] @doc(description: "DownloadableProductSamples defines characteristics of a downloadable product")
+}
+
+type BundleWishlistItem implements WishlistItemInterface {
+    bundle_options: [SelectedBundleOption!]
+}
+
+type GiftCardCartItem implements CartItemInterface {
+    sender_name: String!
+    recepient_name: String!
```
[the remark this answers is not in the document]
paliarush: This is also released by now, need to fix...

design-documents/graph-ql/coverage/Cart.graphqls
paliarush: Again, taken from existing schema as is.. Will correct here

design-documents/graph-ql/coverage/Wishlist.graphqls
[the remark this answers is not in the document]
paliarush: There were couple more places with the same typo: https://github.com/magento/architecture/pull/383/commits/bb67393fb1f161dc1fb2afa4c49ed0e2a8b85545

design-documents/graph-ql/coverage/Cart.graphqls
DrewML: Cool, if it's already out there let's just go with it 👍

PR #384, opened 2020-06-12
ECP-756: GraphQL Schema for Rewards

paliarush:
https://jira.corp.magento.com/browse/ECP-756

### DIFF
--- a/design-documents/graph-ql/coverage/reward-points.graphqls
+++ b/design-documents/graph-ql/coverage/reward-points.graphqls
@@ -2,16 +2,33 @@ type Customer {
     reward_points: RewardPoints @doc(description: "Customer reward points")
 }
 
+type Cart {
+    applied_reward_points: RewardPointsAmount
+}
+
+type Mutation {
+    applyRewardPointsToCart(cartId: ID!): ApplyRewardPointsToCartOutput
+    removeRewardPointsFromCart(cartId: ID!): RemoveRewardPointsFromCartOutput
+}
+
+type ApplyRewardPointsToCartOutput {
+    cart: Cart
+}
+
+type RemoveRewardPointsFromCartOutput {
+    cart: Cart
+}
+
 type RewardPoints {
-    balance: RewardPointsBalance @doc(description: "Current reward points balance")
-    exchange_rate: RewardPointsExchangeRate @doc(description: "Current reward points exchange rate")
+    balance: RewardPointsAmount @doc(description: "Current reward points balance")
+    exchange_rates: RewardPointsExchangeRates @doc(description: "Current reward points exchange rates")
     subscription_status: RewardPointsSubscriptionStatus @doc(description: "Reward points related email subscription status")
     balance_history: [RewardPointsBalanceHistoryItem] @doc(description: "Reward points balance history. Visibility of this section depends on the configuration")
 }
 
-type RewardPointsBalance {
-    points: Float! @doc(description: "Reward points balance in points")
-    amount: Money! @doc(description: "Reward points balance in store currency")
+type RewardPointsAmount {
+    points: Float! @doc(description: "Reward points amount in points")
+    money: Money! @doc(description: "Reward points amount in store currency")
 }
 
 type RewardPointsExchangeRates @doc (description: "Exchange rates depend on the customer group"){
@@ -30,7 +47,7 @@ enum SubscriptionStatus {
 }
 
 type RewardPointsBalanceHistoryItem {
-    balance: RewardPointsBalance @doc(description: "Reward points balance after the completion of the transaction")
+    balance: RewardPointsAmount @doc(description: "Reward points balance after the completion of the transaction")
     points_change: Float! @doc(description: "Number of points added or deducted in the transaction")
     change_reason: String! @doc(description: "Reason for balance change")
     date: String! @doc(description: "Transaction date")
--- a/design-documents/graph-ql/coverage/reward-points.graphqls
+++ b/design-documents/graph-ql/coverage/reward-points.graphqls
@@ -12,11 +12,11 @@ type Mutation {
 }
 
 type ApplyRewardPointsToCartOutput {
-    cart: Cart
+    cart: Cart!
 }
 
 type RemoveRewardPointsFromCartOutput {
-    cart: Cart
+    cart: Cart!
 }
 
 type RewardPoints {
--- a/design-documents/graph-ql/coverage/reward-points.graphqls
+++ b/design-documents/graph-ql/coverage/reward-points.graphqls
@@ -23,7 +23,7 @@ type RewardPoints {
     balance: RewardPointsAmount @doc(description: "Current reward points balance")
     exchange_rates: RewardPointsExchangeRates @doc(description: "Current reward points exchange rates")
     subscription_status: RewardPointsSubscriptionStatus @doc(description: "Reward points related email subscription status")
-    balance_history: [RewardPointsBalanceHistoryItem] @doc(description: "Reward points balance history. This field will be set to null if disabled in the admin")
+    balance_history: [RewardPointsBalanceHistoryItem!] @doc(description: "Reward points balance history. This field will be set to null if disabled in the admin")
 }
 
 type RewardPointsAmount {
--- a/design-documents/graph-ql/coverage/reward-points.graphqls
+++ b/design-documents/graph-ql/coverage/reward-points.graphqls
@@ -1,0 +1,38 @@
+type Customer {
+    reward_points: RewardPoints @doc(description: "Customer reward points")
+}
+
+type RewardPoints {
+    balance: RewardPointsBalance @doc(description: "Current reward points balance")
+    exchange_rate: RewardPointsExchangeRate @doc(description: "Current reward points exchange rate")
+    subscription_status: RewardPointsSubscriptionStatus @doc(description: "Reward points related email subscription status")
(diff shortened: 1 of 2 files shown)


design-documents/graph-ql/coverage/reward-points.graphqls
```diff
@@ -0,0 +1,38 @@
+type Customer {
+    reward_points: RewardPoints @doc(description: "Customer reward points")
+}
+
+type RewardPoints {
+    balance: RewardPointsBalance @doc(description: "Current reward points balance")
+    exchange_rate: RewardPointsExchangeRate @doc(description: "Current reward points exchange rate")
+    subscription_status: RewardPointsSubscriptionStatus @doc(description: "Reward points related email subscription status")
+    balance_history: [RewardPointsBalanceHistoryItem] @doc(description: "Reward points balance history. Visibility of this section depends on the configuration")
```
DrewML, line 9: > Visibility of this section depends on the configuration

This comment doesn't make it clear whether the type definition will not exist in the schema based on config, or if it'll just be `null` depending on config

paliarush: Thank you, please check adjusted description.

DrewML: Thanks! New description is much more clear

design-documents/graph-ql/coverage/reward-points.graphqls
```diff
@@ -0,0 +1,70 @@
...
+    reward_points: RewardPoints @doc(description: "Customer reward points details")
+}
+
+type Cart {
+    applied_reward_points: RewardPointsAmount @doc(description: "Reward points ammount applied to the cart")
+}
+
+type Mutation {
+    applyRewardPointsToCart(cartId: ID!): ApplyRewardPointsToCartOutput @doc(description: "Apply all available points up to the cart total. Partial redemption is not available")
+    removeRewardPointsFromCart(cartId: ID!): RemoveRewardPointsFromCartOutput @doc(description: "Cancel usage of the previously applied reward points to cart")
```
DrewML, line 11: Probably not the end of the world, but the symmetry here seems off with the naming

- `apply`
- `remove`

I think `apply` makes sense, since you're not conceptually _adding_ to the cart.

But if you're not `adding` something, you wouldn't expect to be able to `remove` that thing.

tl;dr: I think `apply` and `remove` don't work well together, but I'm not sure of a better alternative

design-documents/graph-ql/coverage/reward-points.graphqls
```diff
@@ -0,0 +1,70 @@
...
+    money: Money! @doc(description: "Reward points amount in store currency")
+}
+
+type RewardPointsExchangeRates @doc (description: "Exchange rates depend on the customer group"){
+    earning: Float! @doc(description: "How many points are earned per 1 currency spent")
+    redemption: Float! @doc(description: "How many points need to be redeemed to get 1 currency discount at the checkout")
+}
+
+type RewardPointsSubscriptionStatus {
+    balance_updates: SubscriptionStatus! @doc(description: "Customer subscription status to 'Reward points balance updates' emails")
```
DrewML, line 40: `balance_updates` seems a bit generic here, while the return type and the description give a lot more context.

Can we name this field in a way that it's less likely folks need to reference the docs? Maybe `balance_update_email_subscription`? Going to be a trade-off with verbosity, maybe we can find a happy middle ground.

design-documents/graph-ql/coverage/reward-points.graphqls
paliarush: This the the naming we currently have for coupons and gift cards.. not sure whether there is a good alternative. Also seems better to keep similar naming for consistency.

design-documents/graph-ql/coverage/reward-points.graphqls
paliarush: The type name already contains "SubscriptionStatus", so if you go from there it is "balance updates subscription status". Let me know if you still believe it is better to change it.

design-documents/graph-ql/coverage/reward-points.graphqls
DrewML: > Also seems better to keep similar naming for consistency.

I think this could be argued in both directions - depends on whether you're optimizing for folks new to the API, or folks familiar with the existing API surface.

But I also can't think of better wording so it's a moot point 😂. Makes sense to me to just move forward by matching what we have 👍

design-documents/graph-ql/coverage/reward-points.graphqls
```diff
@@ -0,0 +1,70 @@
...
+    applied_reward_points: RewardPointsAmount @doc(description: "Reward points ammount applied to the cart")
+}
+
+type Mutation {
+    applyRewardPointsToCart(cartId: ID!): ApplyRewardPointsToCartOutput @doc(description: "Apply all available points up to the cart total. Partial redemption is not available")
+    removeRewardPointsFromCart(cartId: ID!): RemoveRewardPointsFromCartOutput @doc(description: "Cancel usage of the previously applied reward points to cart")
+}
+
+type ApplyRewardPointsToCartOutput {
+    cart: Cart
```
DrewML, line 15: We've discussed that, for object types, we want non-scalar fields to be nullable, which we've followed here.

But, for mutation return types, I'm wondering if we expect a case where we could do a _write_ to the cart service, but then not be able to do a read?

The reason for nullable on non-scalar types is to guard ourselves from cascading failures when some services are down. Do we have concerns that cross service boundaries here (honest question, I truly do not know)?

One trade-off we make with nullability is that the client-side has to do conditional checks when it gets a response, and handle additional states.

```js
function MyComponent() {
   // ... run mutation here
   if (response.error) {
      // handle error updating cart
   }
   if (response.cart) {
      // safe to use cart data when rendering
   } else {
      // handle case where we were able to successfully update the cart,
      // but weren't able to query for new results
   }
}
```

when we go non-nullable, that can turn into:

```js
function MyComponent() {
   if (response.error) {
      // handle error updating cart
   }
   
   // render UI with cart data
}
```

I think non-nullable makes a bit more sense in the context of a mutation return type since we have more context about what the client would expect. Thoughts?

design-documents/graph-ql/coverage/reward-points.graphqls
DrewML: That's a good point - I think it's good as-is 👍

design-documents/graph-ql/coverage/reward-points.graphqls
```diff
@@ -0,0 +1,70 @@
...
+
+type RewardPointsBalanceHistoryItem {
+    balance: RewardPointsAmount @doc(description: "Reward points balance after the completion of the transaction")
+    points_change: Float! @doc(description: "Number of points added or deducted in the transaction")
+    change_reason: String! @doc(description: "Reason for balance change")
+    date: String! @doc(description: "Transaction date")
+}
+
+type StoreConfig {
+    magento_reward_general_is_enabled: String @doc(description: "Reward points functionality status: enabled/disabled")
```
DrewML, line 57: Is there something in core preventing us from using a Boolean here?

design-documents/graph-ql/coverage/reward-points.graphqls
paliarush: Sounds reasonable, fixing it here. @DrewML , can I ask you to write these rules down somewhere in https://github.com/magento/architecture/tree/master/design-documents/graph-ql ?

1. Non-nullable for mutations
1. Nullable for objects in queries
1. Non-nullable for scalars in queries

design-documents/graph-ql/coverage/reward-points.graphqls
paliarush: The logic here is, we take data from `core_config_data` as is and run it through the list of fields declared in this GraphQL schema to leave only allowed ones. 
These values are currently stored in DB as 0/1. I don't think GraphQL framework will properly convert them to boolean. It may be possible, but will require additional effort to add support.

DrewML: I can definitely see how this makes the implementation easier/lower effort.

My concern is that the work doesn't go away, we're just pushing it to every client instead. Every client now has to do:

```js
if (magento_reward_general_is_enabled === 'enabled') {

}
```

rather than us returning a more descriptive type (`Boolean`)

fnhipster: My 2 cents: there is already a lot of unnecessary front-end logic in the client. e.g. mapping and transforming data coming from GraphQL. Although this might seem small, it does make a more descriptive and elegant API. We gotta start trimming logic where we can in order to enable clients to be leaner.

paliarush: I would address it in scope of separate story related to `StoreConfig`. At the time it was implemented we agreed to the tradeoff. 
Otherwise Reward implementation will be blocked by the rework of `StoreConfig`. 

@nrkapoor are you okay with creating a separate story for improving DX of `StoreConfig` query?

nrkapoor: @paliarush Yes I will create a separate ticket for storeconfig clean.

DrewML: Sounds good - thanks y'all 👍

design-documents/graph-ql/coverage/reward-points.graphqls
DrewML: Thanks for updating!

Left myself a TODO to add a document covering our experiences/preferences with nullability

DrewML:
Great work @paliarush!

PR #385, opened 2020-06-16
Proposal: Deprecate `Mutation.createEmptyCart`, add `Mutation.createGuestCart`

DrewML:
[Rendered Proposal](/magento/architecture/blob/deprecate-createEmptyCart-add-createCart/design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md)

### DIFF
--- a/design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
+++ b/design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
@@ -1,0 +1,49 @@
+# Deprecate and replace `Mutation.createEmptyCart`
+
+## What
+
+- Mark `Mutation.createEmptyCart` as deprecated
+- Add `Mutation.createGuestCart` as a replacement that includes `Cart` in its return type
+
+## Why
+
+This PR started from a Slack discussion about mutations and client-side caching. At one point [@sirugh](https://github.com/sirugh) mentioned:
+
+> This makes me wonder if the createCart mutation should return a cart type instead of just the id
+
+There are 2 problems that I'm aware of with the current schema returning just an `ID` instead of the `Cart` type:
+
+1. Requires manual cache handling with libraries like Apollo, to link the `ID` back to a `Cart` object
+2. Requires 2 round trips for the client to get cart data for a guest. Even though the cart is empty, it still has defaults the pwa-studio components will use to render an empty cart
+
+During discussions, it was also noted that `createEmptyCart` has some behavior that's not clearly described by our schema. That is, If a customer is logged-in and has items in the cart, `createEmptyCart` returns an ID referencing a cart that is _not_ empty, which is extremely unintuitive. It was mentioned that the `pwa-studio` code [aliases this query](https://github.com/magento/pwa-studio/blob/38d652a4fbc797a4ac8ac0c3efa611003152c090/packages/venia-ui/lib/queries/createCart.graphql#L4) because of that.
+
+## Intended Usage
+Because the UI knows if it has a token for a customer, `Query.cart` should be used by the UI to get the cart for a logged-in user. If there is _not_ a customer token present, the UI should use `createGuestCart`, which will return a `Cart` object, the same return type as `Query.cart`.
+
+
+## Proposed Changes
+
+```diff
+type Mutation {
++    createGuestCart(
++        input: CreateGuestCartInput
++    ): CreateCartOutput @doc(description: "Create a new shopping cart")
+    createEmptyCart(
+        input: createEmptyCartInput
++    ): String @deprecated(reason: "Use `Mutation.createGuestCart`, or `Query.cart` for logged-in shoppers")
+-    ): String @doc(description:"Creates an empty shopping cart for a guest or logged in user")
+}
+
++input CreateGuestCartInput {
++    cart_id: String @doc(description: "Optional client-generated ID")
++}
++
++type CreateGuestCartOutput {
++    cart: Cart
++}
+
+input createEmptyCartInput {
+    cart_id: String
+}
+```
--- a/design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
+++ b/design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
@@ -28,7 +28,7 @@ Because the UI knows if it has a token for a customer, `Query.cart` should be us
 type Mutation {
 +    createGuestCart(
 +        input: CreateGuestCartInput
-+    ): CreateCartOutput @doc(description: "Create a new shopping cart")
++    ): CreateGuestCartOutput @doc(description: "Create a new shopping cart")
     createEmptyCart(
         input: createEmptyCartInput
 +    ): String @deprecated(reason: "Use `Mutation.createGuestCart`, or `Query.cart` for logged-in shoppers")
@@ -36,7 +36,7 @@ type Mutation {
 }
 
 +input CreateGuestCartInput {
-+    cart_id: String @doc(description: "Optional client-generated ID")
++    cart_uid: ID @doc(description: "Optional client-generated ID")
 +}
 +
 +type CreateGuestCartOutput {


design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createCart.md
```diff
@@ -0,0 +1,40 @@
...
+
+It's not ideal that a common flow like this requires 2 round-trips, one after the other.
+
+## Proposed Changes
+
+```diff
+type Mutation {
++    createCart(
++        input: CreateCartInput
++    ): CreateCartOutput @doc(description: "Create a new shopping cart")
```
sirugh, line 22: The current `createEmptyCart` mutation has some magic to it. If the request is authenticated with a bearer token the mutation returns the user's existing cart id otherwise it returns a new cart's id. 

I see in the new response type the `cart` is returned, which is ideal, but would it also still return a user's existing cart?

DrewML: > If the request is authenticated with a bearer token the mutation returns the user's existing cart id otherwise it returns a new cart's id.

Does it return the existing cart ID even in the case where that cart _is not_ empty? If so that makes the `createEmptyCart` name a bit counter-intuitive.

DrewML: > Does it return the existing cart ID even in the case where that cart is not empty

Checked, it does. 

- [`createCartForCustomer->execute`](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/QuoteGraphQl/Model/Resolver/CreateEmptyCart.php#L70) in resolver
- [`createCustomerCart`](https://github.com/magento/magento2/blob/c28466437c5b6867bd1d517c184e51d690ebfc46/app/code/Magento/Quote/Model/QuoteManagement.php#L269) in `createEmptyCartForCustomer`
- [`quoteRepository->getActiveForCustomer`](https://github.com/magento/magento2/blob/c28466437c5b6867bd1d517c184e51d690ebfc46/app/code/Magento/Quote/Model/QuoteManagement.php#L351) in `createCustomerCart`

That's confusing! I think the name is more confusing than the API, because you wouldn't want to flush a customer's existing cart, but a customer can only have a single cart.

I wonder what a better name is here? A name that makes it clear that this returns an existing thing, or falls back to creating a new thing.

sirugh: Yes, it returns the user's existing cart id in both cases. And yes, I agree. That's why we alias over it in [our code](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/createCart.graphql).

sirugh: > I wonder what a better name is here? A name that makes it clear that this returns an existing thing, or falls back to creating a new thing.

In PWA Studio a user can only have one cart visible and active. We have no concept of fetching data for a "stashed" cart. So I can't see why you ever need the concept of `new` cart. I would think that simply querying for `cart` would suffice regardless. You're either getting a new cart or an active one if authed.

DrewML: > So I can't see why you ever need the concept of new cart

Y'all need it for guest users, right? You don't need the concept of create for logged-in because there's a 1:1 relationship between customer/cart, so the server can hide creation from the client. But for a guest you'd need to ask to create _something_ to track cart items.

---

A bit confused about the "stashed cart" here. I'm talking about these 2 use-cases specifically that the `createEmptyCart` API handles today:

1. Getting existing cart (or quietly creating new one) if logged-in
2. Creating a guest cart

You mentioned the behavior works well, but we agree the name `createEmptyCart` is off. Just trying to identify a name that better describes that functionality

sirugh: Sorry - shouldn't have even mentioned "stashed cart". I don't even know if Magento has the idea of a customer having multiple carts at once.

`getCart` seems reasonable, but the name sounds more like a query.

DrewML: Went and looked at venia a bit more, and now I'm questioning if `createEmptyCart` API handling logged-in users at all.

@sirugh what's the situation where we would want to use `createEmptyCart` for a logged-in user? It seems like, at the time venia sends this mutation, we already should know if the user is a guest or not, right?

I think these are the cases we need to cover, but _please_ correct me if I'm wrong:

Use Case | Covered by
-- | --
Fetch cart for logged-in user | Query.customerCart
Fetch cart for guest user | Query.cart
Get a cart id to use for guest user | Mutation.createEmptyCart

If the above is correct, is seems like `createEmptyCart` should be something like `createGuestCart`, and you can skip that call entirely when a user is logged-in. Thoughts?

DrewML: > getCart seems reasonable, but the name sounds more like a query.

Yeah, I keep wanting to say it, but then stop myself each time because it doesn't make it clear that it's a side-effecty operation.

Hopefully my proposal above this comment improves things?

sirugh: > @sirugh what's the situation where we would want to use createEmptyCart for a logged-in user? It seems like, at the time venia sends this mutation, we already should know if the user is a guest or not, right?

I don't remember why we used the single mutation but it seemed to make the logic simpler.

> If the above is correct, is seems like createEmptyCart should be something like createGuestCart. Thoughts?

Sure that sounds fine.

DrewML: > I don't remember why we used the single mutation but it seemed to make the logic simpler.

Makes sense. I guess we ended up just moving the logic into an odd place: a `createEmptyCart` query that can return a non-empty cart.

I pushed up some changes to the proposal based on this thread. Can you give it another read through?

DrewML:
@nrkapoor Can confirm - not a breaking change. Only change to an existing API is marking for deprecation, but not strict requirement to remove until down the line when we're sure usage has mostly dropped off.

nrkapoor:
> @nrkapoor Can confirm - not a breaking change. Only change to an existing API is marking for deprecation, but not strict requirement to remove until down the line when we're sure usage has mostly dropped off.

@DrewML Thanks. Please let me know when we are ready to merge this change so I can create a ticket on the graphql board for implementation.

design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
```diff
@@ -0,0 +1,49 @@
...
+This PR started from a Slack discussion about mutations and client-side caching. At one point [@sirugh](https://github.com/sirugh) mentioned:
+
+> This makes me wonder if the createCart mutation should return a cart type instead of just the id
+
+There are 2 problems that I'm aware of with the current schema returning just an `ID` instead of the `Cart` type:
+
+1. Requires manual cache handling with libraries like Apollo, to link the `ID` back to a `Cart` object
+2. Requires 2 round trips for the client to get cart data for a guest. Even though the cart is empty, it still has defaults the pwa-studio components will use to render an empty cart
+
+During discussions, it was also noted that `createEmptyCart` has some behavior that's not clearly described by our schema. That is, If a customer is logged-in and has items in the cart, `createEmptyCart` returns an ID referencing a cart that is _not_ empty, which is extremely unintuitive. It was mentioned that the `pwa-studio` code [aliases this query](https://github.com/magento/pwa-studio/blob/38d652a4fbc797a4ac8ac0c3efa611003152c090/packages/venia-ui/lib/queries/createCart.graphql#L4) because of that.
```
paliarush, line 19: This sounds like a bug. Registered customers are supposed to use `customerCart` query to get cart object (either new or existing)

design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
```diff
@@ -0,0 +1,49 @@
...
++        input: CreateGuestCartInput
++    ): CreateCartOutput @doc(description: "Create a new shopping cart")
+    createEmptyCart(
+        input: createEmptyCartInput
++    ): String @deprecated(reason: "Use `Mutation.createGuestCart`, or `Query.cart` for logged-in shoppers")
+-    ): String @doc(description:"Creates an empty shopping cart for a guest or logged in user")
+}
+
++input CreateGuestCartInput {
++    cart_id: String @doc(description: "Optional client-generated ID")
```
paliarush, line 39: Why not 
```suggestion
+    cart_id: ID @doc(description: "Optional client-generated ID")
```

design-documents/graph-ql/coverage/deprecate-createEmptyCart-add-createGuestCart.md
```diff
@@ -0,0 +1,49 @@
...
+Because the UI knows if it has a token for a customer, `Query.cart` should be used by the UI to get the cart for a logged-in user. If there is _not_ a customer token present, the UI should use `createGuestCart`, which will return a `Cart` object, the same return type as `Query.cart`.
+
+
+## Proposed Changes
+
+```diff
+type Mutation {
++    createGuestCart(
++        input: CreateGuestCartInput
++    ): CreateCartOutput @doc(description: "Create a new shopping cart")
```
melnikovi, line 31:
```suggestion
+    ): CreateGuestCartOutput @doc(description: "Create a new shopping cart")
```

PR #386, opened 2020-06-17
Catalog media tech vision

buskamuza:
Technical vision for future of managing catalog media.


design-documents/media/catalog-images.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Questions:
+
+1. Do we sync full image URL to SF or provide Base CDN URL as configuration for the store?
+   1. Full URLs as part of product data: full reindex will be required in case CDN URL changes.
+   2. Base CDN URL as a store configuration + relative asset path as part of product data: added complexity of the SF App due to additional knowledge about CDN/Media Base URLs (especially in case multiple should be supported).
+
+## Dependencies
+
+1. To allow only original image URLs stored at the Store-Front side, new format of transformed images should be supported with transformation parameters are passed as parameters to the original URL. Current transformed URL: `https://magento.store.com/media/catalog/product/cache/1/2/3/<hash>.jpg` (where `<hash>` is generated based on transformation parameters and those become invisible). Desired transformed URL: `https://store.cdn.com/media/catalog/product/1/2/3/product-image.jpg?w=100&h=100` (transformation parameters are clearly visible).
```
kandy, line 70: Can we add hash as parameter to desired URL format for backward compatibility purposes, like `?w=100&h=100&hash=<hash>`?
And what about other parameters like watermark?

buskamuza: 1. Hash. It won't give good performance I assume because currently hash is part of the file path and it's retrieved bypassing any PHP part of the application. Tough, possibly a web server can be configured to replicate same behavior. So yes, it's possible to support such format.
2. Other parameters. Would be similar. The parameters would be part of the client application based on which CDN is used in the application. I don't think Magento's Catalog responsibility (or any backend) would be to provide URL format. Drawback would be that client should know format of the transformation URL based on underlying technologies. So far, at least some CDNs can support same format: Fastly (has its own format), CloudFront relies on custom Lambda anyways (so it can support same Fastly format), Nginx/Apache  is configurable (so can also support same format as Fastly).
   1. Watermark. Yes, this part is the most questionable (mostly from UX side). Image transformation service should support image image overlay.

Will add more details about different types of image transformation.

DrewML:
Can we include an overview of the current state, and what problems we're trying to solve with these changes?

buskamuza:
> Can we include an overview of the current state, and what problems we're trying to solve with these changes?

It is for the new Storefront service, which doesn't exist yet (not released at least). So the purpose of this document is to describe how new Storefront should handle assets.

Though, there is something that currently exists: existing Magento Catalog GraphQL returns image URLs, which are transformed images, not originals. Image transformation is covered in the document. And there is "Breaking Changes" section in the end, which explains that the new vision makes GraphQL return original URLs instead (problems with this are also described). Other than this, I don't think anything really exists.

PR #387, opened 2020-06-17
Make destination param nullable in mergeCarts

DrewML:
[Rendered Proposal](/magento/architecture/blob/mergeCarts-optional-destination/design-documents/graph-ql/coverage/mergeCarts-optional-destination.md)

There are no breaking changes and no deprecations.

### DIFF
--- a/design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
+++ b/design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
@@ -1,0 +1,40 @@
+# `mergeCarts` - Make `destination_cart_id` optional
+
+## What
+
+- Make the `destination_cart_id` argument optional in the `mergeCarts` mutation
+
+## Why
+
+This came up in a discussion with [@sirugh](https://github.com/sirugh) from the [`pwa-studio`](https://github.com/magento/pwa-studio) team.
+
+When a user logs in (creates a new token), one of the first things the UI needs to do is merge the current guest cart (if items are present) into the customer account's cart.
+
+If `destination_cart_id` is required, this requires 3 round trips:
+
+1. Call for `Mutation.generateCustomerToken`
+2. Call for `Query.cart` to get customer cart ID
+3. Call for `Mutation.mergeCarts` to merge guest cart ID into customer cart
+
+Because a customer can only have a single cart, and this API only works for authenticated users, `destination_cart_id` is an unnecessary requirement here. If we make it optional, the login + cart upgrade for UI can happen in a single request:
+
+```graphql
+# Mutations run serially, in-order. So `mergeCarts` will only execute
+# if `generateCustomerToken` succeeds
+mutation LoginAndMergeCarts($email: String!, $password: String!, $guestCartID: String!) {
+    generateCustomerToken(email: $email, password: $password) {
+        token
+    }
+    mergeCarts(source_cart_id: $guestCartID) {
+        ID
+    }
+}
+```
+
+## Proposed Change
+```diff
+type Mutation {
+-    mergeCarts(source_cart_id: String!, destination_cart_id: String!): Cart!
++    mergeCarts(source_cart_id: String!, destination_cart_id: String): Cart!
+}
+```


design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
```diff
@@ -0,0 +1,40 @@
...
+2. Call for `Query.cart` to get customer cart ID
+3. Call for `Mutation.mergeCarts` to merge guest cart ID into customer cart
+
+Because a customer can only have a single cart, and this API only works for authenticated users, `destination_cart_id` is an unnecessary requirement here. If we make it optional, the login + cart upgrade for UI can happen in a single request:
+
+```graphql
+# Mutations run serially, in-order. So `mergeCarts` will only execute
+# if `generateCustomerToken` succeeds
+mutation LoginAndMergeCarts($email: String!, $password: String!, $guestCartID: String!) {
+    generateCustomerToken(email: $email, password: $password) {
```
DrewML, line 25: @rugh pointed out there's some design weirdness here with how we handle authentication.

Technically, a request is considered "authenticated" if a customer token has been passed. In this example, `mergeCarts` has been called in a request that does _not_ have an `Authorization` header.

The weirdness is that we do know at this point if the user has a token (because `mergeCarts` can't run if `generateCustomerToken` fails. The question is whether we're comfortable propagating that information from `generateCustomerToken` to the rest of the defined operations (in this example, `mergeCarts`).

paliarush: I'm not sure we can make it work in webonyx implementation without dangerous security hacks.

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
```diff
@@ -0,0 +1,40 @@
+# `mergeCarts` - Make `destination_cart_id` optional
```
paliarush, line 1: Please modify existing document https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/shared-cart.md instead of creating a new one.

BTW, this option was listed in the original proposal as an alternative, but was rejected because of less clear semantics.

DrewML: > BTW, this option was listed in the original proposal as an alternative, but was rejected because of less clear semantics.

I looked through the document and didn't see any mention of why this strategy was rejected. Can you help me understand what about the API was unclear? `createEmptyCart` returning a populated cart is pretty unclear as well, so I'd like to solve both problems at once if possible.

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
DrewML: > I'm not sure we can make it work in webonyx implementation without dangerous security hacks.

Can you expand on this more? I'd like to understand what's the limitation(s) will be.

paliarush: Current authentication mechanisms initialize user context (containing user type and user ID) before any resolvers are executed. If I understand correctly your suggestion is that the first resolver will modify the user context, so that when the second resolver is executed, it will work as if the user is already authenticated. This looks like a hack.

DrewML: >  If I understand correctly your suggestion is that the first resolver will modify the user context

Sort of. I was just suggesting an extremely simplified view - I'm not making the literal suggestion of a resolver reaching out and doing something like `context.userState = newUserState` 😄.

> If I understand correctly your suggestion is that the first resolver will modify the user context, so that when the second resolver is executed, it will work as if the user is already authenticated. This looks like a hack.

`context` is a bit of an implementation detail I suppose - let's generalize a bit and say I'm suggesting that underlying state be modified as a side-effect of running `generateCustomerToken`, and the next mutations executed _in that same operation_ have visibility of that side-effect's state change. I think this meshes well with the spirit of mutations, based on their description:

> It is expected that the top level fields in a mutation operation perform side‐effects on the underlying data system. Serial execution of the provided mutations ensures against race conditions during these side‐effects.
http://spec.graphql.org/draft/#sec-Mutation

> Current authentication mechanisms initialize user context (containing user type and user ID) before any resolvers are executed

Something to note is that we chose this as an implementation, and we can expand/improve upon it! I think in general it's worthwhile for us to design a schema that works for client needs, and then improve our implementation regularly to deliver on those needs. Right now it sounds like we're limiting the design of our schemas to only match how our framework works today.

---

I think a lot of this is going to boil down to technical purity vs real world needs. One of our big selling points for GraphQL was to minimize the number of requests needed to perform typical storefront operations - this is a case where we haven't delivered on the ability to satisfy a common operation (logging in and transferring a cart) in a single request.

DrewML: @paliarush As an alternative, what would your thoughts be on us just adding something like this:

```diff
type Mutation {
   generateCustomerToken(email: String!, password: String!): CustomerToken
+  generateTokenAndMergeCart(input: GenerateTokenAndMergeCartInput): GenerateTokenAndMergeCartOutput
}

+input GenerateTokenAndMergeCartInput {
+  email: String!
+  password: String!
+  guestCartID: ID
+}

+type GenerateTokenAndMergeCartOutput {
+  token: String!
+  cart: Cart!
+  customer: Customer!
+}
```

Maybe I'm just trying to solve too many use-cases with a single mutation, and we can just provide some more high-level APIs like this. Thoughts?

sirugh: A single auth call that returns ids for all the new objects would be swell. We make four separate requests right now: get auth token, get customer cart id, merge carts, get user details. Would be amazing if we could do this in one.

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
DrewML: > I would still prefer to work in scope of one document because otherwise it will be confusing. This repository should ideally contain the desired state, not two different states in different documents.

No problem - I'll move it over.

> Initially we had a requirement to merge guest carts as well. I propose to discuss it again with PO and in case the only supported case is guest cart merge into customer cart

I can do that - happen to know who the PM is?

paliarush: @nrkapoor is usually making decisions on GraphQL schema and involves someone else if needed

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
paliarush: @DrewML authentication logic is reused across all web APIs, admin and storefront. I would hate to make differentiation for GraphQL specifically, it will increase complexity and lead to potential security bugs.

If there is a need, I would say we should optimize for the specific use case. I like creating a new mutation, which does all the things for the use case, more than the idea of modifying auth state from the resolver.

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
DrewML: Oh cool - wasn't sure if we were talking a PO than manages Ordering/Checkout or just GraphQl in general. I'll ping him

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
DrewML: That's totally fair.

Now that I'm looking at the new mutation, I prefer it as well. TBH I don't think we really need to touch `mergeCarts` in that case, we can leave it for folks that want that separate operation on its own. Sound like a reasonable path forward?

design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
DrewML: If we end up introducing a new mutation (discussed in the other comment thread), it won't be critical for us to make changes to `mergeCarts`

cpartica: this was merged part of MC-37413  as described here. i will update https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/shared-cart.md as suggested by Alex

cpartica (approved):

PR #391, opened 2020-06-25
req list items pagiantion

dani97:
## Problem

Requisition List Items size can be huge and can be a performance issue. Current schema doesn't  handle it.
Update Requisition List items cannot handle edit config variation, adding the support from Alex request
## Solution
Adding pagination to Requisition List Items
Update to Req list items Input for supporting custom options, configurable options.

## Requested Reviewers
@nrkapoor 
@DrewML 
@paliarush 
@cpartica


nrkapoor:
@DrewML @paliarush Can you look at these changes and provide feedback?

PR #381, opened 2020-05-27
B2B NegotiableQuote GraphQL Schema

G-Arvind:
## Problem

Coverage for B2B Negotiable quotes

## Solution

Schema for B2B Requisition List

Queries

- Get single NegotiableQuote
- Get All NegotiableQuotes
- Estimate shipping cost for NegotiableQuote
- Estimate Shipping Cost by AddressId for NegotiableQuote
- estimate Shipping information for NegotiableQuote

Mutations

- Update Price
- Request a NegotiableQuote
- Set shipping method to NegotiableQuote

PR #392, opened 2020-06-25
Added design document with nullability recommendations for GraphQL schemas

DrewML:
[Rendered Document](/magento/architecture/blob/nullability/design-documents/graph-ql/nullability.md)

Prior discussions about nullability:

- https://github.com/magento/architecture/pull/372
- https://github.com/magento/architecture/pull/381#discussion_r431408317
- https://github.com/magento/architecture/pull/368#discussion_r414907353

### DIFF
--- a/design-documents/graph-ql/nullability.md
+++ b/design-documents/graph-ql/nullability.md
@@ -30,7 +30,14 @@ Making fields non-nullable alleviates some of these issues for clients.
 
 When a field's resolver has an error, GraphQL requires that the value of the field be set to `null`. But, if a field is non-nullable, a `null` value would break that contract.
 
-Instead, an error in a resolver for a non-nullable field propagates up to the nearest nullable ancestor field. This can end up causing a client to lose more data farther up the response tree, even though that data may have been resolved without errors.
+Instead, an error in a resolver for a non-nullable field propagates up to the nearest parent field. It that field is also non-nullable, the error propagates up to the next parent field. This propagation continues until either:
+
+- A nullable parent field is found
+- We reach the root `Query` object
+
+ A non-nullable type in the wrong place can end up causing a client to lose more data farther up the response tree, even though that data may have been resolved without errors.
+
+_[Further reading on the relationship between non-null types and errors](http://spec.graphql.org/draft/#sec-Errors-and-Non-Nullability)_
 
 ### Example: Bad use of non-nullable field
 
--- a/design-documents/graph-ql/nullability.md
+++ b/design-documents/graph-ql/nullability.md
@@ -4,7 +4,7 @@ By default, all field values in GraphQL are _nullable_
 
 ```graphql
 type Query {
-	product(id: ID): Product # Server can also return "null"
+    product(id: ID): Product # Server can also return "null"
 }
 ```
 
@@ -12,7 +12,7 @@ GraphQL provides a wrapper/modifier that can be used to disallow null values. It
 
 ```graphql
 type Query {
-   product(id: ID): Product! # Server cannot return "null"
+    product(id: ID): Product! # Server cannot return "null"
 }
 ```
 
@@ -50,7 +50,7 @@ type Product {
     price: Money!
     description: String!
     url: String!
-    related_products: [RelatedProduct!]!
+    related_products: RelatedProducts!
 }
 ```
 **Query**:
@@ -65,10 +65,12 @@ query ProductDetailsPage($id: ID) {
        # down/unreachable at the time this query runs, we can't fulfill
        # the contract
        related_products {
-          id
-          name
-          price
-          url
+           items {
+               id
+               name
+               price
+               url
+           }
        }
     }
 }
@@ -97,27 +99,55 @@ With this in mind, when in doubt, the safer route is to make a field nullable, w
 
 ## Recommendations
 
-### Scalars
+### Top-level Query fields should always be nullable
 
-If the field's value is obtained from a backing data source, and that value is required in the backing data source, then the field in the GraphQL schema should be non-nullable
+Because of error propagation, if an error occurs in a non-null field on the `Query` root, the client will lose data from all other top-level queries.
 
-#### Example
 ```graphql
-type Product {
-   id: ID! # Products service would not allow a product without an ID, so id is non-nullable
-   color: String # Backing products DB does not require color, so it's nullable
+type Query {
+    # If a client queries for both fields, and one of them fails,
+    # the client will receive no data
+    product(id: ID): Product!
+    category(id: ID): Category!
 }
 ```
 
-### Object Types
+### Primary keys (id field) should always be non-nullable
 
-Object type fields are frequently used to create joins between types. Object type fields should be nullable when the backing data is likely to live in a different service from the parent object type. This rule will require more thought since it depends where we think domain boundaries will be drawn as we decompose to services.
+It very rarely makes sense to have a resource that _can_ have an ID but might not.
 
-#### Example
+IDs are extremely important for caching in most GraphQL clients, so it's worthwhile to be safe here.
+
 ```graphql
 type Product {
-    price: Price! # non-nullable, because price is highly likely to be part of a products service
-	recommended_products: ProductRecommendations # nullable, because product recommendations would likely be a function that lives outside of the products service
+    id: ID! # Rarely makes sense for this to be nullable
+}
+```
+
+### Consider the Parent Type
+
+If you're not dealing with an `id` field or a top-level `Query` field, the most important question to ask is:
+
+**Is the parent object still usable if this field has an error?**
+
+#### Example: Parent still usable with field error
+
+```graphql
+type Product {
+    # Recommended products are not critical data on a product page, and a UI can represent
+    # a product safely without related products, so we keep the field nullable
+	recommended_products: ProductRecommendations
+}
+```
+
+#### Example: Parent not usable with field error
+
+```graphql
+type Product {
+    # A user would not be able to add a product to the cart from the Product
+    # details page if this field fails, because it may have required options.
+    # We make the field's type non-nullable
+    options: ProductOptions!
 }
 ```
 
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,159 @@
...
+```graphql
+type Product {
+   id: ID! # Products service would not allow a product without an ID, so id is non-nullable
+   color: String # Backing products DB does not require color, so it's nullable
+}
+```
+
+### Object Types
+
+Object type fields are frequently used to create joins between types. Object type fields should be nullable when the backing data is likely to live in a different service from the parent object type. This rule will require more thought since it depends where we think domain boundaries will be drawn as we decompose to services.
```
buskamuza, line 114: From the problem statement, it looks like the rule should be "field may be nullable if it is not critical for the parent object". So, if it's ok to have partial result returned in case child object is not available, then this child object should be nullable. If the parent object will lose integrity in case child object is unavailable, then the child should be non-nullable.
Storage may change with changes in design, but scenarios are usually the same doesn't matter what's the underlying design of the system.

DrewML: This is a great point. I re-wrote this document a few times last night, and didn't notice towards the end of the document when I wrote this line that it would apply to all fields, not just list fields:

> A list _field_ should be non-nullable when its absence would make the parent type fairly useless.

I'll refactor the document a bit to clean things up. Thanks for the useful feedback!

DrewML: @buskamuza I pushed an update to reword things - can you give it another read?

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,159 @@
...
+
+Non-nullable fields help form a contract with clients that makes it safe for them to work with a field's value without having to inspect its shape for possible `null` values.
+
+For clients that are written in a language with a static type system, the compiler will force them to wrap their code in conditionals anytime they're accessing a possibly null field. These checks can become tedious if every single field is nullable.
+
+For clients that are written in more dynamic languages (JavaScript), developers need to be very aware of nullable fields in the schema to prevent trying to accessing data on a `null` value. This leads to lots of defensive coding (or bugs).
+
+Making fields non-nullable alleviates some of these issues for clients.
+
+## Why would I ever _not_ use non-nullable fields?
```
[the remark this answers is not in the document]
DrewML: Updated those titles based on our Slack discussion 👍

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,159 @@
...
+   },
+   "errors": [
+      // details about failure in related_products field populated here
+	]
+}
+```
+
+## Backwards Compatibility
+
+- It is a safe, non-breaking change to move a field from nullable to non-nullable
```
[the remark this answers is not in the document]
DrewML: Added!

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,159 @@
...
+
+For clients that are written in more dynamic languages (JavaScript), developers need to be very aware of nullable fields in the schema to prevent trying to accessing data on a `null` value. This leads to lots of defensive coding (or bugs).
+
+Making fields non-nullable alleviates some of these issues for clients.
+
+## Why would I ever _not_ use non-nullable fields?
+
+When a field's resolver has an error, GraphQL requires that the value of the field be set to `null`. But, if a field is non-nullable, a `null` value would break that contract.
+
+Instead, an error in a resolver for a non-nullable field propagates up to the nearest nullable ancestor field. This can end up causing a client to lose more data farther up the response tree, even though that data may have been resolved without errors.
```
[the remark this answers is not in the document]
DrewML: Pushed an update to consolidate this language. Thanks for pointing that out!

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,206 @@
...
+    options: ProductOptions!
+}
+```
+
+### List Types
+
+List fields have [2 distinct forms of nullability](http://spec.graphql.org/draft/#sec-Combining-List-and-Non-Null):
+
+1. Field nullability (same as all other fields)
+2. List nullability (whether a list can have `null` items inside it)
```
paliarush, line 168: How about the following? If modified should be fixed in other places where "List nullability" is mentioned.

```suggestion
2. List item nullability (whether a list can have `null` items inside it)
```

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,206 @@
...
+Because of error propagation, if an error occurs in a non-null field on the `Query` root, the client will lose data from all other top-level queries.
+
+```graphql
+type Query {
+    # If a client queries for both fields, and one of them fails,
+    # the client will receive no data
+    product(id: ID): Product!
+    category(id: ID): Category!
+}
+```
```
prabhuram93, line 122: Hi @DrewML much needed document! Can you also mention best practices for subqueries. Can we generalize that the return type of any query, be it a parent or a subquery, can be nullable? 
I see some conflicts in the upcoming features
Screen Shot 2020-07-06 at 4 56 30 PM

DrewML: Thanks @prabhuram93 for the review and feedback!

My hope/intent was that the case you've raised would be covered by this section:

image

In your example screenshot, I can think of many cases where a response for part of `Customer` would still be valuable if `Customer.reviews` failed, so ideally `Customer.reviews` _would_ be nullable.

Can you think of what wording we might be able to change to make that stand out more?

DrewML:
Added a section with some guidance on scalars based on feedback from @paales

DrewML:
> Please correct indentation in the examples (verify by opening the "rich diff": #392 (files) )

@paliarush pushed a fix for the formatting

design-documents/graph-ql/nullability.md
DrewML: Good suggestion! Pushed updates

jimbo:
Makes sense to me. I'll concur with the takeaway from @buskamuza here:

> field may be nullable if it is not critical for the parent object

It's worth noting that for many list fields, if we were to query for them independently, the resulting list would have to be nullable anyway because of the rule you defined earlier:

> Top-level Query fields should always be nullable

For example, if we requested `RelatedProducts` using a top-level query with `productId` as an argument, the result would have to be nullable. So we might as well also make such lists nullable when they're included as fields in other queries.

design-documents/graph-ql/nullability.md
prabhuram93: The recommendations are clear and does stand out. Since you were very specific about Top level Queries, I just wanted to imply/consult that the same is applicable for queries at any level.

DrewML: > I just wanted to imply/consult that the same is applicable for queries at any level

Definitely! 😄 

I tried to avoid getting too prescriptive about nested fields because the rules there depend a lot more on the domain object (i.e. is a product that has options useful without its options data?).

The reason for going prescriptive with root queries was because errors bubbling up to that point and finding a non-null would lose other, likely unrelated data.

More than happy to add in some clarifications if you think they'd help though! Hard not to have tunnel vision with your own writing 😅

DrewML:
We've had this open for 2 weeks with several proposals. Going to merge, but if new concerns arise please open an issue or a follow-up PR and we can iterate.

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,216 @@
...
+- It is a safe, non-breaking change to move a field from nullable to non-nullable
+- It is an unsafe, breaking change to move a field from non-nullable to nullable
+
+With this in mind, when in doubt, the safer route is to make a field nullable, with the agreement that we'll iterate and improve schemas as we get feedback on usage from clients.
+
+**Note**: The [GraphQL reference implementation](https://github.com/graphql/graphql-js) has a `findBreakingChanges` feature that can be inspected to see a [list of all known schema changes that cause breaks for clients](https://github.com/graphql/graphql-js/blob/2232ebdef828566f3add3ed2a31709d3c1710c0e/src/utilities/findBreakingChanges.js#L41-L67). It [explicitly notes](https://github.com/graphql/graphql-js/blob/2232ebdef828566f3add3ed2a31709d3c1710c0e/src/utilities/findBreakingChanges.js#L461) that moving from a nullable value to a non-nullable value of the same type is safe.
+
+## Recommendations
+
+### Top-level Query fields should always be nullable
```
MartijnHols, line 111: As a M2 GraphQL consumer I disagree. If a query is guaranteed to return data, it should not be null and making it nullable introduces unnecessary defensive code, slowing down development or worse risking null-errors due to excessive use of non-null-assertions.

Most of the time queries returning arrays will always return an array but empty when there are no results. e.g `countries: [Country!]!`. In this case I don't think there is any reason why `countries` should ever be null. There are probably others where this is the case.

I don't think there are a lot of queries (or mutations) in M2 that return a non-nullable singular type. I think `currency` and `storeConfig` are two that should never return `null`. I reckon the recommendation should be to look at this case-by-case. Anything that takes a param such as id to search for an item should be nullable (as the id may not exist and `null` is usually used for "not found" in GraphQL).

design-documents/graph-ql/nullability.md
```diff
@@ -0,0 +1,216 @@
...
+    foo3: [Foo!]  # foo3 can be a list or null. If list, every item must be a 'Foo'
+    foo4: [Foo]!  # foo4 must be a list, and the list can have null values
+}
+```
+
+To decide whether the _field_ should be nullable, use the other recommendations provided in this document.
+
+When deciding whether _List items_ should be nullable, the most important question to ask is:
+
+**"Is the parent object still usable if at least one item in this list has an error?"**
```
MartijnHols, line 194: This is very inconvenient, requiring each iterator to check for null which is quite a PITA for functional programming often used in React with TypeScript. If a child has a GraphQL error (and I hope the goal is to have 0 cases of that), I would much prefer the entire field to get an error than to have to work around nullability within every single array where it never happens that an item is actually null.

PR #393, opened 2020-06-26
Add document with suggested guidelines around mutation error design

DrewML:
[Rendered Document](/magento/architecture/blob/mutation-error-design/design-documents/graph-ql/mutation-error-design.md)

cc @fnhipster @sirugh who I've had recent discussions with re: error discoverability for headless UI

Some related blog posts and ideas:
- https://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/
- https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/
- https://graphqlme.com/2018/02/24/handling-errors-in-graphql/

### DIFF
--- a/design-documents/graph-ql/mutation-error-design.md
+++ b/design-documents/graph-ql/mutation-error-design.md
@@ -74,7 +74,7 @@ type AddItemToCartOutput {
     # can still get the entire state of the cart post-mutation
     cart: Cart
     # can directly access errors that a shopper should be notified about
-    addItemUserErrors: [AddItemUserError!]!
+    add_item_user_errors: [AddItemUserError!]!
 }
 
 type AddItemUserError {
@@ -104,7 +104,7 @@ mutation {
             }
         }
 
-        addItemUserErrors {
+        add_item_user_errors {
             message
             type
         }
@@ -118,7 +118,7 @@ mutation {
         "cart": {
             // cart data here
         },
-        "addItemUserErrors": [{
+        "add_item_user_errors": [{
             "message": "Item 'cool-hat' not added to cart (Out of Stock)",
             "type": "OUT_OF_STOCK"
         }]
@@ -150,5 +150,5 @@ Instead, you would likely have your `addItemToCart` function return something th
 
 ## Open Questions
 
-1. Should we be consistent with the field name that represents these first-class errors? `userErrors` vs something like `addToCartUserErrors`
+1. Should we be consistent with the field name that represents these first-class errors? `user_errors` vs something like `add_to_cart_user_errors`
 2. Should there be a basic interface that mutations should implement to enforce the pattern of returning a list of user errors?
--- a/design-documents/graph-ql/mutation-error-design.md
+++ b/design-documents/graph-ql/mutation-error-design.md
@@ -14,7 +14,7 @@ type Mutation {
 }
 ```
 
-This schema only describes one the happy path of adding an item to the cart. But some things can go wrong when adding an item to the cart, and we know what many of those things are:
+This schema only describes the happy path of adding an item to the cart. But some things can go wrong when adding an item to the cart, and we know what many of those things are:
 
 - Item went out of stock since the product page was loaded
 - Merchant disabled the product since the product page was loaded
--- a/design-documents/graph-ql/mutation-error-design.md
+++ b/design-documents/graph-ql/mutation-error-design.md
@@ -1,0 +1,153 @@
+# Mutation Error Design
+
+Mutations are _the_ single way to trigger side-effects in a GraphQL API. Side-effects are where the large majority of our errors are going to happen, as a result of changing some underlying data.
+
+Today, the Magento 2 GraphQL schema does not do a great job of describing what could go wrong as a result of running a mutation. We should fix that!
+
+## Example of the problem
+
+Take this simplified schema to add an item to a shopper's cart:
+
+```graphql
+type Mutation {
+    addItemToCart(sku: String!, quantity: Float): Cart
+}
+```
+
+This schema only describes one side-effect that can happen as a result of adding an item to the cart: it succeeds! But some things can go wrong when adding an item to the cart, and we know what many of those things are:
+
+- Item went out of stock since the product page was loaded
+- Merchant disabled the product since the product page was loaded
+- Cart hit the max quantity allowed of item per customer. Full qty selected was not added, but some were
+- Extensions can add additional rules that would limit adding an item to the cart
+
+Now put yourself in the shoes of a UI developer: how do you know all the error states your UI needs to cover? For most of the Magento 2 mutations today, you would need run the mutation itself (or dig through PHP) to see what will be returned in the `errors` key of the GraphQL response:
+
+```js
+{
+    "data": {
+        "cart": {
+            // cart data here
+        }
+    },
+    "errors": [{
+        "message": "Item 'cool-hat' not added to cart (Out of Stock)",
+        "path": ["addItemToCart"]
+    }]
+}
+```
+
+This has some problems:
+
+1. It's unreasonable to expect a UI developer to exercise every possible input to a resolver to find the error states manually
+2. These errors are not enforced by the schema, so they're not versioned with the schema.
+3. If the UI wants to customize the message for a specific error state, they'd have to match on the exact response string, because there's no concept of error codes
+
+Until recently, we didn't need to think quite as carefully about modeling errors and versioning them, because we owned the single reference theme for the application (Luma). In the world of headless UIs, it's critical that our errors are versioned and discoverable.
+
+## Solution: Design Mutation errors into the schema
+
+If we move our known error states for various mutations into the schema design itself, we'll get a few benefits:
+
+1. Versioning of errors
+2. Discoverability of error states for UI devs
+
+### Example: Fixing our `addItemToCart` mutation
+
+In the previous example of `addItemToCart`, we had the following schema:
+
+```graphql
+type Mutation {
+    addItemToCart(sku: String!, quantity: Float): Cart
+}
+```
+
+Let's design some known error states directly into the API
+
+```graphql
+type Mutation {
+    addItemToCart(sku: String!, quantity: Float): AddItemToCartOutput
+}
+
+type AddItemToCartOutput {
+    # can still get the entire state of the cart post-mutation
+    cart: Cart
+    # can directly access errors that a shopper should be notified about
+    addItemUserErrors: [AddItemUserError!]!
+}
+
+type AddItemUserError {
+    # if a UI has their own text for message, they can just
+    # not use this field, but serves as a descriptive default
+    message: string!
+    type: AddItemUserErrorType!
+}
+
+# This enum is just an example. Non-exhaustive, and of course
+# not all errors we'd want a specific type for
+enum AddItemUserErrorType {
+    OUT_OF_STOCK
+    MAX_QTY_FOR_USER
+    NOT_AVAILABLE
+}
+```
+
+With this new schema, we get the following query and response:
+
+```graphql
+query {
+    addItemToCart(sku: "abc", quantity: 1) {
+        cart {
+            items {
+                # select item fields
+            }
+        }
+
+        addItemUserErrors {
+            message
+            type
+        }
+    }
+}
+```
+
+```js
+{
+    "data": {
+        "cart": {
+            // cart data here
+        },
+        "addItemUserErrors": [{
+            "message": "Item 'cool-hat' not added to cart (Out of Stock)",
+            "type": "OUT_OF_STOCK"
+        }]
+    }
+}
+```
+## Should we still use the `errors` key at all?
+
+Yes! _But_, we should consider the `errors` key to be of similar use to the `catch` clause when using `try/catch` in a programming language.
+
+The `errors` key should be for _exceptional_ circumstances, not for describing well-known states in an ecommerce app. It's unlikely you'd write this application code on the backend:
+
+```javascript
+try {
+    var product = addItemToCart('sku');
+    return product;
+} catch (error) {
+    if (error.code === 'OUT_OF_STOCK') {
+        //
+    } else if (error.code === 'MAX_QTY_FOR_USER') {
+        //
+    } else {
+        // something else failed, maybe the DB connection?
+    }
+}
+```
+
+Instead, you would likely have your `addItemToCart` function return something that describes the various possible results of the operation. 
+
+## Open Questions
+
+1. Should we be consistent with the field name that represents these first-class errors? `userErrors` vs something like `addToCartUserErrors`
+2. Should there be a basic interface that mutations should implement to enforce the pattern of returning a list of user errors?
--- a/design-documents/graph-ql/mutation-error-design.md
+++ b/design-documents/graph-ql/mutation-error-design.md
@@ -1,6 +1,6 @@
 # Mutation Error Design
 
-Mutations are _the_ single way to trigger side-effects in a GraphQL API. Side-effects are where the large majority of our errors are going to happen, as a result of changing some underlying data.
+Mutations are _the_ single way to modify application state in a GraphQL API. Mutations are where the large majority of our errors are going to happen, as a result of changing some underlying data.
 
 Today, the Magento 2 GraphQL schema does not do a great job of describing what could go wrong as a result of running a mutation. We should fix that!
 
@@ -14,7 +14,7 @@ type Mutation {
 }
 ```
 
-This schema only describes one side-effect that can happen as a result of adding an item to the cart: it succeeds! But some things can go wrong when adding an item to the cart, and we know what many of those things are:
+This schema only describes one the happy path of adding an item to the cart. But some things can go wrong when adding an item to the cart, and we know what many of those things are:
 
 - Item went out of stock since the product page was loaded
 - Merchant disabled the product since the product page was loaded
@@ -42,8 +42,9 @@ This has some problems:
 1. It's unreasonable to expect a UI developer to exercise every possible input to a resolver to find the error states manually
 2. These errors are not enforced by the schema, so they're not versioned with the schema.
 3. If the UI wants to customize the message for a specific error state, they'd have to match on the exact response string, because there's no concept of error codes
+4. Internationalization becomes challenging because the error states/strings are not known ahead of time
 
-Until recently, we didn't need to think quite as carefully about modeling errors and versioning them, because we owned the single reference theme for the application (Luma). In the world of headless UIs, it's critical that our errors are versioned and discoverable.
+In the world of headless UIs, it's going to be critical for our errors to be discoverable, translatable, and versioned
 
 ## Solution: Design Mutation errors into the schema
 
--- a/design-documents/graph-ql/mutation-error-design.md
+++ b/design-documents/graph-ql/mutation-error-design.md
@@ -95,7 +95,7 @@ enum AddItemUserErrorType {
 With this new schema, we get the following query and response:
 
 ```graphql
-query {
+mutation {
     addItemToCart(sku: "abc", quantity: 1) {
         cart {
             items {


DrewML:
Looks like this pattern has already appeared before :)

https://github.com/magento/magento2/pull/27914#issuecomment-642847466

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,153 @@
...
+        // something else failed, maybe the DB connection?
+    }
+}
+```
+
+Instead, you would likely have your `addItemToCart` function return something that describes the various possible results of the operation. 
+
+## Open Questions
+
+1. Should we be consistent with the field name that represents these first-class errors? `userErrors` vs something like `addToCartUserErrors`
```
sirugh, line 152: I don't think it's necessary to namespace the first class errors.

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,153 @@
...
+2. Should there be a basic interface that mutations should implement to enforce the pattern of returning a list of user errors?
```
sirugh, line 153: Sounds reasonable if the property isn't namespaced :)

paliarush:
> Looks like this pattern has already appeared before :)
> 
> https://github.com/magento/magento2/blob/b9e28fd3dc2b32a03984f4f62fc5cf612842932f/app/code/Magento/SalesGraphQl/etc/schema.graphqls#L27

Yes, it was mentioned in the proposal, but limited to the specific use case needing it, without generalization: https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer-reorder.graphqls

DrewML:
> Yes, it was mentioned in the proposal, but limited to the specific use case needing it, without generalization

That's great! I like the strategy of trying things out before we adopt. Wonder if anyone has feedback on that API so far.

design-documents/graph-ql/mutation-error-design.md
DrewML: > Sounds reasonable if the property isn't namespaced :)

Haha, very good point. So that's one argument _for_ a consistent name.

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,153 @@
+# Mutation Error Design
+
+Mutations are _the_ single way to trigger side-effects in a GraphQL API. Side-effects are where the large majority of our errors are going to happen, as a result of changing some underlying data.
```
paliarush, line 3: Please see https://graphql.org/learn/queries/#mutations . I believe "side effect" is usually referring to unexpected effect like data modification during query.

```suggestion
Mutations should be _the_ single way used to modify server-side data in a GraphQL API. The large majority of our errors are going to happen during mutations, as a result of changing some underlying data.
```

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+Let's design some known error states directly into the API
+
+```graphql
+type Mutation {
+    addItemToCart(sku: String!, quantity: Float): AddItemToCartOutput
+}
+
+type AddItemToCartOutput {
```
paliarush, line 72: Wrappers for mutation results have been required for quite some time https://github.com/magento/architecture/blob/master/design-documents/graph-ql/extensibility.md and are implemented for most of the mutations in Magento. I would remove example above where the Cart is returned directly.

design-documents/graph-ql/mutation-error-design.md
paliarush: It does not seem to be possible if we want to have custom error code enums for each mutation. Please provide an example of how it can look like.

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,153 @@
...
+    # can still get the entire state of the cart post-mutation
+    cart: Cart
+    # can directly access errors that a shopper should be notified about
+    addItemUserErrors: [AddItemUserError!]!
+}
+
+type AddItemUserError {
+    # if a UI has their own text for message, they can just
```
paliarush, line 80: Let's add `path` and `location`, see https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer-reorder.graphqls

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,153 @@
+# Mutation Error Design
+
+Mutations are _the_ single way to trigger side-effects in a GraphQL API. Side-effects are where the large majority of our errors are going to happen, as a result of changing some underlying data.
+
+Today, the Magento 2 GraphQL schema does not do a great job of describing what could go wrong as a result of running a mutation. We should fix that!
```
paliarush, line 5: While I see the value of providing a list of possible error codes in the schema (particularly for green field application), it may be problematic to extract all of them from the implementation of existing PHP-resolvers which rely on existing models and services. Those models and services do not properly declare exceptions in their signatures. It means that it will be extremely difficult to list all error codes for the specific mutation, we can discuss if partial list makes sense.
For the same reason I believe we cannot introduce a shared interface for all mutation output types.

In scope of https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer-reorder.graphqls we agreed that custom error types make sense for validation errors (when multiple errors can happen during the same mutation)

design-documents/graph-ql/mutation-error-design.md
DrewML: The language used is from the GraphQL specification's description of mutations:

> Mutation
If the operation is a mutation, the result of the operation is the result of executing the mutation’s top level selection set on the mutation root object type. This selection set should be executed serially.

> It is expected that the top level fields in a mutation operation perform side‐effects on the underlying data system. Serial execution of the provided mutations ensures against race conditions during these side‐effects.

http://spec.graphql.org/June2018/

design-documents/graph-ql/mutation-error-design.md
DrewML: > It does not seem to be possible if we want to have custom error code enums for each mutation. Please provide an example of how it can look like.


```graphql
type Mutation {
    addToCart(sku: String!): AddToCartOutput
    removeFromCart(sku: String!): RemoveFromCartOutput
}

type AddToCartOutput implements MutationOutput {
    cart: Cart
    userErrors: [AddToCartError!]
}

type RemoveFromCartOutput implements MutationOutput {
    removedItems: [Product!]
    userErrors: [RemoveFromCartError!]
}

type AddToCartError implements UserError {
    field: [String!]
    message: String!
    code: CartAddError
}

type RemoveFromCartError implements UserError {
    field: [String!]
    message: String!
    code: CartRemoveError
}

interface UserError {
    field: [String!]
    message: String!
}

interface MutationOutput {
    userErrors: [UserError!]
}

enum CartAddError {
    OUT_OF_STOCK
    NOT_ALLOWED
    BAD_CART_ID
}

enum CartRemoveError {
    NOT_IN_CART
}
```

design-documents/graph-ql/mutation-error-design.md
DrewML: > While I see the value of providing a list of possible error codes in the schema (particularly for green field application), it may be problematic to extract all of them from the implementation of existing PHP-resolvers which rely on existing models and services

Sure! I'm not suggesting we surface every possible error state known. But for many operations, the common set of failures is both obvious and relevant to a shopper. If we have `addProductToCart`, we know the obvious causes are bad sku and out of stock.

>  we can discuss if partial list makes sense.

Yep, that's what I'm advocating for!

> For the same reason I believe we cannot introduce a shared interface for all mutation output types

Provided a [sample schema](https://github.com/magento/architecture/pull/393#discussion_r453972900) in another comment. Let me know your thoughts.

DrewML:
Here's an example of where the `pwa-studio` team currently has to hardcode strings from the `errors` list to identify errors

https://github.com/magento/pwa-studio/pull/2576#discussion_r460193166

design-documents/graph-ql/mutation-error-design.md
danielrenaud: Agree. I might call it `mutationErrors`, but in any case a single global type should suffice.

design-documents/graph-ql/mutation-error-design.md
paliarush: Still not convinced. The intention, the expected effect of the mutation is state modification. Why do we call it side-effect in this case? 

Which of these (from the wikipedia) cases do we have?
```
Example side effects include modifying a non-local variable, modifying a static local variable, modifying a mutable argument passed by reference, performing I/O or calling other side-effect functions.
```

DrewML: > an operation, function or expression is said to have a side effect if it modifies some state variable value(s) outside its local environment, that is to say has an observable effect besides returning a value (the main effect) to the invoker of the operation

Note that the terminology of side-effects is also used in the HTTP specification when describing "safe methods." We're just sharing the common terminology both of these specifications use.

DrewML:
Meeting scheduled for August 4th with wider audience to discuss

design-documents/graph-ql/mutation-error-design.md
buskamuza: Agree, that side-effect is usually referred to any state modifications. It's different from the same non-science term (which has a negative connotation).

DrewML: I replaced usage of the term "side-effect" with something that should be clearer 😃

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -14,7 +14,7 @@ type Mutation {
 }
 ```
 
-This schema only describes one side-effect that can happen as a result of adding an item to the cart: it succeeds! But some things can go wrong when adding an item to the cart, and we know what many of those things are:
+This schema only describes one the happy path of adding an item to the cart. But some things can go wrong when adding an item to the cart, and we know what many of those things are:
```
paliarush, line 17:
```suggestion
This schema only describes the happy path of adding an item to the cart. But some things can go wrong when adding an item to the cart, and we know what many of those things are:
```

design-documents/graph-ql/mutation-error-design.md
paliarush: @DrewML I see `path` in the updated schema, what about `location`?

maghamed:
The proposal makes sense to me.
There are two things I want to point out, which need to be defined more explicitly.
1. Taking into account that we are not deprecating existing codes, and will use them for “unchecked” or “non-recoverable” error execution. We need to define for which cases we going to use error codes and for which we would use Error mutation output.  Here we might follow a similar description which Java has with checked/unchecked exceptions.
      Usually, the definition of checked exceptions is for recoverable conditions and runtime exceptions for programming errors (like file does not exist or connection lost, etc).
       Out of stock error is a good example of a recoverable or expected situation as this is part of the business logic the client code should be aware of.
      So, it would be great to introduce explicit differentiation to follow
2.   The strategy of migration. The point which @paliarush  mentioned - if we start introducing new Mutation Output for some mutation, we need to handle all the cases to make it compliant with the proposed approach. Thus, we need to avoid a situation when we have a transient state which does not follow neither existing approach nor desirable state on the level of one mutation.

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+```
+
+Instead, you would likely have your `addItemToCart` function return something that describes the various possible results of the operation. 
+
+## Open Questions
+
+1. Should we be consistent with the field name that represents these first-class errors? `userErrors` vs something like `addToCartUserErrors`
+2. Should there be a basic interface that mutations should implement to enforce the pattern of returning a list of user errors?
```
billkalter, line 154: Great document!  Would it make sense to incorporate union types rather than relying on each response to include an error portion?  This allows for common errors to be standardized across multiple calls in a uniform way, such as extending your example with multiple methods such as "addToCart", "addAllToCart", "putWishlistItemInCart", and so on.  This could be handled using interfaces as previously suggested but union type force the caller to acknowledge the error case.  It also provides for cleaner separation between happy path and failure path objects. 
 For example:

```graphql
type Mutation {
    addItemToCart(sku: String!, quantity: Float): AddItemToCartOutput
    putWishlistItemInCart(wishlistId: String!, sku: String): WishlistItemMovedToCartOutput
}

type CartUpdatedOutput {
    cart: Cart
}

type WishlistUpdatedOutput {
    cart: Cart
    wishlist: Wishlist
}

type CartUpdateError {
    message: string!
    reasons: [CartUpdateErrorReason!]!
}

union AddItemToCartOutput = CartUpdatedOutput | CartUpdateError
union WishlistItemMovedToCartOutput = WishlistUpdatedOutput | CartUpdateError

# and so on...
```

design-documents/graph-ql/mutation-error-design.md
prabhuram93: @DrewML This looks good https://github.com/magento/architecture/pull/393/files#r453972900. It makes sense to standardize this approach. For alignment with existing reorder mutation and convention, can we use 'code' instead of 'type'. Also another suggestion came up Alex when we discussed this, field names should be snake cased. So suggesting 'user_errors' instead of 'userErrors'.

design-documents/graph-ql/mutation-error-design.md
DrewML: Thanks for the feedback @billkalter! I'm a big fan of this approach, and it's similar to [the approach that artsy took](https://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/) for their API. If I was building a new GraphQL API today, this is likely the approach I'd go with.

The main reason I opted not to suggest this approach is that we already have a fair number of mutations today, and we'd have to deprecate each one and introduce a new mutation if we wanted to change the result type to a union.

Because we already have requirements around so-called "output objects" for all mutations (missed on a few very old ones), I thought it would make sense for us to piggyback on this and avoid the breaking changes.

Thoughts?

<details>
<summary>Click to see Mutation type from Magento core</summary>

```graphql
    type Mutation {
        addBundleProductsToCart(
            input: AddBundleProductsToCartInput
        ): AddBundleProductsToCartOutput
        addConfigurableProductsToCart(
            input: AddConfigurableProductsToCartInput
        ): AddConfigurableProductsToCartOutput
        addDownloadableProductsToCart(
            input: AddDownloadableProductsToCartInput
        ): AddDownloadableProductsToCartOutput
        addSimpleProductsToCart(
            input: AddSimpleProductsToCartInput
        ): AddSimpleProductsToCartOutput
        addVirtualProductsToCart(
            input: AddVirtualProductsToCartInput
        ): AddVirtualProductsToCartOutput
        applyCouponToCart(
            input: ApplyCouponToCartInput
        ): ApplyCouponToCartOutput

        """
        Apply a pre-defined gift card code to the specified cart.
        """
        applyGiftCardToCart(
            input: ApplyGiftCardToCartInput
        ): ApplyGiftCardToCartOutput

        """
        Apply store credit to the specified cart.
        """
        applyStoreCreditToCart(
            input: ApplyStoreCreditToCartInput!
        ): ApplyStoreCreditToCartOutput

        """
        Changes the password for the logged-in customer
        """
        changeCustomerPassword(
            currentPassword: String!
            newPassword: String!
        ): Customer

        """
        Creates Client Token for Braintree Javascript SDK initialization.
        """
        createBraintreeClientToken: String!

        """
        Create customer account
        """
        createCustomer(input: CustomerInput!): CustomerOutput

        """
        Create customer address
        """
        createCustomerAddress(input: CustomerAddressInput!): CustomerAddress

        """
        Creates an empty shopping cart for a guest or logged in user
        """
        createEmptyCart(input: createEmptyCartInput): String

        """
        Initiates a transaction and receives a token. Use this mutation for Payflow Pro and Payments Pro payment methods
        """
        createPayflowProToken(
            input: PayflowProTokenInput!
        ): CreatePayflowProTokenOutput

        """
        Initiates an Express Checkout transaction and receives a token. Use this
        mutation for Express Checkout and Payments Standard payment methods.
        """
        createPaypalExpressToken(
            input: PaypalExpressTokenInput!
        ): PaypalExpressTokenOutput

        """
        Delete customer address
        """
        deleteCustomerAddress(id: Int!): Boolean

        """
        Delete a customer payment token
        """
        deletePaymentToken(public_hash: String!): DeletePaymentTokenOutput

        """
        Retrieve the customer token
        """
        generateCustomerToken(email: String!, password: String!): CustomerToken

        """
        Handles payment response and saves payment in Quote. Use this mutations for Payflow Pro and Payments Pro payment methods.
        """
        handlePayflowProResponse(
            input: PayflowProResponseInput!
        ): PayflowProResponseOutput

        """
        Merges the source cart into the destination cart
        """
        mergeCarts(source_cart_id: String!, destination_cart_id: String!): Cart!
        placeOrder(input: PlaceOrderInput): PlaceOrderOutput

        """
        Redeem gift card for store credit.
        """
        redeemGiftCardBalanceAsStoreCredit(
            input: GiftCardAccountInput!
        ): GiftCardAccount
        removeCouponFromCart(
            input: RemoveCouponFromCartInput
        ): RemoveCouponFromCartOutput
        removeGiftCardFromCart(
            input: RemoveGiftCardFromCartInput
        ): RemoveGiftCardFromCartOutput
        removeItemFromCart(
            input: RemoveItemFromCartInput
        ): RemoveItemFromCartOutput

        """
        Remove applied store credit from the specified cart.
        """
        removeStoreCreditFromCart(
            input: RemoveStoreCreditFromCartInput!
        ): RemoveStoreCreditFromCartOutput

        """
        Adds all products from a customer's previous order to the cart.
        """
        reorderItems(orderNumber: String!): ReorderItemsOutput

        """
        Revoke the customer token
        """
        revokeCustomerToken: RevokeCustomerTokenOutput

        """
        Recommends Product by Sending Single/Multiple Email
        """
        sendEmailToFriend(
            input: SendEmailToFriendInput
        ): SendEmailToFriendOutput
        setBillingAddressOnCart(
            input: SetBillingAddressOnCartInput
        ): SetBillingAddressOnCartOutput
        setGuestEmailOnCart(
            input: SetGuestEmailOnCartInput
        ): SetGuestEmailOnCartOutput
        setPaymentMethodAndPlaceOrder(
            input: SetPaymentMethodAndPlaceOrderInput
        ): PlaceOrderOutput
            @deprecated(
                reason: "Should use setPaymentMethodOnCart and placeOrder mutations in single request."
            )
        setPaymentMethodOnCart(
            input: SetPaymentMethodOnCartInput
        ): SetPaymentMethodOnCartOutput
        setShippingAddressesOnCart(
            input: SetShippingAddressesOnCartInput
        ): SetShippingAddressesOnCartOutput
        setShippingMethodsOnCart(
            input: SetShippingMethodsOnCartInput
        ): SetShippingMethodsOnCartOutput
        updateCartItems(input: UpdateCartItemsInput): UpdateCartItemsOutput

        """
        Update the customer's personal information
        """
        updateCustomer(input: CustomerInput!): CustomerOutput

        """
        Update customer address
        """
        updateCustomerAddress(
            id: Int!
            input: CustomerAddressInput
        ): CustomerAddress
    }
```
</details>

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,154 @@
...
+        // something else failed, maybe the DB connection?
+    }
+}
+```
+
+Instead, you would likely have your `addItemToCart` function return something that describes the various possible results of the operation. 
+
+## Open Questions
+
+1. Should we be consistent with the field name that represents these first-class errors? `userErrors` vs something like `addToCartUserErrors`
```
melnikovi, line 153:
```suggestion
1. Should we be consistent with the field name that represents these first-class errors? `user_errors` vs something like `add_to_cart_user_errors`
```

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+
+```js
+{
+    "data": {
+        "cart": {
+            // cart data here
+        },
+        "addItemUserErrors": [{
```
melnikovi, line 121:
```suggestion
        "add_item_user_errors": [{
```

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,154 @@
...
+```graphql
+mutation {
+    addItemToCart(sku: "abc", quantity: 1) {
+        cart {
+            items {
+                # select item fields
+            }
+        }
+
+        addItemUserErrors {
```
melnikovi, line 107:
```suggestion
        add_item_user_errors {
```

design-documents/graph-ql/mutation-error-design.md
```diff
@@ -0,0 +1,154 @@
...
+```graphql
+type Mutation {
+    addItemToCart(sku: String!, quantity: Float): AddItemToCartOutput
+}
+
+type AddItemToCartOutput {
+    # can still get the entire state of the cart post-mutation
+    cart: Cart
+    # can directly access errors that a shopper should be notified about
+    addItemUserErrors: [AddItemUserError!]!
```
melnikovi, line 77:
```suggestion
    add_item_user_errors: [AddItemUserError!]!
```

PR #395, opened 2020-07-08
Add document with suggestions for ID fields

DrewML:
[Rendered Document](/magento/architecture/blob/id-fields/design-documents/graph-ql/id-fields-object-types.md)

### DIFF
--- a/design-documents/graph-ql/id-fields-object-types.md
+++ b/design-documents/graph-ql/id-fields-object-types.md
@@ -18,7 +18,7 @@ When a GraphQL Object Type is used to represent an entity that can be referenced
 - Include info in the client-facing description describing how the field is encoded/decoded (should be opaque)
     - Example: The client shouldn't know if an `ID` is a base64-encoded integer
 - Use `String` or `Int` type
-- Use duplicate IDs across a concrete type. In other words, if the client wants to produce a cache key, the concetenation of a `__typename` + `id` field should _always_ be unique
+- Use duplicate IDs across a concrete type. In other words, if the client wants to produce a cache key, the concatenation of a `__typename` + `id` field should _always_ be unique
 
 ## Examples where an ID is not helpful
 
--- a/design-documents/graph-ql/id-fields-object-types.md
+++ b/design-documents/graph-ql/id-fields-object-types.md
@@ -7,6 +7,7 @@ When a GraphQL Object Type is used to represent an entity that can be referenced
 ### Do
 
 - Use the `ID` scalar type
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/id-fields-object-types.md
```diff
@@ -0,0 +1,23 @@
...
+    - Most GraphQL client libs automatically use this field for caching. Any other field name will require manual caching logic on the client
+        - https://www.apollographql.com/docs/react/caching/cache-configuration/#assigning-unique-identifiers
+        - https://formidable.com/open-source/urql/docs/graphcache/normalized-caching/#key-generation
+- Include an ID field anytime an Object Type _could_ have an ID field
+
+### Do Not
+- Include info in the client-facing description describing how the field is encoded/decoded (should be opaque)
+    - Example: The client shouldn't know if an `ID` is a base64-encoded integer
+- Use `String` or `Int` type
+
```
tjwiebell, line 20: It feels odd to include this, but SelectedConfigurableOption violates this rule, and causes odd behavior with automatic caching. This `id` field is based on the `ConfigurableProductOption.id`, which is not unique across all responses. Doesn't hurt to be explicit, but consider this optional.
```suggestion
- Use an ID value you cannot guarantee is unique across all responses
```

DrewML: > This id field is based on ConfigurableProductOption.id, which is not unique across all responses.

Whoa. Just to make sure I'm reading this right, you're saying that I can get 2 different `ConfigurableProductOption` types from the API that collide on the `id` field, but hold different data?

tjwiebell: Screen Shot 2020-07-08 at 4 31 47 PM

DrewML: > which I didn't consider might also not be unique 

Oh noooooo. Well, you've definitely highlighted that we've got some work to do, but trending in the right direction!

Going to push an update to the document

DrewML: Added this under `Do Not`, let me know if you have suggestions to tweak the wording:

> Use duplicate IDs across a concrete type. In other words, if the client wants to produce a cache key, the concetenation of a `__typename` + `id` field should _always_ be unique.

DrewML:
Added the following under `Do` based on feedback from @prabhuram93:

> Make the `ID` field non-nullable (syntax: `ID!`)

DrewML:
Never have we reached consensus this easily 😂

PR #397, opened 2020-07-09
ECP-766: Returns Schema Design

paliarush:
GraphQL schema design for returns (RMA)


design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+
+type Mutation {
+    createReturn(input: CreateReturnInput!): CreateReturnOutput @doc(description: "Create a new return.")
+    addReturnComment(input: AddReturnCommentInput!): AddReturnCommentOutput @doc(description: "Add a comment to an existing return.")
+    addReturnTracking(input: AddReturnTrackingInput!): AddReturnTrackingOutput @doc(description: "Add tracking information to the return.")
+    removeReturnTracking(input: RemoveReturnTrackingInput!): RemoveReturnTrackingOutput @doc(description: "Remove tracking information from the return.")
+}
+
+input CreateReturnInput {
+    items: [CreateReturnItemInput]!
```
DrewML, line 16:
```suggestion
    items: [CreateReturnItemInput!]!
```

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    comment_text: String
+}
+
+input CreateReturnItemInput {
+    order_item_id: ID!
+    quantity_to_return: Float!
+    selected_custom_attributes: [ID!]
```
DrewML, line 23: Are `selected_custom_attributes` and `entered_custom_attributes` both used to represent the custom fields that a merchant can add to the item return interface?

image


If so, it might be nice to be a bit more specific here in the naming. It took me awhile to figure out the purpose of these input fields

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    entered_custom_attributes: [EnteredCustomAttributeInput!]
+}
+
+input EnteredCustomAttributeInput {
+    id: ID!
+    value: String!
+}
+
+type CreateReturnOutput {
+    return: Return
```
DrewML, line 33: In Luma today, after you create a return, you're taken back to the `My Returns` view. It would be really nice if we could have a `returns: [Return]` here so the UI has the option to grab that view's data without multiple, dependent round trips

DrewML: Basically asking for the same field from `Customer` to also be here:

```graphql
type CreateReturnOutput {
    return: Return
    returns(
        pageSize: Int = 20 @doc(description: "Specifies the maximum number of results to return at once. Defaults to 20."),
        currentPage: Int = 1 @doc(description: "Specifies which page of results to return. The default value is 1."),
    ): Returns @doc(description: "Information about the customer returns.")
}
```

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+enum ReturnItemStatus {
+    PENDING
+    AUTHORIZED
+    RECEIVED
+    APPROVED
+    REJECTED
+    DENIED
+}
+
+type StoreConfig {
```
DrewML, line 187: Should we also be exposing the `sales/magento_rma/enabled_on_product` config? If my understanding is correct (very well could be wrong), that settings let's you open a return for a whole order, but not by line item.

https://github.com/magento/magento2ee/blob/2f1f767fbaeb4b9abc0cbad7fecbc1f070b2ce7f/app/code/Magento/Rma/Helper/Data.php#L637-L642

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+        currentPage: Int = 1 @doc(description: "Specifies which page of results to return. The default value is 1."),
+    ): Returns @doc(description: "Information about the customer returns.")
+    return(id: ID!): Return @doc(description: "Get customer return details by its ID.")
+}
+
+type CustomerOrder {
+    returns(
+        pageSize: Int = 20 @doc(description: "Specifies the maximum number of results to return at once. Defaults to 20."),
+        currentPage: Int = 1 @doc(description: "Specifies which page of results to return. The default value is 1."),
+    ): Returns @doc(description: "Information about the order returns.")
```
DrewML, line 84: I think the description of this field could provide more detail. Unlike `Customer.returns`, this just gives you returned limit to the particular `CustomerOrder` right?

DrewML: How about `Returns on this order` or `Returns associated with this order`?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    contact_name: String
+    street: [String]!
+    city: String!
+    region: Region!
+    postcode: String!
+    country: Country!
+    telephone: String
+}
+
+enum ReturnStatus {
```
DrewML, line 163: There's no `PARTIALLY_REJECTED` or `PARTIALLY_APPROVED` in the devdocs - is this a case of the docs being wrong or the schema being wrong (legit question, I have no clue)?

https://docs.magento.com/user-guide/sales/rma-lifecycle.html#rma-status

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
+type Query{
+    # See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/framework/attributes-metadata.md
+    pageSpecificCustomAttributes(
```
antonkril, line 3: Is this approach already established? I'm not a big fan of having "page" concept in gql

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+}
+
+input AddReturnTrackingInput {
+    return_id: ID!
+    carrier_id: ID!
+    tracking_number: String!
+}
+
+type AddReturnTrackingOutput {
+    return: Return
```
DrewML, line 52: It would be nice if this also included `tracking: ReturnShippingTracking` - client libraries will almost always want to be able to get to the `ID` (at least) of a resource that was just created.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+}
+
+input RemoveReturnTrackingInput {
+    return_tracking_id: ID!
```
DrewML, line 56: This field's value is read by a client from `ReturnShippingTracking.id`. It's a bit hard to search for which type this ID comes from at the moment (I originally thought to look for `type ReturnTracking`. Can we rename either the field or the `ReturnShippingTracking` type, so it's clear the IDs are connected/related?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+
+type Returns {
+    items: [Return] @doc(description: "List of returns")
+    page_info: SearchResultPageInfo @doc(description: "Pagination metadata")
+    total_count: Int @doc(description: "Total count of customer returns")
+}
+
+type Return @doc(description: "Customer return") {
+    id: ID! @doc(description: "Return number")
+    order_id: String! @doc(description: "ID of the order based on which the return was created.")
```
DrewML, line 96: Should probably be an `ID`

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    request_quantity: Float!
+    quantity: Float!
+    status: ReturnItemStatus!
+}
+
+# See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/custom-attributes-container.md
+type CustomAttribute {
+    id: ID!
+    label: String!
+    value: String! @doc(description: "JSON encoded value of the attribute.")
```
DrewML, line 120: The linked document shows the `value` field here as just being a plain old string. Can you explain the comment about JSON-encoding?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+}
+
+type ReturnComment {
+    id: ID! @doc(description: "Comment ID.")
+    created_by: String! @doc(description: "The name of the user who created posted the comment.")
```
DrewML, line 125: Separate firstname/lastname fields tend to be helpful for the client.

Sometimes you want to abbreviate one of the names to accommodate small screen sizes (_A. Paliarush_ vs _Alex Paliarush_)

DrewML: Here's what I did for B2B negotiable quotes

https://github.com/magento/architecture/pull/399/files#diff-4d6fb9dda54abab9b58373010823458bR326-R332

We can unify approaches

design-documents/graph-ql/coverage/returns.md
```diff
@@ -0,0 +1,348 @@
...
+    }
+  }
+}
+```
+
+#### Render return form with dynamic RMA attributes
+
+```graphql
+{
+  pageSpecificCustomAttributes(page_type: RETURN_ITEM_EDIT_FORM) {
```
DrewML, line 154: Can you explain this scenario in terms of the UI it's meant to support? It's not entirely clear to me right now how the UI is intended to use this data.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Yes, these two fields represent EAV attributes of the return item.
We should have common strategy for EAV attributes here, in Customer (not supported yet in GraphQL), Customer address. Plus in product and categories we name them `custom_attributes` in the output, however there are no custom attributes which customer can input.

Not sure how to improve it, please provide concrete suggestion.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: This will complicate server-side implementation, but may make sense from client perspective. I saw you followed the same approach in negotiable quote schema. 
@nrkapoor what do you think?

design-documents/graph-ql/coverage/returns.graphqls
paliarush: I don't think this setting should be exposed directly on the storefront. My idea was that it should affect `items_eligible_for_return`

design-documents/graph-ql/coverage/returns.graphqls
paliarush: I like the second variant.

design-documents/graph-ql/coverage/returns.graphqls
nrkapoor: @DrewML @paliarush I agree with Andrew. It will provide more flexibility.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Yeah, good observation. I found these in the codebase and have already reached out to product and docs team to verify and correct the docs. If the docs are right - they should reach back to us to fix the schema

design-documents/graph-ql/coverage/returns.graphqls
paliarush: We agreed on this some time ago (see the link on line 2), but this is the first implementation. So we can easily reconsider it if you have better ideas.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: See example below (took it from .md file). Isn't this enough?

```graphql
mutation {
  addReturnTracking(
    input: {
      return_id: "000005", 
      carrier_id: "carrier-id", 
      tracking_number: "4234213"
    }
  ) {
    return {
      shipping {
        tracking {
          id
          carier {
            label
          }
          tracking_number
        }
      }
    }
  }
}
```

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Renamed the field.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: It should not be a problem to have multiple `ID` fields in the type, right? It only indicates that it is not human-readable string.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Look for `Here value is JSON-encoded value of the custom attribute.` in https://github.com/magento/architecture/blob/master/design-documents/graph-ql/custom-attributes-container.md
Maybe I do not understand the question.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: @nrkapoor would you like us to extend current schema to have separate First name and Last name for comment author?

design-documents/graph-ql/coverage/returns.md
paliarush: We need to know which attributes need to be shown on the specific page. Not all EAV attributes of return items should be shown on return creation page, for example. 
Admin user can create new attributes and mark them as available for return creation form.

Additional examples are mentioned in comment for other entities:
```
    # See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/framework/attributes-metadata.md
    # PRODUCTS_COMPARE
    # PRODUCTS_LISTING
    # ADVANCED_CATALOG_SEARCH
```

design-documents/graph-ql/coverage/returns.graphqls
nrkapoor: @paliarush yes. that would make it cleaner.

design-documents/graph-ql/coverage/returns.graphqls
DrewML: > It should not be a problem to have multiple ID fields in the type, right? It only indicates that it is not human-readable string.

Yeah, won't be a problem. Just there to make it clear that it's an opaque value

design-documents/graph-ql/coverage/returns.graphqls
DrewML: Isn't `data.return.shipping.tracking` a list? The client would have to filter this list on `tracking_number`, and then update the cache based on that.

Ideally, I think all mutations would at minimum give easy access to the resource it is intended to create.

design-documents/graph-ql/coverage/returns.graphqls
DrewML: image

Which part of this response would be in the JSON-encoded `value` field? Is it an actual JSON object?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type ReturnShippingCarrier {
+    id: ID!
+    label: String!
+}
+
+type ReturnShippingTracking {
+    id: ID!
+    carier: ReturnShippingCarrier!
```
DrewML, line 146:
```suggestion
    carrier: ReturnShippingCarrier!
```

design-documents/graph-ql/coverage/returns.graphqls
DrewML: > This will complicate server-side implementation

I'm a bit curious to hear more about the complications here. Is it that the work is harder to do, or concerns around load on the server?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
+type Query{
+    # See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/framework/attributes-metadata.md
+    pageSpecificCustomAttributes(
+        page_type: CustomAttributesPageEnum!
+    ): CustomAttributeMetadata
+}
+
+type Mutation {
+    createReturn(input: CreateReturnInput!): CreateReturnOutput @doc(description: "Create a new return.")
```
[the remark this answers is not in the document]
DrewML: @paliarush since negotiable quotes also creates a request for the merchant to respond to, we can find similar naming.

I'm using [`requestNegotiableQuote`](https://github.com/magento/architecture/pull/399/files#diff-4d6fb9dda54abab9b58373010823458bR26) in that PR. We could do `requestReturn`.

Open to other suggestions, but unified naming would be good.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,201 @@
...
+    addReturnComment(input: AddReturnCommentInput!): AddReturnCommentOutput @doc(description: "Add a comment to an existing return.")
+    addReturnTracking(input: AddReturnTrackingInput!): AddReturnTrackingOutput @doc(description: "Add tracking information to the return.")
+    removeReturnTracking(input: RemoveReturnTrackingInput!): RemoveReturnTrackingOutput @doc(description: "Remove tracking information from the return.")
```
DrewML, line 12: > Remove tracking information from the return.

Since it's 1 item being removed from a collection, we could probably be more explicit in the description.

**Ideas**:
```suggestion
    removeReturnTracking(input: RemoveReturnTrackingInput!): RemoveReturnTrackingOutput @doc(description: "Remove a single tracked shipment from the return.")

```

```suggestion
    removeReturnTracking(input: RemoveReturnTrackingInput!): RemoveReturnTrackingOutput @doc(description: "Remove a single tracking number from the return.")

```

design-documents/graph-ql/coverage/returns.graphqls
DrewML: Adding link for reference: https://github.com/magento/magento2ee/blob/2f1f767fbaeb4b9abc0cbad7fecbc1f070b2ce7f/app/code/Magento/Rma/Model/Rma/Source/Status.php#L68-L98

I think the docs are wrong

design-documents/graph-ql/coverage/returns.md
```diff
@@ -0,0 +1,350 @@
...
+          }
+          tracking_number
+        }
+      }
+    }
+  }
+}
+```
+
+If the user decides to view the status of the return, it can be retrieved using the following query:
```
DrewML, line 305: > If the user decides to view the status of the return, it can be retrieved using the following query

Think this query is for the status of the tracking on the return for a specific tracking item, right?

design-documents/graph-ql/coverage/returns.md
```diff
@@ -0,0 +1,350 @@
...
+        created_at
+        created_by
+      }
+      items {
+        id
+        product {
+          sku
+          name
+        }
+        custom_attributes {
```
DrewML, line 93: For this scenario, how does the UI know the formats of each different `value` field?

For example, UI developers will want to render a Yes/No field's value differently than a single line text field's value.

design-documents/graph-ql/coverage/returns.md
```diff
@@ -0,0 +1,350 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### Create a return
+
+#### Determine whether any order items are eligible for return
```
DrewML, line 131: re: Anton's comment https://github.com/magento/architecture/pull/397/files#r458432579

> Why separate list instead of marking order items as eligible for return?

For new UIs that aren't Luma, they may want to visually distinguish items that can be returned, while still on the order details page.

It would be a lot simpler for the UI if this data was located in the order line items, rather than in a separate list requires nested loops or client-side normalization.

design-documents/graph-ql/coverage/returns.md
DrewML: Guessing a second round trip has to be made for `customAttributeMetadata`?

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Done.

@DrewML Extra resolvers should be implemented.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: I would like to the full `Return` object for maximum flexibility. 

We can add `ReturnShippingTracking` to the mutation output, but it seems excessive. 
Considering that multiple tracking numbers will be rare I don't see a problem in refreshing the whole list of tracking numbers after the mutation.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: At least, `value` of multiselect attribute should be encoded somehow, there could be more examples of custom types of EAV attributes that need this.

Even though multiselect is not supported by RMA EAV attributes in the native Magento admin UI, I would prefer to keep the agreement that JSON is possible here. 

When the value is a string - it is still just a string in JSON too, so no complication here.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Changed to
```
    author_firstname: String! @doc(description: "First name of the user who posted the comment.")
    author_lastname: String! @doc(description: "Last name of the user who posted the comment.")
```

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    ): Returns @doc(description: "Information about the customer returns.")
+    return(id: ID!): Return @doc(description: "Get customer return details by its ID.")
+}
+
+type CustomerOrder {
+    returns(
+        pageSize: Int = 20 @doc(description: "Specifies the maximum number of results to return at once. Defaults to 20."),
+        currentPage: Int = 1 @doc(description: "Specifies which page of results to return. The default value is 1."),
+    ): Returns @doc(description: "Information about the order returns.")
+    items_eligible_for_return: [OrderItemInterface] @doc(description: "A list of order items eligible for return.")
```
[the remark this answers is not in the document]
paliarush: It will require extra processing on the client side. Also if we introduce pagination for order items and order has a lot of items (B2B), this may hit performance.

design-documents/graph-ql/coverage/returns.graphqls
DrewML: > I would like to have the full Return object for maximum flexibility.

Not suggesting we get rid of the `Return` object, just suggesting we extend the output type.

If a client is adding an item to a collection, I'd expect it to be fairly straight-forward for the client to access just that resource if it wants, without having to grab extra data and filter.

-----

https://www.apollographql.com/docs/react/data/mutations/#making-all-other-cache-updates

> If a mutation modifies multiple entities, or **if it creates or deletes entities**, the Apollo Client cache is not automatically updated to reflect the result of the mutation. To resolve this, your call to useMutation can include an update function.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    status: ReturnStatus @doc(description: "Return status.")
+    shipping: ReturnShipping @doc(description: "Shipping information for the return.")
+    comments: [ReturnComment] @doc(description: "A list of comments posted for the return.")
+    items: [ReturnItem] @doc(description: "A list of items being returned.")
+    available_shipping_carriers: [ReturnShippingCarrier] @doc(description: "A list of shipping carriers available for returns.")
+}
+
+type ReturnItem {
+    id: ID!
+    product: ProductInterface!
```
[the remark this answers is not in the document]
paliarush: Looks like complication to me but may make sense depending on the use cases we want to support. Currently we show product name and sku.

If we would like to show selected options, discounts, quantities (like refunded, returned, shipped etc) then it makes sense to change reference to OrderItem here. @nrkapoor what do you think?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    custom_attributes: [CustomAttribute]
+    request_quantity: Float!
+    quantity: Float!
+    status: ReturnItemStatus!
+}
+
+# See https://github.com/magento/architecture/blob/master/design-documents/graph-ql/custom-attributes-container.md
+type CustomAttribute {
```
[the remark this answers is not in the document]
paliarush: Because it does not exist in `develop` branch yet (even though was discussed and approved in another PR, referenced on the line above) and is required for the Returns schema to work.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
DrewML: > Looks like complication to me but may make sense depending on the use cases we want to support. Currently we show product name and sku.

I think the big value in adding edges to the graph like this is that we don't need to predict all use-cases ahead of time - client will be able to traverse and get what they need from the order line item for their use-case

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,197 @@
...
+    PENDING
+    AUTHORIZED
+    RECEIVED
+    APPROVED
+    REJECTED
+    DENIED
+}
+
+type StoreConfig {
+    sales_magento_rma_enabled: String! @doc(description: "Returns functionality status on the storefront: enabled/disabled.")
```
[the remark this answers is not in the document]
paliarush: We cannot create custom names here. This field name is based on the existing config path.

It is used as a whitelist to determine which configs are allowed to be returned to storefront.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
nrkapoor: @paliarush it would be more flexible just reference OrderItem for sure. That way we can future proof the schema.

design-documents/graph-ql/coverage/returns.md
paliarush: Changed to `If the user decides to view the status of the specific tracking item, it can be retrieved using the following query`

design-documents/graph-ql/coverage/returns.md
paliarush: For now the Idea was the developer should use `pageSpecificCustomAttributes` first to get metadata for the supported attributes.

We can also add metadata required for rendering to the `custom_attributes`.

design-documents/graph-ql/coverage/returns.md
paliarush: @DrewML We can add it to order line items for such use cases. But isn't it simpler for current UI to have a separate list? See my other reply to the Anton's comment.

paliarush: Added support for:

```
{
  customer {
    orders(filter: {number: {eq: "00000008"}}) {
      items {
        items {
          id
          eligible_for_return
        }
      }
    }
  }
}
```

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+}
+
+type Return @doc(description: "Customer return") {
+    id: ID! @doc(description: "Return number")
+    order_id: ID! @doc(description: "ID of the order based on which the return was created.")
+    creation_date: String! @doc(description: "The date when the return was requested.")
+    customer_email: String! @doc(description: "Email of the customer who created the return.")
+    customer_name: String @doc(description: "The name of the person returning the item")
+    status: ReturnStatus @doc(description: "Return status.")
+    shipping: ReturnShipping @doc(description: "Shipping information for the return.")
```
joni-jones, line 109: I think it should be `shipment` not a `shipping`. RMA related with shipments.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+    text: String! @doc(description: "The comment text.")
+}
+
+type ReturnShipping {
+    address: ReturnShippingAddress @doc(description: "Return shipping address, which is specified by the admin.")
+    tracking(id: ID): [ReturnShippingTracking] @doc(description: "Tracking information for all or a single tracking record when ID is provided.")
+}
+
+type ReturnShippingCarrier {
+    id: ID!
```
joni-jones, line 145: What does `ID` mean? Usually, we represent shipping carriers like carrier + shipping method names.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+    text: String!
+    type: ReturnShippingTrackingStatusType!
+}
+
+enum ReturnShippingTrackingStatusType {
+    INFORMATION
+    ERROR
+}
+
+type ReturnShippingAddress {
```
joni-jones, line 166: I think an optional contact email also makes sense here.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+    removeReturnTracking(input: RemoveReturnTrackingInput!): RemoveReturnTrackingOutput @doc(description: "Remove a single tracked shipment from the return.")
+}
+
+input RequestReturnInput {
+    items: [RequestReturnItemInput!]!
+    comment_text: String
+}
+
+input RequestReturnItemInput {
+    order_item_id: ID! @doc(description: "ID of the order associated with the return.")
```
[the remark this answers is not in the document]
joni-jones: What exactly this ID represents? PK from the database?

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+
+type Returns {
+    items: [Return] @doc(description: "List of returns")
+    page_info: SearchResultPageInfo @doc(description: "Pagination metadata")
+    total_count: Int @doc(description: "Total count of customer returns")
+}
+
+type Return @doc(description: "Customer return") {
+    id: ID! @doc(description: "Return number")
+    order_id: ID! @doc(description: "ID of the order based on which the return was created.")
```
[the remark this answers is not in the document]
DrewML: Should probably be the order object - clients can then grab the `ID` from it if they want. Otherwise we have another missing edge in the graph

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+    eligible_for_return: Boolean @doc(description: "Indicates whether the order item is eligible for return.")
+}
+
+type Returns {
+    items: [Return] @doc(description: "List of returns")
+    page_info: SearchResultPageInfo @doc(description: "Pagination metadata")
+    total_count: Int @doc(description: "Total count of customer returns")
+}
+
+type Return @doc(description: "Customer return") {
```
joni-jones, line 102: I think the `Return` type should have `number` similar to other sales entities and we should do all search/filtration queries by number by not by ID, similar to what we do for orders.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
joni-jones: I think it definitely makes sense from developer usage perspective but it will have performance impact on the current implementation as order entity overloaded and very "heavy".

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Added

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
paliarush: Changed to `OrderItem`. 

To make it clear, product info will be limited to sku, sale price, name and url key, since `OrderItem` does not have reference to the product object. It should be enough to render current UI.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
paliarush: It represents ID of the `OrderItemInterface` GraphQL type, otherwise the flow will not work on the client.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+    comment_text: String!
+}
+
+type AddReturnCommentOutput {
+    return: Return
+}
+
+input AddReturnTrackingInput {
+    return_id: ID!
+    carrier_id: ID!
```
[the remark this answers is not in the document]
paliarush: This is an ID returned as part of `ReturnShippingCarrier` while using `available_shipping_carriers` on `Return` type. Should be used for return tracking input forms. and may be used by client to cache available carriers.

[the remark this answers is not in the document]
joni-jones: Yes, I get it but how it will be represented? Usually, we use a carrier code + shipping method code to represent the delivery method. We do not store all carriers in the storage.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: Same question, let's discuss here https://github.com/magento/architecture/pull/397#discussion_r460143090

design-documents/graph-ql/coverage/returns.graphqls
paliarush: This is info from the Magento admin (the address, merchant wants the returns to be sent to), I don't think it is available at the moment.

Do you still think it makes sense to add the field, which we know will never be populated?

joni-jones: If we can't populate it, I don't think it makes sense to add it.

design-documents/graph-ql/coverage/returns.graphqls
paliarush: I was assuming we will use number as ID, but we can separate them.

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -0,0 +1,210 @@
...
+}
+
+type Returns {
+    items: [Return] @doc(description: "List of returns")
+    page_info: SearchResultPageInfo @doc(description: "Pagination metadata")
+    total_count: Int @doc(description: "Total count of customer returns")
+}
+
+type Return @doc(description: "Customer return") {
+    id: ID! @doc(description: "Return number")
```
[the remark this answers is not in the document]
paliarush: okay, let's do it.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
DrewML: > it will have performance impact on the current implementation as order entity overloaded and very "heavy".

It'll be more of a performance impact if the client has to make multiple requests, and spin up a new request thread + run magento bootstrap for each one.

There are ways we can address performance impacts. If the only field requested is the `id`, you can design the resolver to only grab the ID.

DrewML:
Meeting scheduled for Monday with @paliarush and @akaplya to discuss the Custom Attribute schema design in more detail

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
paliarush: It is up to the implementor, from the API point of view it does not matter because ID field is not expected to be human-readable.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
DrewML: Think I asked @nrkapoor awhile back to create a story to clean up our config. It's not ideal that we have `String` fields with the literal values `enabled` and `disabled`

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
joni-jones: It's not up to the implementor, we have logic in Sales module which depends on the format of how shipping method and carrier are stored and represented, currently, it's `carrier_method`. For RMA functionality it should be consistent.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
joni-jones: Usually, for RMA we just show only order number as a reference to original order.

> There are ways we can address performance impacts. If the only field requested is the id, you can design the resolver to only grab the ID.

I'm not sure if we have something similar right now, as far as I know, we extract all entity fields from the database. In the case with an order entity, it's important to take into account because it's a "heavy" entity with a lot of relations.

design-documents/graph-ql/coverage/returns.graphqls
joni-jones: For order and related entities, we use separate fields.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
paliarush: Turns out we already have alases supported. Changed the field name to `returns_enabled`.
Custom types are not supported yet, there is a story to investigate/add support for types other than `String`

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
DrewML: > Usually, for RMA we just show only order number as a reference to original order.

Right, agree that's how it works today in Luma. As we move to becoming a headless platform with a 1st class API, though, the UX requirements won't be set by us anymore, so it'll be our job to make data reachable in the graph.

> I'm not sure if we have something similar right now, as far as I know, we extract all entity fields from the database. In the case with an order entity, it's important to take into account because it's a "heavy" entity with a lot of relations.

Code won't need to load the entire order entity from the database, though; only if that data is requested by the client. If our resolvers are written to load _all_ data, regardless of whether it was selected by the client, sounds more like technical debt we can improve upon.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
DrewML: From what I'm seeing, I think y'all might be discussing slightly separate things.

I think @paliarush is trying to make it clear that the client doesn't need to know how this ID was generated, because it's just an opaque reference for that object in the graph.

I think @joni-jones is trying to say that the implementor of the resolver in Magento core will need to make sure they're encoding/decoding the `ID` value in a similar manner as it's done in other places in the schema.

How wrong am I? 😄

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
joni-jones: > Code won't need to load the entire order entity from the database, though; only if that data is requested by the client. If our resolvers are written to load all data, regardless of whether it was selected by the client, sounds more like technical debt we can improve upon.

I completely agree that the code shouldn't load the entire entity. My concern that in the near feature will have only current business domain implementation which has issues with performance and doesn't have such partial loading.
I think we always can extend the schema and add a reference to the order entity instead of the order number.

design-documents/graph-ql/coverage/returns.graphqls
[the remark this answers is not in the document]
joni-jones: @DrewML you are right) I don't disagree with Alex. My point is following, as our GQL schema not only a schema for storefront developers but also an implementation guide for backend developers, some things we need to say explicitly. Like we did for order/invoice/shipment entities ID field.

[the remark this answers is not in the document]
DrewML: Definitely! Think we've been trying to get better about using inline comments (`#`) in SDL files for implementation-specific notes, and the `description` for consumer-facing docs.

nrkapoor:
@DrewML @paliarush Can we change all 'id' to 'uid' to be consistent with the new design?

PR #402, opened 2020-07-16
Admin Panel Configuration propagation to Storefront

buskamuza:
## Problem

Existing Magento monolith is responsible for the whole cycle of data and user workflow.
Magento Admin Panel contains configuration settings for all areas of the application, including Admin Panel behavior, storefront UI, etc.

With Storefront being a separate service (or set of services), should configuration settings in Magento Admin Panel be propagated to storefront service?
If yes, in what way?

Task: https://github.com/magento/catalog-storefront/issues/154

## Solution

Each case will be different and will require dedicated attention.
Provided are guidelines on how to make a decision in each such case.

### DIFF
--- a/design-documents/storefront/Readme.md
+++ b/design-documents/storefront/Readme.md
@@ -1,0 +1,15 @@
+# Storefront
+
+This section of documents contains architectural and design decisions related to separation of Magento Storefront as a separate application.
+
+## High-Level Vision of Application Separation
+
+![monolith to services separation](https://app.lucidchart.com/publicSegments/view/24d251f7-ba0e-4c90-9190-3f397ad650b2/image.png)
+
+Where **Domain** is application domain, such as Catalog, Customer, etc.
+See [Magento Service Isolation Vision](../service-isolation.md) for more details on domains and services.
+
+## Resources
+
+1. [Magento Service Isolation Vision](../service-isolation.md)
+2. [Catalog Storefront project overview](https://github.com/magento/catalog-storefront/wiki/Catalog-Storefront-Service)
--- a/design-documents/storefront/configuraiton-propagation.md
+++ b/design-documents/storefront/configuraiton-propagation.md
@@ -50,7 +50,7 @@ For example, time for configuration propagation in case reindexation is chosen, 
 
 ### 3. Configuration that impacts UI representation of Storefront data
 
-Example: include or exclude taxes in displayed prices. 
+Example: number of products on products listing page. 
 
 This kind of configuration has nothing to do with Storefront data itself, but is still necessary for the client to know how to display the data.
 
--- a/design-documents/storefront/configuraiton-propagation.md
+++ b/design-documents/storefront/configuraiton-propagation.md
@@ -1,0 +1,65 @@
+# Configuration Propagation from Admin Panel to Storefront
+
+## Problem Statement
+
+Existing Magento monolith is responsible for the whole cycle of data and user workflow.
+Magento Admin Panel contains configuration settings for all areas of the application, including Admin Panel behavior, storefront UI, etc.
+
+With Storefront being a separate service (or set of services), should configuration settings in Magento Admin Panel be propagated to storefront service?
+If yes, in what way?
+
+## Configuration Propagation Guidelines
+
+This document covers general agreements for decisions about configuration propagation from Admin Panel to storefront. 
+A separate design decision should be made and documented for specific use cases.
+
+There are three use cases for the configuration, based on its impact.
+
+### 1. Configuration that impacts Admin Panel behavior
+
+Examples: admin ACLs that allow or deny parts of functionality for the admin user, reindex on update or by schedule.
+
+This type of configuration has nothing to do with Storefront and should not be propagated to storefront, as well as should not be exposed via export API. 
+
+### 2. Configuration that impacts data provided by Storefront
+
+Example: Base Media URL is used for calculation of media image URLs returned by Storefront API.
+
+This type of configuration should be propagated to Storefront in one of the following ways:
+
+1. Pre-calculate final data on back office side and provide final result to the Storefront during synchronization.
+   1. Pros: simpler implementation of Storefront due to eliminated necessity for Storefront to keep knowledge about additional configuration, including data calculation algorithms.
+   2. Cons: massive (up to full) reindexation necessary in case the configuration is changed.
+   
+To choose the right approach in a specific case, consider the following:
+
+1. How frequently the configuration is expected to change?
+   1. Frequent configuration changes leading every time to massive reindexation may be unacceptable.
+   2. Configuration changes expected a few times in the store lifespan may not be worth additional complexity on the Storefront side and full reindexation may be better in this case.
+2. Is it acceptable to have significant delay in data propagation after the configuration change?
+   1. Changing Base URL may be not a big issue, especially if a redirect can be setup. So it may be acceptable to have URLs to be fully updated in a few hours.
+   2. Changes in prices, on the other side, may not stand long delays.
+
+### 3. Configuration that impacts UI representation of Storefront data
+
+Example: include or exclude taxes in displayed prices. 
+
+This kind of configuration has nothing to do with Storefront data itself, and so it is suggested to have a separate service that provides this configuration.
+Client application may call the specialized configuration service to get configuration it needs.
+Some clients may not need all or part of UI configuration.
+GraphQL serves as a single entry point for both data and UI requests to simplify client implementation and have more control of which APIs are publicly exposed.
+
+![UI Configuration Propagation](https://app.lucidchart.com/publicSegments/view/b7ec5763-eb23-48ac-9092-6b92821040fb/image.png)
+
+Pros and cons of a separate service compared to configuration being part of the domain service are described bellow.
+
+Pros:
+
+1. More control of the deployment, scalability, technologies for the domain and config services.
+2. Ability to exclude config service from deployment in case it becomes unnecessary for the specific use case.
+   1. In case of multi-tenant deployment, of course it can't be done based on specific customer needs, but still may allow resources optimization in case specific merchant does not need UI configuration propagated from Magento.
+   1. Even in case of multi-tenant deployment, it is easier to eliminate configuration feature (and free resources needed for it) if it's represented by a separate service.
+
+Cons: 
+
+1. Additional request to configuration service instead of in-process request to the storage.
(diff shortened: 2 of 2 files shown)


design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. Pre-calculate final data on back office side and provide final result to the Storefront during synchronization.
+   1. Pros: simpler implementation of Storefront due to eliminated necessity for Storefront to keep knowledge about additional configuration, including data calculation algorithms.
+   2. Cons: massive (up to full) reindexation necessary in case the configuration is changed.
+   
+To choose the right approach in a specific case, consider the following:
+
+1. How frequently the configuration is expected to change?
+   1. Frequent configuration changes leading every time to massive reindexation may be unacceptable.
+   2. Configuration changes expected a few times in the store lifespan may not be worth additional complexity on the Storefront side and full reindexation may be better in this case.
```
mslabko, line 38: Lets' consider example with base media URL. Looks like this is not often operation, but it assumes that image will not be valid on Storefront until full indexation. To solve it we need to create redirects, but who will be responsible for that? 

I propose to add one more criterion:
If using configuration on Storefront service allows providing different results with the simple operation (using 1 of 2 fields or concatenation) then it's better to calculate this value on SF side

buskamuza: All simple things turn into complex ones with time. And, we all know, that with Magento scopes simple becomes complex due to fallbacks. But I agree the criterion about working store with both old and new data should be made more clear.

buskamuza: Media URLs will become even more complex in case an external DAM comes into play. If we decide to support both Magento and external DAM (which have different URLs), we would need to calculate which product and which image is served from where. It's not necessarily how we will implement it (we may decide to not support mixed mode), but most likely such mode will be necessary anyways (at least for transition period). It's mach simpler to offload this work from SF to backoffice and agree that Catalog SF doesn't need to know anything about DAM systems and it just ingests and then serves image URLs.

design-documents/storefront/Readme.md
```diff
@@ -0,0 +1,15 @@
+# Storefront
```
melnikovi, line 1: Should put design documents related to storefront application in the storefront repo? We already have some design documents there. If not, it would make sense to move all design documents to this repo.

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,65 @@
...
+
+Examples: admin ACLs that allow or deny parts of functionality for the admin user, reindex on update or by schedule.
+
+This type of configuration has nothing to do with Storefront and should not be propagated to storefront, as well as should not be exposed via export API. 
+
+### 2. Configuration that impacts data provided by Storefront
+
+Example: Base Media URL is used for calculation of media image URLs returned by Storefront API.
+
+This type of configuration should be propagated to Storefront in one of the following ways:
```
melnikovi, line 28: Only one way is described below.

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,65 @@
...
+1. How frequently the configuration is expected to change?
+   1. Frequent configuration changes leading every time to massive reindexation may be unacceptable.
+   2. Configuration changes expected a few times in the store lifespan may not be worth additional complexity on the Storefront side and full reindexation may be better in this case.
+2. Is it acceptable to have significant delay in data propagation after the configuration change?
+   1. Changing Base URL may be not a big issue, especially if a redirect can be setup. So it may be acceptable to have URLs to be fully updated in a few hours.
+   2. Changes in prices, on the other side, may not stand long delays.
+
+### 3. Configuration that impacts UI representation of Storefront data
+
+Example: include or exclude taxes in displayed prices.
```
melnikovi, line 45: My understanding was that we don't want to do price calculation in the catalog storefront service. Maybe message broker should be responsible for price calculation for price storefront service?

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,65 @@
...
+
+This type of configuration should be propagated to Storefront in one of the following ways:
+
+1. Pre-calculate final data on back office side and provide final result to the Storefront during synchronization.
+   1. Pros: simpler implementation of Storefront due to eliminated necessity for Storefront to keep knowledge about additional configuration, including data calculation algorithms.
+   2. Cons: massive (up to full) reindexation necessary in case the configuration is changed.
+   
+To choose the right approach in a specific case, consider the following:
+
+1. How frequently the configuration is expected to change?
```
melnikovi, line 36: Would be good to have more examples of configuration we for catalog storefront service.

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,65 @@
...
+   2. Configuration changes expected a few times in the store lifespan may not be worth additional complexity on the Storefront side and full reindexation may be better in this case.
+2. Is it acceptable to have significant delay in data propagation after the configuration change?
+   1. Changing Base URL may be not a big issue, especially if a redirect can be setup. So it may be acceptable to have URLs to be fully updated in a few hours.
+   2. Changes in prices, on the other side, may not stand long delays.
+
+### 3. Configuration that impacts UI representation of Storefront data
+
+Example: include or exclude taxes in displayed prices. 
+
+This kind of configuration has nothing to do with Storefront data itself, and so it is suggested to have a separate service that provides this configuration.
```
melnikovi, line 47: What are the benefits of having specific configuration service and not have configuration as part of main service? From what I have seen, usually one service responsible for both. We can also have generic configuration service. We can add these questions to the document to investigate in the future, unless we already set on the approach (in that case I'm curios what are the benefits).

design-documents/storefront/Readme.md
buskamuza: I thought this repo is source for all design docs. I added a link to this PR in the SF wiki sidebar, so people can find it. When the document is approved and merged, I'll update the link.

design-documents/storefront/configuraiton-propagation.md
buskamuza: It's not calculation, it's setting for presentation layer: do we show price including tax, excluding tax or both? I assume either both prices should be calculated by the back office (or whoever is responsible for it) and then client app just decides which price to use based on this config, or price calculation is synced with UI settings. I think 2nd option is too risky to follow, as for example a store may have multiple client apps (PC web site, mobile app, in-store app) and they may need to display prices differently. So I'd assume that price API should provide both incl and excl prices and prices UI config should say which of those should be displayed.

design-documents/storefront/configuraiton-propagation.md
buskamuza: Should be part of https://github.com/magento/catalog-storefront/issues/155 . This document describes general principles for making decisions about configuration.

design-documents/storefront/configuraiton-propagation.md
buskamuza: Pros and Cons are described in the bottom of the document. I'll clarify that those compare separate config service vs config being part of the domain service.

design-documents/storefront/configuraiton-propagation.md
buskamuza: Good point. Updated :)

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,124 @@
...
+Example: Base Media URL is used for calculation of media image URLs returned by Storefront API.
+
+This type of configuration should be propagated to Storefront in one of the following ways:
+
+1. Pre-calculate final data on back office side and provide final result to the Storefront during synchronization.
+   1. Pros: simpler implementation of Storefront due to eliminated necessity for Storefront to keep knowledge about additional configuration, including data calculation algorithms.
+   2. Cons: massive (up to full) reindexation necessary in case the configuration is changed.
+2. Move/duplicate calculation logic in the Storefront service based on original data is synced from the back office.
+    1. Pros and cons are opposite to option 1. So this option is valuable in case of expected frequent change of configuration that impacts a lot of data.
+
```
buskamuza, line 35: Add note about sync data from 3rd-party systems.

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,124 @@
...
+To choose the right approach in a specific case, consider the following:
```
buskamuza, line 36: Add: consequences are described for the decision made (e.g., sync delay, increased maintenance complexity).

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,124 @@
...
+
```
buskamuza, line 37: Massive reindex should happen only for data that is really changed.

buskamuza:
In stock/out of stock.
@kokoc : keep out of stock products in Catalog SF due to big effort of recalculating product data when it gets back in stock (prices, etc).

buskamuza:
@kokoc : put config (such as in/out of stock) to the deployment phase, to env vars.

buskamuza:
Option 3:

1. UI Config is responsibility of the client (PWA, etc)
2. UI Config is provided by Magento Back Office API (GraphQL) proxied by GraphQL
3. Current option 3.3.

Prioritized in the provided order. Agree with PWA, AEM teams what will be satisfactory.

design-documents/storefront/configuraiton-propagation.md
```diff
@@ -0,0 +1,149 @@
...
+3. Do 3rd-party systems provide similar configuration?
+   1. If 3rd-party systems don't have equivalent configuration, how will it be populated in the Storefront service? It might be better to avoid Magento-specific concepts to simplify integrations, and instead provide indexed data to the Storefront service.
+   1. If a configuration option is pretty common among 3rd-party systems, it may make sense to reflect it in the Storefront service.
+
+Expected consequences are described in the decision document for each case.
+For example, time for configuration propagation in case reindexation is chosen, logic duplication/complexity in case configuration is propagated to Storefront application, performance impact for Storefront read API or for synchronization.
+
+### 3. Configuration that impacts UI representation of Storefront data
+
+Example: include or exclude taxes in displayed prices.
```
buskamuza, line 53: After more discussions: taxes incl/excl is not a good example of UI Config, as it's based on the store (which is based on the country it is run in). So It should be part of service data itself - whether taxes are included or excluded.

Issue #400, opened 2020-07-15
Proposal (GraphQL): Reconsider direction to favor named arguments over input objects


DrewML:
We currently have the following guidance documented for new handling of input data to an operation:

https://github.com/magento/architecture/blob/master/design-documents/graph-ql/extensibility.md

It states that the solution for extensibility is:

> Wrappers for output and merger for arguments

There's a problem with argument lists worth considering.

## Example
Imagine this is a mutation in our schema:

```graphql
type Mutation {
   createFoo(
      arg1: String = "default"
      arg2: Int!
      arg3: String
      arg4: SomeObjectType
   ): CreateFooOutput
}
```

To use this, `pwa-studio` will likely parameterize it:

```graphql
mutation CreateFoo($arg1: String = "default", $arg2: Int!, $arg3: String, arg4: SomeObjectType) {
   createFoo(
      arg1: $arg1
      arg2: $arg2
      arg3: $arg3
      arg4: $arg4
   ) {
      field1
      field2
   }
}
```

There are a few potential problems I'm aware of in this example:

1. `CreateFoo` now has its own default value for `arg1`. There is no way (that I'm aware of) to define this operation where the default from `arg1` will be automatically carried to `$arg1`. @sirugh hit this issue in `pwa-studio` recently.
2. If an optional arg is added to `Mutation.createFoo`, and a user of `pwa-studio` wants to use it, they have to change both UI Components and the query itself (rather than just the UI component).
3. It's very, very verbose. See [this example in pwa-studio](https://github.com/magento/pwa-studio/blob/8c187738be36df7572261d21199f213b7e8cb50c/packages/venia-ui/lib/queries/productSearch.graphql).

## What is the alternative

The alternative is leaning on Input Object types to act as a bag of named arguments. 

### Re-Written Example

```graphql
type Mutation {
   createFoo(input: CreateFooInput): CreateFooOutput
}

input CreateFooInput {
   arg1: String = "default"
   arg2: Int!
   arg3: String
   arg4: SomeObjectType
}
```

```graphql
mutation CreateFoo($input: CreateFooOutput) {
   createFoo(input: $input) {
      field1
      field2
   }
}
```

PR #404, opened 2020-07-17
Propose updates to GraphQL extensibility doc

DrewML:
[Rendered Document](/magento/architecture/blob/updates-to-extensibility-doc/design-documents/graph-ql/extensibility.md)

I noticed the extensibility document was missing some content while I was updating it with proposed solution from #400. This PR contains changes to address #400 and some additional guidelines we (mostly) already follow.

Feedback appreciated :)

Reviewers list populated with folks that commonly contribute to our schema.

### DIFF
--- a/design-documents/graph-ql/extensibility.md
+++ b/design-documents/graph-ql/extensibility.md
@@ -1,46 +1,127 @@
-**Required Extension points**
+**Extensibility**
 
-It should be possible to:
-- Extend or modify the list of arguments from 3rd party extension
-- Add extra data to the output of the mutation in a backward compatible way
-- Evolve arguments list and output type of our API in the future
+## GraphQL Vocabulary
+- _Root Query_ - [Spec](http://spec.graphql.org/June2018/#sec-Query) | [Docs](https://graphql.org/learn/execution/#root-fields-resolvers)
+- _Mutation_ - [Spec](http://spec.graphql.org/June2018/#sec-Mutation) | [Docs](https://graphql.org/learn/queries/#mutations)
+- _Object Type_ - [Spec](http://spec.graphql.org/June2018/#sec-Objects) | [Docs](https://graphql.org/learn/schema/#object-types-and-fields)
+- _Arguments_ - [Spec](http://spec.graphql.org/June2018/#sec-Field-Arguments) | [Docs](https://graphql.org/learn/schema/#arguments)
+- _Input Object_ - [Spec](http://spec.graphql.org/June2018/#sec-Input-Object-Values) | [Docs](https://graphql.org/learn/schema/#input-types)
+
+## Magento GraphQL Vocabulary
+
+- _Output Object_: An _Object Type_ used as the response type for a mutation (primarily to support extensibility)
 
 
-**Solution: Wrappers for output and merger for arguments**
+## Extensibility Requirements
 
-Wrappers for output type along with merging capabilities for arguments can solve extensibility and deprecation issues.
+It should be possible to make the following changes to a schema, _without_ introducing breaking changes:
 
-```$graphqls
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/extensibility.md
```diff
@@ -1,46 +1,127 @@
...
+## Extensibility Requirements
+
+It should be possible to make the following changes to a schema, _without_ introducing breaking changes:
+
+- Add a new root query (backwards-compatible)
+- Add a new mutation (backwards-compatible)
+- Add optional arguments to a query/mutation (some backwards-compat risks)
+- Add extra data to the output of a mutation (backwards-compatible)
+
+The following patterns should be followed to ensure our schema remains extensible, with _minimal_ (ideally no) breaking changes._
```
paliarush, line 24:
```suggestion
The following patterns should be followed to ensure our schema remains extensible, with _minimal_ (ideally no) breaking changes.
```
```suggestion
The following patterns should be followed to ensure our schema remains extensible, with _minimal_ (ideally no) breaking changes._
```

design-documents/graph-ql/extensibility.md
```diff
@@ -1,46 +1,127 @@
...
+### Multiple Arguments
 
-```$graphqls
-type Mutation {
-    generateCustomerToken(
-            email: String!,
-            password: String!,
-            multi_factor_auth_token: String!
-    ): GenerateCustomerTokenOutput!
+If a query or mutation accepts (or will likely accept) > 1 argument, an Input Object should be used instead, and given the argument name `input`.
```
paliarush, line 54: I would say we must always use `input` because we never know how Magento is/will be extended.

DrewML: I think it's very likely you're right here. I have a tendency to try and leave a little wiggle room 😄

Do you want to make any exception for the case of looking something up by ID? I think in the case of a field like `Query.productByID` (as an example) it might be nice to _just_ take the `ID`, and if folks want to add arguments to a unary field, they just have to introduce a new query instead.

If you think this isn's worth it, happy to go forward with your suggested change.

paliarush: Inability to extend is the reason why I don't like `byId` methods. I would use `input` everywhere. Simper rule - less bugs.

DrewML: > Inability to extend is the reason why I don't like byId methods

Does every query need to be infinitely extensible? Can you describe a scenario where a `Query.fooByID` field would need to have its arguments extended, and where that would make more sense than adding a new query?

I think, for single entity lookups (not queries for collections) there aren't many scenarios where it would make sense to take more than an `ID` as input. Most other arguments you'd add to that method would likely change the behavior as well (breaking change).

Please correct me if I'm wrong, but it seems like a lot of the reasons to add more arguments to a field might be a sign that, for simple cases like this, the field is trying to do too much.	

There's also the question of naming - anytime we use a field like `Query.product` instead of `Query.productByID`, we're using a more generic name that's likely to be overloaded later for either uses, leading to things like deprecations and `_v2` fields. Explicit names and granular fields in this case (imo) kind of force the query to be a bit more focused.

> Simper rule - less bugs.

Isn't it the opposite in this case? Adding an argument to an existing field adds branching to a resolver, and a developer needs to take great care to make sure that logic does not impact any existing queries.

paliarush: I don't have strong opinion against `byId`, but would really like to have `input` mandatory.

DrewML: We can do that 👍  Will make changes

DrewML:
@melnikovi @nrkapoor can y'all give this a review when you find time?

PR #405, opened 2020-07-22
[Storefront] Magento price books

kokoc:
## Problem

The final price of the product in Magento monolith depends on multiple variables such as current customer group, current 
website, qty of items in the shopping cart and current date/time. 
Magento monolith calculates all possible permutations of prices in advance and store them in `price index`. These 
calculations are expensive and may not be done in reasonable time for large catalogs.

Few problematic use cases from real merchants:

- Each customer in the system has unique prices for very few products. Per-customer prices are result of a physical contract with pricing included.
The system has 20,000 customer groups, 10,000 products, promotions are excluded from the system. Magento will generate 200,000,000xCOUNT_OF_WEBSITES records to handle all possible combinations for this merchant. For one website it will consume 17,8 GB of space for data and 22.3 GB for index(40 GB total). Due to external promotion system, prices are synchronized periodically. The synchronization process consumes a lot of resources and time and eventually space.

- Customer groups are used for many things, but they are also a multiplier for prices
The system has 56 stores, 58 customer groups, 29,000 products, most customer groups are not global and used on one website only. Existing `price index` contains 25,000,000 records. The reindex process takes more than 7 hours. Customer groups are also used for promotions, cms content, product availability, B2B company, tax status and of course pricing. The real count of the prices is 26 times smaller. Potentially, reindex process may take 16 minutes.

The impact form cases describe above will be doubled(or even tripled) if we introduce a new storefront service with existing index structure inside. Thus, we need some other way to work with prices in storefront.

## Solution

https://github.com/kokoc/architecture/blob/pricing/design-documents/storefront/pricing.md

## Requested Reviewers

@akaplya @kandy @DrewML


design-documents/storefront/pricing.md
```diff
@@ -0,0 +1,311 @@
...
+
+The `price book`(price list) is a new entity that holds a list of product's prices for a sub-set of catalog. 
+The `price book` is a non-scoped entity, but it may hold information about linked customers, websites, etc. 
+Instead of direct lookup in price index by customer_group, website and product, we will detect the customer's price book first, then we will extract two price books from the index: default and current `price book`. 
+The resulting product price will be the value from current price book if it's exists, otherwise the product price will be extracted from default price book:
+
+![Price books diagram](pricing/pricebooks.png)
+
+Guidelines:
+* There is no need to resolve price book on each HTTP call. Resolved price book could be stored in JWT during "login" step
```
DrewML, line 50: Can you explain this more - what specific data is going into a JWT?

kandy:
Can you describe how to indexation will work in monolith?

nrkapoor:
@kokoc Cart rules may also change the price of a product. What's your vision on cart rules? Are we keeping is separate just like today? We have a use case where for a given product we want to get all applicable promotions that impact price.

kokoc:
@nrkapoor 

Yes, I think we have to keep them separate. This proposal is focused on products' base prices. Cart price rules work with shopping cart items and on a different step of user's flow. 

Anyway, the price in price book is not a single float number. I plan to have `original` price and `final` price there. This way we can distinguish base prices from discounts. We may introduce an additional field that contains comment how the discount was calculated.

design-documents/storefront/pricing.md
kokoc: Just in case, here is a doc about jwt - https://github.com/magento/architecture/blob/master/design-documents/service-isolation/detailed-architecture.md

I didn't want to cover this in the proposal because it could be potentially unified with other data in the customer context. However, I see the following options:
* Put price book id (and probably default price book id) in the customer context. Pass this context to all services with each request.
* Store price book ids  in customer context, but extract price book ids on GraphQL side and pass them as arguments to pricing/catalog/search services
* Pass customer_id only to the services. Each service may extract price book ids using separate network call to CPQ service.

I don't like option #3 because of relatively slow performance. #1 and #2 are almost identical to me, but option #1 seems to be more flexible and reusable for other scenarios. However, #1 moves more data with each client->graphl and graphql->services requests.

kokoc:
@kandy added a few more words about indexing. This proposal supposes price calculations in runtime. The indexing process just converts websites-customer groups to price books.

Stepa4man:
@kokoc one more bottleneck here is the full page cache storage size.  20,000 will cause 20,000 different versions of the same page.

paras89:
This is a very specific use case that should need custom development or architecture. To meet such a use case we shouldn't have to change the way Magento works - this will add more complexity to the system.

This case should be handled by custom development.

paras89:
In b2b the pricing contracts reside on the ERP. The erp should expose Apis for the Magento catalog frontend to consume in real time

kokoc:
@paras89 @Stepa4man we plan to release this as a part of the headless solution that will be connected to standalone graphql server. New services described in this PR and any other service in the scope of service isolation work acts as a storefront cache. It may also include some business logic required to place the order. 

Regarding FPC, I think we need to get rid of outer cache for graphql APIs because of few reasons:
- Storefront itself acts as a cache for some scenarios (catalog browsing, pdp)
- Standalone graphql server has underlying caching capabilities - https://www.apollographql.com/docs/apollo-server/performance/caching/

mbrinton01:
We cannot assume that the customer will always want to pay the lowest (minimum) price. In the case of contract B2B pricing, customers may be required to pay the negotiated contract price, even if it is higher than the base price or there is a discount enabled.

kokoc:
> We cannot assume that the customer will always want to pay the lowest (minimum) price. In the case of contract B2B pricing, customers may be required to pay the negotiated contract price, even if it is higher than the base price or there is a discount enabled.

This case is supported by new structure

design-documents/storefront/pricing.md
```diff
@@ -0,0 +1,355 @@
...
+
+This happens because parent product and variation are separate products which could be assigned to different access lists 
+and price books. In order to mitigate this issue products should be isolated, so product options fully define complex products.
+
+The case from example above could be handled by two independent configurable products with different sets of variations 
+for different access lists.
+
+Details will be provided in the separate proposal.
+
+### Prices fallback
```
akaplya, line 249: How many fallbacks will be allowed to set up for the default price book? Is it possible to set up a fallback for the fallback price book?

design-documents/storefront/pricing.md
```diff
@@ -0,0 +1,355 @@
...
+    // already occupied by another price book.
+    repeated string websites = 4;
+}
+
+message PriceBookDeleteInput {
+    string id = 1;
+}
+
+message AssignProductsInput {
+    message ProductPriceInput {
```
akaplya, line 162: How the  MSRP and Tier prices will be supported?

akaplya (changes requested):

design-documents/storefront/pricing.md
```diff
@@ -0,0 +1,355 @@
...
+        ? $product->getSpecialPrice
+        : 0;
+
+    $fixedPrice = min($product->getBasePrice(), $specialPrice);
+
+    $groupPrice = $product->getGroupPrice($customerGroup);
+    if ($groupPrice) {
+        $fixedPrice = ($groupPrice->getType() == 'percentage')
+            ? $product->getBasePrice() - $product->getBasePrice() / 100 * $groupPrice->getValue()
+            : $groupPrice->getValue;
```
mslabko, line 320: we do we ignore special price? I think correct behavior should be:

fixed price = customer_group_price
  ? min (special, customer_group_price)
  : min (special, base_price)

design-documents/storefront/pricing.md
```diff
@@ -0,0 +1,355 @@
...
+Guidelines:
+* There is no need to resolve price book on each HTTP call. Resolved price book could be stored in JWT during "login" step 
+and reused for consequent requests.
+* In order to minimize "query-time" work, customer should have exactly one resolved price book.
+
+
+### Default price book
+
+A `default price book` is a predefined system price book that contains `base prices` for __ALL__ products in the system. User-defined `price books` may contain only sub-set of products. 
+Default `price book` should be used as fallback storage if the price for a specific product doesn't exist in other resolved pricebook.
```
mslabko, line 58: How to determine price for the Default Price book if Product has different prices in several Websites, but product is not assigned to the website marked as "default"?

kandy: "Default" here is global scope, not some "default website" scope

paliarush:
@kokoc Closing due to inactivity. If possible, please finalize it, reopen and merge.

mslabko:
I believe we can merge this PR as is - in general, we are going to use this design for Price Book implementation in scope of the Storefront Application project

PR #407, opened 2020-07-24
Remove global uniqueness requirement from graphql single add to cart …

DrewML:
Follow-up to #396, based on discussion in meeting today and concern mentioned by @akaplya.


nrkapoor:
@prabhuram93 How does this impact the current PR?

PR #411, opened 2020-07-30
[Draft] Catalog storefront, product options and variants management

akaplya:
## Solution

This document proposes to review options and variants management approach to :
* achieve more consistent behaviors across the options.
* reduces data duplication caused by the existing variants matrix and associated products.
* resolve dependency between options and their prices

### DIFF
--- a/design-documents/storefront/catalog/product-options-and-variants.md
+++ b/design-documents/storefront/catalog/product-options-and-variants.md
@@ -1,12 +1,30 @@
 
-## Domain Overview
+## Overview
 
-Magento allows customizing a product before adding the product to the cart.
-Customization happens through the customizable options defined by the merchant.
-Options can represent customization, other products, digital goods so on.
+Product options are a powerful instrument that allows a shopper to customize and (or) personalize a product before adding the product to the cart. 
 
-All Magento product types represented through the options.
+Introduction of the storefront API brings us a unique opportunity to revisit the options management, address the existing limitation, and address known issues:
 
+* Reduce redundancy caused by variants matrix creation. Magento creates variant and simple product for each intersection of options and option values selected to represent the configurable product. Not all the cases in real require a product creation, so extracting the variant as an entity could reduce system load caused by the number of products. 
+
+* Allow to set up a dependency between the different options of a simple product based on previously selected option values. 
+Currently, simple and bundle products do not support dependencies between option values.
+An example, bundle product that represents a computer in your store may need to correlate the list of the available motherboards with the socket of the selected processor.
+
+* Support grouped B2B prices at the options level. There is no way to specify a special price for a customer group for simple, downloadable,  and some of the bundled options of the product.
+
+* Although all the options which represent product variants are pretty similar and most of them have similar properties, the options not generalized in Magento. As a result, it makes the option management more complicated, and for some cases such as option displaying, causes frontend rendering logic duplication. The ultimate goal after the options API revision is to build a generalized view that should to significantly reduce the complexity of the options domain for the storefront application.
+
+### Taxonomy of the options
+
+Due to their origin and structure options segregates on two main subtypes.
+
+***Product Option Variants*** & ***Shopper Input Options***.
+
+The first subtype - ***Product Option Variants***, is much often used and represents product configuration, which was predefined by a merchant.  And so, product customization could be described by the selection of these predefined options, which, in their turn, creates product variants, where each variant represents the selection of one or many option values.
+This document will be focused on the first subtype, Product Option Variants, it just briefly covers the reasons for the options separation, due to the intention do not overload this document with information about the domain.
+
+*Note: Most of the logic that we use to associate with  Magento product types (configurable, bundle, downloadable, etc) in fact is the logic of the different options.*
 
 ### Definitions
 
--- a/design-documents/storefront/catalog/product-options-and-variants.md
+++ b/design-documents/storefront/catalog/product-options-and-variants.md
@@ -356,3 +356,118 @@ This proposal continues the idea of product options unification and aligned with
 * [Single mutation for adding products to cart](https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md)
 * [Configurable options selection](https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/catalog/configurable-options-selection.md)
 * [Gift Registry](https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/gift-registry.md)
+
+
+## Question & Answers 
+
+### How to express two configurable products that share the same attribute set?
+
+*Comment: Although this case natively supported by Magento,
+it could be less common than we used to think since the two regular t-shorts
+from Walmart will have different sets of values for sizes and colors depends
+on the manufacturer. proof https://www.walmart.com/search/?query=tshirt*
+
+
+#### Product #1 
+```json
+{
+  "id": "a82deb7a-dee7-48e7-ab34-75026e576fab",
+  "name": "Fruit of the Loom Men's Short Sleeve Assorted Crew T-Shirt",
+  "options": {
+    "color": {
+      "label": "Color",
+      "values": {
+        "red": {
+          "label": "Red"
+        },
+        "green": {
+          "label": "Green"
+        }
+      }
+    },
+    "size" : {
+      "label": "Size",
+      "values": {
+        "m": {
+          "label": "M"
+        },
+        "l": {
+          "label": "L"
+        }
+      }
+    }
+  }
+}
+```
+
+#### Product 1 variants
+```json
+[
+  {
+    "variantId": "500d0366-777f-4a45-92b6-8e5197ce9992",
+    "optionValueIds": [
+      "a82deb7a-dee7-48e7-ab34-75026e576fab:color/red", "a82deb7a-dee7-48e7-ab34-75026e576fab:size/l"
+    ],
+    "productId": "8b6be8b0-2e21-4763-806c-f383a8591d21"
+  },
+  {
+    "variantId": "b843e139-aa04-44d0-a9a7-b439a17ce941",
+    "optionValueIds": [
+      "a82deb7a-dee7-48e7-ab34-75026e576fab:color/green", "a82deb7a-dee7-48e7-ab34-75026e576fab:size/m"
+    ],
+    "productId": "96a5a8ed-7cfe-4626-be29-f60fe0bf7b33"
+  }
+]
+```
+
+#### Product 2
+```json
+{
+  "id": "747d8c9b-e5fc-437a-8263-271dd8352976",
+  "name": "George Men's Assorted Crew T-Shirt",
+  "options": {
+    "color": {
+      "label": "Color",
+      "values": {
+        "red": {
+          "label": "Red"
+        },
+        "green": {
+          "label": "Green"
+        }
+      }
+    },
+    "size" : {
+      "label": "Size",
+      "values": {
+        "m": {
+          "label": "M"
+        },
+        "l": {
+          "label": "L"
+        }
+      }
+    }
+  }
+}
+```
+
+#### Product 2 variants
+```json
+[
+  {
+    "variantId": "edbb59fb-f303-4970-9f03-889e71374a90",
+    "optionValueIds": [
+      "747d8c9b-e5fc-437a-8263-271dd8352976:color/green", "747d8c9b-e5fc-437a-8263-271dd8352976:size/l"
+    ],
+    "productId": "7f4db047-604f-41ce-8998-a015f578e023"
+  },
+  {
+    "variantId": "2c865d16-1723-49a3-8fd1-9b46613b5c12",
+    "optionValueIds": [
+      "747d8c9b-e5fc-437a-8263-271dd8352976:color/red", "747d8c9b-e5fc-437a-8263-271dd8352976:size/m"
+    ],
+    "productId": "8eab9d67-791a-4c34-bc30-8bd034856ee2"
+  }
+]
+```
--- a/design-documents/storefront/catalog/product-options-and-variants.md
+++ b/design-documents/storefront/catalog/product-options-and-variants.md
@@ -122,7 +122,7 @@ message ProductOption {
 }
 
 message ProductVariant {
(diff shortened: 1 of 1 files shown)


design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,363 @@
...
+Examples: bundle product, customizable product, downloadable products.
+![](https://app.lucidchart.com/publicSegments/view/fea3f950-6f2f-4e46-90e4-f8ee4b6877f7/image.png)
+Both of approaches could be used together.
+
+Example: configurable product with customizable option.
+
+
+## Competitors analysis
+During my research, I analyzed documentation on how options are covered by the competitors.
+commercetools and Shopify attack this domain with product variants that represent all the possible options.
```
tariqjawed83, line 43: Shopify yes, but not commercetools since they don't force you to create any variants unless you create the combination by yourself based on the attributes defined; they also don't have options as Magento and Shopify has. They are using the concept of Product Type/Attributes with constraints to achieve the same goal.

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,363 @@
...
+Despite the domain simplification by reducing the number of entities both systems constrained with the size of a product which makes it impossible to compose complex business cases that we frequently may face in the Magento ecosystem.
```
tariqjawed83, line 44: on commercetools a soft limitation of 100 is based on performance goals, they believe having product with more than 100 variants is not a good catalog design to begin with, same as having too many nested categories. In order to have good performance outcomes, they have applied this soft limit, which can be reviewed based on the project needs.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: the limitation that I mentioned is written on their "Product Modeling" page, products contain all its variation, that's why a content manager has to keep in mind the size of the output JSON payload which is inconvenient and I do not want to repeat this in Magento.

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,363 @@
...
+in some way, I would compare their options features with Magento 2 x functionalities.
+Bigcommerce segregates options from variants and allows to associate intersection of options with a new product, and as a result, I believe, struggles from the same disease as Magento does - redundancy in the variant data.
+
+[Bigcommerce: Product Options](https://support.bigcommerce.com/s/article/Options-SKUs-Rules)
+
+## Goals
+
+* Eliminate hard dependency on prices from options to make possible use promotions and B2B prices for customizable options, bundle, and downloadable products.
+* Eliminate variants redundancy
+    * Catalog should not have a variant per each option values combination, only if necessary.
```
tariqjawed83, line 60: this is exactly what commercetoool is doing, so I would agree with this approach. Also they are not using Options concept at all, rather Product Type and Attributes (with variant constraints).

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,363 @@
...
+    * Catalog should not create a product per variant if this product never used.
+* Hide the variant matrix from storefront client, use it only in the context of operations.
+* Align a similar option with a single data structure to manage them together.
+
+
+## Modeling Options and Variants Data Objects
+
+The great advantage of Magento 2 - modularity was not a real thing for the product options during Magento the lifetime.
+Since times of Magento 1.x option prices coupled into options, inventory records ignored even options with assigned SKUs.
+The only thing that had to play it together "Bundle" product was designed too complex to treat it as universal solution it had to become.
```
tariqjawed83, line 70: Check attribute references from commercetools for bundling approach. In that approach an attribute can have a reference type, which can be used for bundling. Meaning you can create a Product Type, with reference type attributes of other product type.

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,363 @@
...
+    object_id char(36) not null,
+    weight tinyint not null,
+    primary key (value_id, object_id)
+);
+
+alter table product_variant_matrix
+    add foreign key fk_product_variant_matrix_value_id (value_id) references product_option_values(value_id);
+```
+
+`product_variant_matrix` - Stores the correlation between option values and variants, by using this correlation, we can say which of option combination is real.
```
tariqjawed83, line 168: I would also suggest to see the attribute constraint technique used by commercetools in order to check the validity of variant combination. That maybe a better logical approach to achieve this.

https://wiki.corp.adobe.com/display/EntComm/Product+Catalog+-+Competitors+Landscape#ProductCatalogCompetitorsLandscape-CommerceTools

https://docs.commercetools.com/merchant-center/product-types

design-documents/storefront/catalog/product-options-and-variants.md
tariqjawed83: yes but its not like a product matrix, rather whichever variant you have explicitly created under the product, its not all combinations; only created ones by customer are only stored under a JSON document.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: Segregation of options and variants could resolve this limitation, in my opinion, all the limitations that we have at the moment as well as competitors, as for me, caused by exercise dependency on variants on PDP. If the variants matrix goes under the hood and will be used for filtering purposes mostly -  the most known performance limitations we know so far will be resolved.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: product types are entities more similar to Magento attribute sets, and as for me, this is something that we do have right now in M2 but do want to have in the storefront, because of a fact that options can not belong to the template in the most cases. As usual, if you have two products with the same list of options and options values that means that you have a duplicate of a product in your catalog.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: that's exactly what we have with configurable products, it has a perfect sense for admin functionality to enforce data consistency but in fact redundant for the storefront.

akaplya: what I am trying to say, that catalog is a projection od data that's why we probably do not want to integrate back-office constants. especially taking to account mutable sources of products that could be connected.

tariqjawed83: I believe the concept is applied on product catalog modeling, whether storefront or admin; the same concept can be applied.

design-documents/storefront/catalog/product-options-and-variants.md
tariqjawed83: I think you are thinking in terms of Magento Options and Attributes, the way product type and attributes are used in commercetools is quite different, specially how attributes are helping to pave the way for the formation of variants; so the two concepts are merged together beautifully.

akaplya: that's exactly what https://docs.magento.com/user-guide/catalog/product-create-configurable.html does

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Enhanced option values
+
+To reach a better user experience,
+the option could be extended with an image that represents it 
+or info URL that provides additional information about option value.
+
+[Magento: Swatches](https://docs.magento.com/user-guide/catalog/swatches.html)
+
+[Magento: Downloadable products, options with samples](https://docs.magento.com/user-guide/catalog/product-create-downloadable.html)
```
mslabko, line 83: How can we cover the following attributes required for Downloadable product?
- sample URL, Link URL ("imageUrl" can cover only 1 url or we going to create 2 optionValues for "Downloadable Link")
- Shareable 
- Max. Downloads

akaplya: shareable and max download represent admin functionality and should not be accessible at the storefront. 
Samples are not options! you can not select a sample at the checkout, the closest by the meaning for samples are images and videos, so I expect that samples will be managed correspondingly.

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,355 @@
...
+    string imageUrl = 5;
+    string infoUrl = 6;
+}
+
+message ProductOption {
+    string id = 1;
+    string label = 2;
+    string sortOrder = 3;
+    string isRequired = 4;
+    string renderType = 6;
```
melnikovi, line 118: Could you explain one more time how would we distinguish between custom and configurable options?

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,355 @@
...
+    repeated ProductOptionValue values = 5;
+}
+
+message ProductVariant {
+    repeated string optionValueId = 1;
```
kokoc, line 123: Let's try to think about some pizza examples:

We are Dominos. No stock as we can produce all pizza in the world, so variation represents a meaningful price for the pizza. There might be few options: size, style, topping and cheese levels which affect prices. Also, there are other options such as `cut`(pie cut, square cut, no cut) which don't affect the price. Also, there is a constraint on the `styles` based on the `size` of the pizza, these constraints do not affect prices. E.g. `small` -> `hand tossed`, `gluten free` only; `medium` - `hand tossed`, `crunchy thin crust`, `handmade pan`; `large` -> `brooklyn style`, `hand tossed`, `crunchy thin crust`;

- How to define/return optional options like `cut`?
- How to define that `brooklyn style` is only for `large` pizzas? Remember? They have the same price and no stock so it's reasonable to use one variation.
- Sometimes there will be a need to defined one variation in multiple weights. E.g. the same price for no `sauce` option, if you select sauce option you need to select an additional `sauce type` option:
Screenshot 2020-08-04 at 10 43 02
Screenshot 2020-08-04 at 10 42 50

- The `additional cheese` option has the same price for all large pizzas, but large pizzas themself may have different prices. How to define `additional cheese` variation in a way that will allow to reuse it in different pizzas?
- Next level: `Additional cheese cannot be removed from Dominos Handmade Pan pizzas due to product quality`. So,  the common non-required option is actually required for one selected pizza style. How to defined such things in addition to the previous point?
Screenshot 2020-08-04 at 10 51 48


Other notes:
- Having a weight attribute in this interface looks like a leaky abstraction. Is it possible to hide it, so api interfaces will be easier to work with?
- The current interface requires to describe all possible combinations. However, we know that some

design-documents/storefront/catalog/product-options-and-variants.md
```diff
@@ -0,0 +1,355 @@
+
+## Document purpose
+
+Product options are a powerful instrument that allows a shopper to customize and (or) personalize a product before adding the product to the cart. 
+
+Introduction of the storefront API brings us a unique opportunity to revisit the options management, address the existing limitation, and address known issues:
+
+* Reduce redundancy caused by variants matrix creation. Magento creates variant and simple product for each intersection of options and option values selected to represent the configurable product. Not all the cases in real require a product creation, so extracting the variant as an entity could reduce system load caused by the number of products.
```
kokoc, line 8: Please add a section that will describe a data migration plan/algorithm for Magento merchants. 
- How to reduce the number of variations taking into account that Magento already has separate products for variations?
- Where everything is going to happen(Message broker or Monolith)?
- The shape on monolith APIs

tariqjawed83:
Hi - based on reviewing shopify, big-commerce, commercetools, magento and salesforce in detail, I have following suggestion for truly headless product schema.

1. Can we just use Product Template with multiple Attributes?
2. Each Attribute will have their Type; for instance boolean, text, number/money, datetime, reference to other objects, list.
3. Constraint will help define the Variant possiblity and the field can have no constraint also; for instance "Unique for each variant", "Same for all variants", "Unique for combination".
4. Eliminate the need for Variant Matrix by using Attribute Constraints only, which helps to validate the new variant.
5. Product Variant will act a concrete product, based on Product Template only, No need for selecting options, make it simple.
6. Product Variants with hold concrete values of same attributes defined in Product Template. These variants will be created by choice only, not pre-populated or created; we will just validate the variant based on attribute constraints defined above.
7. Don't think product options just for variants; think of them as product attributes in general; yes they can have constraints but if no constraint defined, it should be treated like normal field.
7. Eliminate anything related to UI / User Experience from Product Template / Variant Definition, since it should be completely headless. All we need to do is define the attribute types in such a way that they could store any type of UI field; like file reference, text input, boolean, list etc. (as defined above).
8. Pricing and Inventory can be reference of Product Variant itself, since that is a concrete type; but this can be flexible, some Product Variant can derive their price or inventory from Product Template, and some optionally use their own.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: From the storefront standpoint, no how., Because all the product variants could be represented through the one or many option values. So you do not really care how the API searches the variant, and in some way, what exactly this variant represents. The only thing about it cares is how efficiently select the option. Because  Magento does not, and should not distinguish behavior between configurable option selection and a dropdown option for instance.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: This behavior does not support right now by Magento, and the document does not cover this explicitly. Despite this fact, such behavior could be easaly achieved for the future by adding a property at the productOptionValue, which will restrict other productOptionValues to be shown. 

I removed `weight` attribute, agree with you it was a leak of implementation details that should not be exposed.

design-documents/storefront/catalog/product-options-and-variants.md
akaplya: the first iteration does not assume the entities reduction, since at the first round we will have to support Magento cases only, but idea behind that statement is that for future you may have API and implementation that may share the same variant across the different option selection. At this stage, we are defining APIs which do not contradict with the mentioned behavior.

akaplya:
should be merged as is, since the last conversation the proposal was mostly implemented. So it has more sense merge it, and apply changes on top of the document if required.

PR #415, opened 2020-08-05
MC-36347: ID Improvement Plan

DrewML:
[Rendered Document](https://github.com/magento/architecture/blob/MC-36347-design-doc/design-documents/graph-ql/id-improvement-plan.md)

### DIFF
--- a/design-documents/graph-ql/id-improvement-plan.md
+++ b/design-documents/graph-ql/id-improvement-plan.md
@@ -1,0 +1,227 @@
+# ID Improvement Plan
+
+We've recently agreed on some standardization to how object identifiers are represented in Magento GraphQL schemas, which will require some changes to existing code to reach the desired state.
+
+This work _must not introduce any breaking changes_.
+
+## Approved Proposals
+
+- [Propose renaming id_v2 to something more permanent, and change type #396](https://github.com/magento/architecture/pull/396)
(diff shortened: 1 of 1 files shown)


sirugh:
This is going to be very important for the client. By default, ApolloClient overwrites existing data with incoming data if there is no identifier for the data. A good example of this is two queries that fetch some subset of `shipping_addresses`. 

```gql
# query 1
shipping_addresses {
  firstname
  lastname
}

# query 2
shipping_addresses {
  city
  street
  postcode
}
```

The last query to be processed will have its data stored in the cache (although a [custom merge function](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#merging-arrays-of-non-normalized-objects) can be defined)). For example if query 2 resolves after query 1 then even though we've already fetched `firstname` and `lastname`, the first query will require a new network round trip to resolve. Defining a custom merge function is an option but it requires a lot of overhead on the front end. For example, we can require that all queries for `shipping_addresses` fetch some uniquely identifying combination of fields, but that's essentially just the same thing as using those values to hash an `id` on the server.

As one of our core tenants in PWA is offline capability, serving from cache when possible is quite important, and it'd be preferable to have merging handled automatically via `id` rather than having to define custom merge functions for each field that doesn't expose one.

PR #421, opened 2020-08-13
Catalog permissions synchronization with storefront

melnikovi:
## Solution

Proposal describes synchronization of catalog permissions to storefront.


design-documents/catalog-permissions.md
```diff
@@ -0,0 +1,293 @@
...
+		}
+	}
+}
+```
+
+## Using catalogs to represent visibility
+
+Catalog are going to be used to represent visibility of the product. To represent permissions and websites with catalogs we will have to create catalog for website and customer group (for which we have catalog permissions configured) combinations.
+
+More information on catalogs [here](https://wiki.corp.magento.com/pages/viewpage.action?spaceKey=ARCH&title=Organization+Structure%2C+Magento+Instance+Registration+and+Scope+Isolation).
```
mslabko, line 43: Can you put here what exactly we need from "catalog" and what is it (for those who can't open resource)
Can we implement this proposal without implementation mentioned "catalog"?

PR #418, opened 2020-08-07
MC-36365: GraphQL schema for B2B purchase order

paliarush:
**Requirements**
As a developer, I want to create a flexible purchase order details schema that can be added to the graph so a B2B shopper can see purchase order details in their account.

*Note*
Base branch should be changed to `master` after https://github.com/magento/architecture/pull/414 is merged.

**Additional changes**
- Introduced `country` field in `CustomerAddress` type to be able to fetch country name from the purchase order addresses. Deprecated `region_id` and `country_code`. See https://github.com/magento/architecture/pull/418/commits/57f33eac58b74a4bbd1d0e3b744880f8eb3b0598

PR #422, opened 2020-08-17
negotiable-quote-graphql-schema-improvement

RakeshJesadiya:
## Problem

The NegotiableQuote Graphql architectural schema was not aligned with the implementation. Need to change the type to implementation fit.

## Solution

During the Development time of the Negotiable Quote GraphQl, We have faced ENUMERATION TYPE Null Issue for the NegotiableQuoteHistoryStatusChange with old_status because old_status will be return null for the first iteration of history log of the current Quote and this will throw the error. We have added String type instead of ENUM.

Update schema to reflect the implementation of items object and NegotiableQuoteHistoryStatusChange.

## Requested Reviewers
@paliarush 
@DrewML

### DIFF
--- a/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
@@ -172,7 +172,7 @@ enum NegotiableQuoteCommentCreatorType {
 type NegotiableQuote {
     id: ID!
     name: String!
-    items: [NegotiableQuoteItem!]
+    items: [CartItemInterface!]
     # Attachment Support is dependent on headless File Upload design
     # attachments: [AttachmentContent]
     comments: [NegotiableQuoteComment!]
@@ -196,11 +196,6 @@ enum NegotiableQuoteStatus {
     CLOSED
     DECLINED
     EXPIRED
-}
-
-type NegotiableQuoteItem @doc(description: "A line item on a Negotiable Quote") {
-    id: ID! @doc(description: "ID of the line item (not product) in a Negotiable Quote")
-    item: CartItemInterface @doc(description: "Product assigned to line item, with selected configurations")
 }
 
 input NegotiableQuoteFilterInput {
@@ -247,8 +242,8 @@ type NegotiableQuoteHistoryChanges {
 
 # Usage in Luma: https://github.com/magento/magento2b2b/blob/0e791b5f7cd604ee6ee40b7225807c01b7f70cf2/app/code/Magento/NegotiableQuote/view/base/templates/quote/history.phtml#L40-L73
 type NegotiableQuoteHistoryStatusChange {
-    old_status: NegotiableQuoteStatus @doc(description: "Will be null for the first history entry on a Negotiable Quote")
-    new_status: NegotiableQuoteStatus!
+    old_status: String @doc(description: "Will be null for the first history entry on a Negotiable Quote")
+    new_status: String! @doc(description: "Negotiable Quote History New Status.")
 }
 
 # Usage in Luma: https://github.com/magento/magento2b2b/blob/0e791b5f7cd604ee6ee40b7225807c01b7f70cf2/app/code/Magento/NegotiableQuote/view/base/templates/quote/history.phtml#L48
--- a/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
@@ -99,11 +99,7 @@ type DeleteNegotiableQuotesOutput {
     # Implementation Note: We don't make the deleted quotes accessible
     # because "deleted" means they're hidden from the storefront UI. They will
     # still be visible (and labeled as deleted) for a Seller in the admin
-    negotiable_quotes(
-        filter: NegotiableQuoteFilterInput,
-        pageSize: Int = 20,
-        currentPage: Int = 1
-    ): NegotiableQuotesOutput @doc(description: "List of negotiable quotes available to customer")
+    quote: Boolean
 }
 
 input CloseNegotiableQuotesInput {
@@ -112,11 +108,6 @@ input CloseNegotiableQuotesInput {
 
 type CloseNegotiableQuotesOutput {
     closed_quotes: [NegotiableQuote!] @doc(description: "Quotes that were just closed")
-    negotiable_quotes(
-        filter: NegotiableQuoteFilterInput,
-        pageSize: Int = 20,
-        currentPage: Int = 1
-    ): NegotiableQuotesOutput @doc(description: A (optionally filtered) list of Negotiable Quotes viewable by the logged-in customer")
 }
 
 input RemoveNegotiableQuoteItemsInput {
--- a/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
@@ -111,7 +111,7 @@ input CloseNegotiableQuotesInput {
 }
 
 type CloseNegotiableQuotesOutput {
-    closed_quotes: [NegotiableQuote]! @doc(description: "Quotes that were just closed")
+    closed_quotes: [NegotiableQuote!] @doc(description: "Quotes that were just closed")
     negotiable_quotes(
         filter: NegotiableQuoteFilterInput,
         pageSize: Int = 20,
--- a/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/negotiableQuotes.graphqls
@@ -242,8 +242,8 @@ type NegotiableQuoteHistoryChanges {
 
 # Usage in Luma: https://github.com/magento/magento2b2b/blob/0e791b5f7cd604ee6ee40b7225807c01b7f70cf2/app/code/Magento/NegotiableQuote/view/base/templates/quote/history.phtml#L40-L73
 type NegotiableQuoteHistoryStatusChange {
-    old_status: String @doc(description: "Will be null for the first history entry on a Negotiable Quote")
-    new_status: String! @doc(description: "Negotiable Quote History New Status.")
+    old_status: NegotiableQuoteStatus @doc(description: "Will be null for the first history entry on a Negotiable Quote")
+    new_status: NegotiableQuoteStatus! @doc(description: "Negotiable Quote History New Status.")
 }
 
 # Usage in Luma: https://github.com/magento/magento2b2b/blob/0e791b5f7cd604ee6ee40b7225807c01b7f70cf2/app/code/Magento/NegotiableQuote/view/base/templates/quote/history.phtml#L48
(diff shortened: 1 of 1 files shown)


dani97:
@paliarush @DrewML https://spec.graphql.org/June2018/#sec-Enum-Value from Graphql spec we cannot make it null, can we add a ENUM value that represents null

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
```diff
@@ -247,8 +242,8 @@ type NegotiableQuoteHistoryChanges {
 
 # Usage in Luma: https://github.com/magento/magento2b2b/blob/0e791b5f7cd604ee6ee40b7225807c01b7f70cf2/app/code/Magento/NegotiableQuote/view/base/templates/quote/history.phtml#L40-L73
 type NegotiableQuoteHistoryStatusChange {
-    old_status: NegotiableQuoteStatus @doc(description: "Will be null for the first history entry on a Negotiable Quote")
-    new_status: NegotiableQuoteStatus!
+    old_status: String @doc(description: "Will be null for the first history entry on a Negotiable Quote")
```
DrewML, line 245: Can you help me understand why changing this to `String` fixes the issue? `old_status` set to `NegotiableQuoteStatus` can still be `null`

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
```diff
@@ -172,7 +172,7 @@ enum NegotiableQuoteCommentCreatorType {
 type NegotiableQuote {
     id: ID!
     name: String!
-    items: [NegotiableQuoteItem!]
+    items: [CartItemInterface!]
```
DrewML, line 175: Good catch!

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
dani97: Andrew it seems graphql spec doesn't allow null for enum, can we add a extra value for enum instead of string

dani97: https://spec.graphql.org/June2018/#sec-Enum-Value reference

DrewML: > it seems graphql spec doesn't allow null for enum

`enum` is a specific type, but all fields, by default, can be `nullable`. The type assigned to a field shouldn't impact it, unless the type is wrapped in a Non-Null wrapper

DrewML: Are you hitting errors implementing that? If so, please provide the details here - could be a webonyx bug.

dani97: yes, @RakeshJesadiya please share the error thrown.

RakeshJesadiya:
@DrewML I have updated the schema file again. Keep Enum Status type as it, I have resolved the issue for that and modify total object with Money Object is not mandatory.
In the history log most of the time, old and new price value will not set, and its display 0, if we want to keep `Money!` The output looks like,
```
"total": {
              "old_price": {
                "value": 0,
                "currency": "USD"
              },
              "new_price": {
                "value": 0,
                "currency": "USD"
              }
            },

If an old and new price is not set its always display as
 "total": {
              "old_price": null,
              "new_price": null
}

File: design-documents/graph-ql/coverage/negotiableQuotes.graphqls [hunk 3]
```diff
@@ -266,8 +261,8 @@ type NegotiableQuoteHistoryCommentChange {
 
 # Usage in Luma: https://github.com/magento/magento2b2b/blob/0e791b5f7cd604ee6ee40b7225807c01b7f70cf2/app/code/Magento/NegotiableQuote/view/base/templates/quote/history.phtml#L259-L294
 type NegotiableQuoteHistoryTotalChange {
-    old_price: Money!
-    new_price: Money!
+    old_price: Money
```
DrewML, line 264: Looks good 👍

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
DrewML: Looks like @rakeshjesadiya got it figured out https://github.com/magento/architecture/pull/422#issuecomment-675407366

DrewML:
> In the history log most of the time, old and new price value will not set, and its display 0, if we want to keep Money! The output looks like,

Agree with your changes there - let's go with your switch to nullable, and use `null` to present no value (since 0.00 has specific meaning)

RakeshJesadiya:
@DrewML @paliarush I have updated the schema.graphql with the removal of the `addNegotiableQuoteItems` schema.
Buyer Can't able to add items from the LUMA storefront. The only Seller Can Add a new item to the existing quote.
schema is unused for the storefront GraphQl API.

DrewML:
> @DrewML @paliarush I have updated the schema.graphql with the removal of the addNegotiableQuoteItems schema.
Buyer Can't able to add items from the LUMA storefront. The only Seller Can Add a new item to the existing quote.
schema is unused for the storefront GraphQl API.

Thanks @RakeshJesadiya - great catch! I confirmed this with @mbrinton01

DrewML:
> Example, Negotiable quote with id 10 is already closed, if the user passes mutation with quote_ids: [10] response will be null instead of NegotiableQuote object.

I wonder if this is the desirable behavior. I'm curious if it would make more sense to just return the quote with an ID of 10 here. What's the thinking behind returning `null` for that instead?

RakeshJesadiya:
@DrewML Please check the modified schema file. I am not sure how we can filter out the Output in graphql schema. Is it acceptable criteria?

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
```diff
@@ -104,24 +99,15 @@ type DeleteNegotiableQuotesOutput {
...
+    quote: Boolean
 }
 
 input CloseNegotiableQuotesInput {
     quote_ids: [ID!]! @doc(description: "A List of IDs from NegotiableQuote objects")
 }
 
 type CloseNegotiableQuotesOutput {
-    closed_quotes: [NegotiableQuote]! @doc(description: "Quotes that were just closed")
-    negotiable_quotes(
```
DrewML, line 120: Why were these fields removed? `closed_quotes` and `negotiable_quotes`?

RakeshJesadiya: @DrewML I have only removed negotiable_quotes from the output. Keep closed_quotes as it is. I am not sure can we use filtration on output also? The negotiable_quotes field has to add a user input. How can we set filter on output?
'''type CloseNegotiableQuotesOutput {
}'''
is the Output of closeNegotiableQuote.

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
```diff
@@ -104,24 +99,15 @@ type DeleteNegotiableQuotesOutput {
     # Implementation Note: We don't make the deleted quotes accessible
     # because "deleted" means they're hidden from the storefront UI. They will
     # still be visible (and labeled as deleted) for a Seller in the admin
-    negotiable_quotes(
```
melnikovi, line 107: I see that we have inconsistency on how we design output type in similar cases. Some output types [return boolean](https://github.com/magento/architecture/pull/410/files#diff-1f75a7597d5d9a4c23b7b09139bdde27R193) field, others [allows to query](https://github.com/magento/architecture/pull/418/files#diff-929518073d3c60214efac263e7df1d8cR22) items. Looking into how can we align this. Ability to query items seems more convenient, why do you prefer returning a boolean?

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
melnikovi: Here an example (it will not work because user is not logged in after creation, but shows how you can do filtration on the output). Let me know if I misunderstood the question.
```
mutation {
  createCustomer(input: {email: "email@domain.com", firstname: "John", lastname: "Doe", password: "test123$"}) {
    customer {
      email
      orders (filter: {number: {eq: "1"}}) {
        items {
          id
        }
      }
    }
  }
}
```

DrewML:
Reverted the approval review state for now while we're still committing changes and discussing

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
melnikovi: Here are the pros and cons that I see.

```
    negotiable_quotes(
        filter: NegotiableQuoteFilterInput,	
        pageSize: Int = 20,	
        currentPage: Int = 1	
    ): NegotiableQuotesOutput @doc(description: "List of negotiable quotes available to customer")	
}
```

We allow to query remaining items, which might be very helpful for updating the list of the items on the client side after mutation is finished. It would be easier to avoid issue of having a stale data. If developer doesn’t want to update the list from the output of this query some data will have to be retrieved anyways and ignored, which is ok.

```
    successful: Boolean!
}
```

Need to check if we get false in case of an error or it will return an error (not for this specific mutation, but for similar mutations we have in the code). I would expect it to return false in case of an error, but it's not very convenient for a developer. You also can’t update list on the client side with output from the server with out making a separate query.

Other options. Not in our code base, but I have seen other APIs, return list of removed items and sometimes count. I like this a little better than second option, but the first option I like the most.

I would propose to go with the first option and thing about applying this pattern in other places. What do you think @DrewML, @RakeshJesadiya?

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
RakeshJesadiya: @melnikovi Thanks I am looking for that filterable in the output. I have that query only. thanks. I need to look core example of how we can filter out the result on the output.

RakeshJesadiya:
@DrewML @melnikovi I have reverted back the changes for the Close and Delete Negotiable quote.
I hope this is the last final changes for the Negotiable Quote. Waiting for your comment and approvals.

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
```diff
@@ -104,7 +99,7 @@ type DeleteNegotiableQuotesOutput {
...
+    negotiableQuotes(
```
melnikovi, line 102:
```suggestion
    negotiable_quotes(
```

Should be a snake case per our conventions.

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
```diff
@@ -116,12 +111,13 @@ input CloseNegotiableQuotesInput {
 }
 
 type CloseNegotiableQuotesOutput {
-    closed_quotes: [NegotiableQuote]! @doc(description: "Quotes that were just closed")
-    negotiable_quotes(
+    closed_quotes: [NegotiableQuote!] @doc(description: "Quotes that were just closed")
+    #optionally display all negotiable quotes
+    negotiableQuotes(
```
melnikovi, line 116:
```suggestion
    negotiable_quotes(
```

design-documents/graph-ql/coverage/negotiableQuotes.graphqls
RakeshJesadiya: @melnikovi Should we change to that snake case on line no. 8 also?

melnikovi: On the first level it's camel case.

PR #423, opened 2020-08-19
Update GraphQl company create mutation output

munkhulzii:
## Problem

The type of CreateCompanyOutput for company create mutation is defined as type "Company". The fields of Company in current implementation have their own resolver in which the authorization will be checked. It would fail for creating company because in this case we do not have a logged in user. 

## Solution

Update schema with base fields like the response for company registration in frontend. No additional fields like "role", "users", "structure" etc.

## Requested Reviewers

@paliarush
@cpartica 
@danielrenaud

### DIFF
--- a/design-documents/graph-ql/coverage/company.md
+++ b/design-documents/graph-ql/coverage/company.md
@@ -174,7 +174,18 @@ type DeleteCompanyTeamOutput @doc(description: "Delete company team output data 
 }
 
 type CreateCompanyOutput @doc(description: "Create company output data schema.") {
-    company: Company! @doc(description: "New company instance.")
+    company: CompanyOutput! @doc(description: "Create company output instance.")
+}
+
+type CompanyOutput {
+    id: ID! @doc(description: "Company id.")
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/company.md
```diff
@@ -174,7 +174,18 @@ type DeleteCompanyTeamOutput @doc(description: "Delete company team output data
 }
 
 type CreateCompanyOutput @doc(description: "Create company output data schema.") {
-    company: Company! @doc(description: "New company instance.")
+    company: CompanyOutput! @doc(description: "Create company output instance.")
+}
+
+type CompanyOutput {
```
melnikovi, line 180: Propose to include these fields under `CreateCompanyOutput` and not have `CompanyOutput`. Also, let's rename the following fields for consistency and copy description for the fields from `CompanyCreateInput`.
`name` to `company_name`
`email` to `company_email`
`vat_id` to `vat_tax_id`

melnikovi:
Decided to not modify `CreateCompanyOutput` type.

PR #424, opened 2020-08-21
AEM uid updates

DrewML:
## Problem

The AEM team gave us a list of fields they want us to fix in accordance with our [ID Improvement Plan](https://github.com/magento/architecture/pull/415).

## Solution

Apply the schema changes from the ID Improvement Plan, for now just in the places flagged as high priority for AEM.

The request list is:

> CartItemUpdateInput.cart_item_id
RemoveItemFromCartInput.cart_item_id
CartItemQuantity.cart_item_id
CategoryInterface.id and Query.category(id: Int)
ProductInterface.id
Breadcrumb.category_id
Customer.id

Note that `CartItemQuantity` is already deprecated in core, so that field was skipped. It was only used in `cart_items` fields, but there is now a `cart_items_v2` that returns `CartItemInterface` instead (which is updated as part of this PR).

## Notes
The schemas we're changing were not fully present in this repository, so I copied over up-to-date copies from `2.4-develop`.

I split up my commits between changes and code copying to make it easier to review. See the `Commits` tab to aid your review

[This link](https://github.com/magento/architecture/pull/424/files/b83b5384e57658ee6e374a23a44b6738d40f9a62..a66f16687c053f43926be93bfed9a288dcf4c13c) should include all the commits in a diff, without the code copying.

## Requested Reviewers

@melnikovi 
@nrkapoor


design-documents/graph-ql/coverage/quote.graphqls
```diff
@@ -73,7 +73,10 @@ input UpdateCartItemsInput {
 }
 
 input CartItemUpdateInput {
-    cart_item_id: Int!
+    # Implementation Note: The resolver should accept either `cart_item_id` or `cart_item_uid`,
+    # but _must_ return an error when both fields are provided in a request
+    cart_item_id: Int! @deprecated(reason: "Use `CartItemUpdateInput.cart_item_uid` instead")
+    cart_item_uid: ID! @doc(description: "Unique Identifier from objects implementing `CartItemInterface`")
```
cjelger, line 79: Maybe I'm wrong but I think there is a problem here. You cannot have both arguments mandatory, otherwise the schema will reject any request if both `cart_item_id` and `cart_item_uid` are not set. I quickly did a test with GraphiQL in Chrome with a Magento schema, and I'm for example forced to set `cart_id` in `AddSimpleProductsToCartInput` (it's also declared with `!`). This is going to be problematic, especially if you want to return an error when both fields are provided.

You could resolve this by adding an extra input type like
```
input CartItemIdInput {
    cart_item_id: Int @deprecated(reason: "Use `CartItemUpdateInput.cart_item_uid` instead")
    cart_item_uid: ID @doc(description: "Unique Identifier from objects implementing `CartItemInterface`")
}
```
and make this one mandatory like in
```
input CartItemUpdateInput {
    cart_item_id: Int
    cart_item_id_input: CartItemIdInput!
}
```
but this would make everything complicated and awkward. The easy fix is to remove the mandatory parts for both fields, but this is also not very nice, and you cannot only enforce the new ID or you'll break backwards-compatibility.

design-documents/graph-ql/coverage/quote.graphqls
```diff
@@ -83,7 +83,10 @@ input CartItemUpdateInput {
 
 input RemoveItemFromCartInput {
     cart_id: String!
-    cart_item_id: Int!
+    # Implementation Note: The resolver should accept either `cart_item_id` or `cart_item_uid`,
+    # but _must_ return an error when both fields are provided in a request
+    cart_item_id: Int! @deprecated(reason: "Use `RemoveItemFromCartInput.cart_item_uid` instead")
+    cart_item_uid: ID! @doc(description: "Unique Identifier from objects implementing `CartItemInterface`")
```
cjelger, line 89: Same comment as for `CartItemUpdateInput`

design-documents/graph-ql/coverage/catalog.graphqls
```diff
@@ -11,7 +11,10 @@ type Query {
     ): Products
     @resolver(class: "Magento\\CatalogGraphQl\\Model\\Resolver\\Products") @doc(description: "The products query searches for products that match the criteria specified in the search and filter attributes.") @cache(cacheIdentity: "Magento\\CatalogGraphQl\\Model\\Resolver\\Product\\Identity")
     category (
-     id: Int @doc(description: "Id of the category.")
+    # Implementation Note: The resolver should accept either `id` or `categoryInterfaceUID`,
+    # but _must_ return an error when both fields are provided in a request
+     id: Int @deprecated(reason: "Use the `categoryInterfaceUID` argument instead")
+     categoryInterfaceUID: ID @doc(description: "Unique identifier from objects implementing `CategoryInterface`")
```
cjelger, line 17: Why not just `uid`? (I think with the description it's clear what it is)

design-documents/graph-ql/coverage/catalog.graphqls
```diff
@@ -267,7 +267,8 @@ interface CategoryInterface @typeResolver(class: "Magento\\CatalogGraphQl\\Model
 }
 
 type Breadcrumb @doc(description: "Breadcrumb item."){
-    category_id: Int @doc(description: "Category ID.")
+    category_id: Int @deprecated(reason: "Use the `category_interface_uid` field instead")
+    category_interface_uid: ID! @doc(description: "Unique identifier from objects implementing CategoryInterface")
```
cjelger, line 271: Why not just `category_uid` to be more consistent with `category_id`?

cjelger (commented):
I guess you also checked all possible mutation input types and variables to support the new uid types. Check my comments regarding the cart input types, I think this is a generic problem you might have everywhere where you want to deprecate the old id in input types.

design-documents/graph-ql/coverage/quote.graphqls
rafaelbarbosa-diconium: I guess both as mandatory would also be a problem for backwards compatibility. Older clients will not be sending cart_item_uid, right?

DrewML: > Maybe I'm wrong but I think there is a problem here

You are not wrong - good catch! They're both marked as non-nullable so this fails the query at the validation step, before it ever hits the resolvers.

This is unfortunately one of those problems that's hard to model without polymorphic input types, [which there has been a _lot_ of discussion/debate about for awhile](https://github.com/graphql/graphql-spec/blob/d0b76b0bd285b61df2210a57144b9a70d8d7dc4e/rfcs/InputUnion.md)

How would you feel about the following change, which is just re-iterating your idea:

1. Both fields (`cart_item_id` and `cart_item_uid`) are made nullable (removing `!`) in schema
2. Validation of required fields (i.e. the mandate that one exists) is validated in the resolver for the mutation

This would provide the following benefits:
1. Backwards-compatible for existing clients
1. Impossible to pass both IDs without an error
1. Impossible to pass no IDs without an error
1. (Mostly) Safe to make `cart_item_uid` non-nullable (`!`) later when we remove `cart_item_id`

But with the following downsides:
1. We lose some useful information in the schema, because we're side-stepping the type system and using our own validation in a way that isn't easy to introspect.

design-documents/graph-ql/coverage/quote.graphqls
DrewML: 👍 We'll keep discussing the general problem in that thread, and apply it across this PR once we reach consensus

design-documents/graph-ql/coverage/catalog.graphqls
DrewML: Let me try and provide some quick context around the naming.

https://github.com/magento/architecture/blob/master/design-documents/graph-ql/id-improvement-plan.md

> **Terminology**
> **Primary Identifier**: An ID owned by the current Object Type (ex: Product.id)
> **Foreign Identifier**: An ID referencing another Object Type (ex: PaymentMethodInput.code)

We've tried, for object fields, to make it easy/predictable to determine what an `ID` field is referencing, by including the concrete type in the name of the field. The general rule of thumb has been `typename_uid` to reference an `ID` value from another object type.

------

Unfortunately, as you've pointed out, there's 2 problems:

1. When you're doing a simple lookup of a known type like `CategoryTree` with an argument, it's redundant/verbose to specify `category(categoryInterfaceUID: "<value>")`
2. That linked document didn't account for naming conventions when referencing an `ID` field defined in an interface, rather than an object type (which you pointed out in another comment about `Breadcrumb.category_interface_uid`)

For this case (a query on the `Query` root, doing a lookup by ID) I agree it makes sense to just switch to `uid`, since it's obvious to a consumer of the API 👍

design-documents/graph-ql/coverage/catalog.graphqls
DrewML: Provided a bit of background context in https://github.com/magento/architecture/pull/424#discussion_r477626040 for why it's named this way.

I'm a bit conflicted about `category_interface_uid` vs `category_id`. On one hand, `category_uid` seems like a cleaner API surface. 

My concern is that `Category` isn't actually a type in GraphQL (we have `CategoryInterface` and `CategoryTree`) so there's a small bit of misdirection here.

One goal I've had for our refactoring of IDs has been to have a shared naming convention, that would allow us to better analyze the GraphQL schema in the future. Right now, there's no way to programmatically find all references to a specific `ID`. You mentioned in a comment that:

> I guess you also checked all possible mutation input types and variables to support the new uid types.

Right now, any tasks like this are manual and have to be done by a human. The goal with this naming convention is to make it possible to have the schema answer these questions in the future

design-documents/graph-ql/coverage/quote.graphqls
cjelger: +1 for this solution, I don't think there is any better solution that doesn't introduce its own set of issues or backwards-incompatible problems. And thanks for the link on `inputunion`, I didn't know about that, indeed very interesting reading.

design-documents/graph-ql/coverage/catalog.graphqls
cjelger: Thank you for the extra background info: I understand now the naming, it's indeed a good idea if you can later automatically analyse and process/refactor the schema to find all these foreign identifiers.

design-documents/graph-ql/coverage/catalog.graphqls
DrewML: Pushed update changing arg name to `uid`

design-documents/graph-ql/coverage/quote.graphqls
DrewML: Fixed in a66f16687c05

design-documents/graph-ql/coverage/quote.graphqls
DrewML: Fixed in a66f166

cjelger (approved):

design-documents/graph-ql/coverage/quote.graphqls
```diff
@@ -0,0 +1,418 @@
...
+    #    a field error advising the client that `cart_item_uid` is required
+    # 2. If _both_ `cart_item_id` and `cart_item_uid` are provided, raise a field error advising
+    #    the client to only use `cart_item_uid`
+    #
+    # GraphQL does not provide a way for an `input` field to allow use of only 1 field or another,
+    # which is why we're enforcing non-nullability in the resolver, instead of via the schema language.
+    # When `cart_item_id` is removed in a future version, we can mark `cart_item_uid` as non-nullable
+    # with minimal disruption
+    cart_item_id: Int @deprecated(reason: "Use the `cart_item_uid` field instead")
+    cart_item_uid: ID @doc(description: "Required field. Unique Identifier from objects implementing `CartItemInterface`")
```
melnikovi, line 89:
```suggestion
    cart_item_interface_uid: ID @doc(description: "Required field. Unique Identifier from objects implementing `CartItemInterface`")
```

Need to have word interface in it if we want to use the convention.

design-documents/graph-ql/coverage/quote.graphqls
```diff
@@ -73,20 +73,40 @@ input UpdateCartItemsInput {
...
```
melnikovi, line 109:
```suggestion
    cart_item_interface_uid: ID @doc(description: "Required field. Unique Identifier from objects implementing `CartItemInterface`")
```

PR #427, opened 2020-08-27
magento2/issues/29479 Returns : graphql implementation

sasha19957099:
## Problem

Working with ticket https://github.com/magento/magento2/issues/29479 I see that in RMA module we have additional fields in the request (order_id, customer_custom_email)

### DIFF
--- a/design-documents/graph-ql/coverage/returns.graphqls
+++ b/design-documents/graph-ql/coverage/returns.graphqls
@@ -13,6 +13,8 @@ type Mutation {
 }
 
 input RequestReturnInput {
+    order_id: ID!
+    customer_custom_email: String!
     items: [RequestReturnItemInput!]!
     comment_text: String
 }
--- a/design-documents/graph-ql/coverage/returns.graphqls
+++ b/design-documents/graph-ql/coverage/returns.graphqls
@@ -14,7 +14,7 @@ type Mutation {
 
 input RequestReturnInput {
     order_id: ID!
-    customer_custom_email: String!
+    contact_email: String
     items: [RequestReturnItemInput!]!
     comment_text: String
 }
--- a/design-documents/graph-ql/coverage/returns.md
+++ b/design-documents/graph-ql/coverage/returns.md
@@ -211,6 +211,8 @@ Existing schema of `Attribute` and `AttributeOption` must be extended to provide
 mutation {
   requestReturn(
     input: {
+      order_id: 12345
+      customer_custom_email: "returnemail@magento.com"
       items: [
         {
           order_item_id: "absdfj2l3415", 
--- a/design-documents/graph-ql/coverage/returns.md
+++ b/design-documents/graph-ql/coverage/returns.md
@@ -211,7 +211,7 @@ Existing schema of `Attribute` and `AttributeOption` must be extended to provide
 mutation {
   requestReturn(
     input: {
-      order_id: 12345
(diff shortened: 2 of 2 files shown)


design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -13,6 +13,8 @@ type Mutation {
 }
 
 input RequestReturnInput {
+    order_id: ID!
+    customer_custom_email: String!
```
melnikovi, line 17:
```suggestion
    contact_email: String
```

Let's make field optional as it's optional in the Luma.

nrkapoor:
@melnikovi can we also update all the "id" fields to "uid" to be consistent with the new proposal by @DrewML?

design-documents/graph-ql/coverage/returns.md
```diff
@@ -211,6 +211,8 @@ Existing schema of `Attribute` and `AttributeOption` must be extended to provide
 mutation {
   requestReturn(
     input: {
+      order_id: 12345
```
melnikovi, line 214:
```suggestion
      order_id: 000000001
```

design-documents/graph-ql/coverage/returns.graphqls
```diff
@@ -13,6 +13,8 @@ type Mutation {
 }
 
 input RequestReturnInput {
+    order_id: ID!
```
melnikovi, line 16:
```suggestion
    order_uid: ID!
```

Suggestion from @nrkapoor, pending discussion.

design-documents/graph-ql/coverage/returns.md
```diff
@@ -211,6 +211,8 @@ Existing schema of `Attribute` and `AttributeOption` must be extended to provide
 mutation {
   requestReturn(
     input: {
+      order_id: 000000001
```
melnikovi, line 214:
```suggestion
      order_uid: MQ==
```

Suggestion from @nrkapoor, pending discussion.

melnikovi:
Currently `CustomerOrder` type has `id` and `number` fields, in the future id will be replaced with `uid`.

- `id` - unique identifier for the order
- `number` - order number that is visible to a customer; not uniquer for all orders; unique in the scope of website and store; it's also sequential; generation can be customized
- `uid` - see [this](https://github.com/magento/architecture/pull/415) proposal for mode details

We can use `uid` and `number` fields in the mutations input types that reference CustomerOrder`, here are some pros and cons.

`uid` - will be aligned with other places, we are going to a state when types reference each other by `uid` and `uid` used in mutations.
`number` - in some cases it would make more sense for querying data. If we suggest to use it for mutations it might not follow the pattern used in other mutations and be confusing for developers.

We can't have just `uid` and store value of `number` there as `uid` supposed to be unique and `number` currently is not unique. Not having `uid` and having `number` for `CustomerOrder` has similar implications.

It makes sense to have both fields and use `uid` in mutations that reference `CustomerOrder` type.

paliarush:
`uid` should be of type `ID` (not human-readable)
`number` should be of type `String` (human-readable)

design-documents/graph-ql/coverage/returns.graphqls
melnikovi: Discussed with @nrkapoor and agree to replace `id` with `uid` everywhere in schema for returns.

PR #428, opened 2020-08-27
Document existing GraphQL naming conventions

DrewML:
## Problem

Just documenting the conventions that are already being followed, so we have something to link to.

### DIFF
--- a/design-documents/graph-ql/naming-conventions.md
+++ b/design-documents/graph-ql/naming-conventions.md
@@ -47,12 +47,40 @@ type Query {
 
 Top-level fields on Query and Mutation should always be _camelCase_.
 
-```
+```graphql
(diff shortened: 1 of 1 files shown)


DrewML:
TODO before merge: Need to add a section about top-level fields on `Query` and `Mutation` being camel case instead of snake case

design-documents/graph-ql/naming-conventions.md
```diff
@@ -41,4 +41,18 @@ type Query {
         twoWords: String
     ): String
 }
-```
\ No newline at end of file
+```
+
+## Top-level Fields on Query and Mutation
```
DrewML, line 46: Thanks @melnikovi for pushing this!

PR #429, opened 2020-08-31
[GraphQL] Remove dynamic attributes from ProductInterface

akaplya:
## Problem

At the moment, Magento exposes Product Interface as an interface with the dynamic structure. As a result, the Product Interface in Magento has a state. This creates unnecessary complexity on the GraphQL tier. (it is challenging to cache, read schema, aggregate stores behind the same gateway so on).

## Solution

Remove the state from the interface.

### DIFF
--- a/design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
+++ b/design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
@@ -1,0 +1,45 @@
+# Remove product dynamic attributes from ProductInterface
+
+## Current state
+
+Currently, Magento exposes EAV attributes as part of the product schema as top-level fields of ProductInterface. As consequences:
+* The merchant can change the GraphQL from the admin, which makes it hard to cache GraphQL schema.
+* Regular product interface overloaded with fields and almost unreadable.
+* It is quite often practice to create product programmatically, so the number of attributes could be really excessive with tens thousands of attributes.
+* Two Magento instances do not have the same GraphQL schema. As a result, it is near impossible to build a simple client/SDK, which could be easily transferred from one instance to another.
+* It is impossible to aggregate the several Magento instances behind the common [BFF](https://docs.microsoft.com/en-us/azure/architecture/patterns/backends-for-frontends).
+
+## Proposed solution
+
+We should to remove attributes dynamic attributes,
+except system attributes which makes sense for the storefront,
+from the product top level and use a [dynamic container](https://github.com/magento/architecture/blob/master/design-documents/graph-ql/custom-attributes-container.md) instead.
+
+
+```graphql
+## ProductInteface with system EAV attributes 
+type ProductInterface {
+    uid: ID
+    name: String
+    sku: String
+    attribute_set_id: Int
+    description: ComplexTextValue
+    short_description: ComplexTextValue
+    meta_title: String
+    meta_keyword: String
+    meta_description: String
+    created_at: String
+    updated_at: String
+    custom_attributes: [CustomAttribute]
+    ...
+}
+```
+
+## System EAV attributes that should be deprecated at the storefront
+
+* `special_price: Float`  These fields should not be used a the source of the discount since the discount also could be caused by group price or rule price, which are not reflected in this field.
+The same is true for `special_from_date: String` and `special_to_date`.
+
+* `options_container: String` - this field just an exposure of `catalog_product_entity` table field, information about product options could be explicitly retrieved from the corresponding field. 
+* `manufacturer: Int` - just a sample attribute from Magento 1.x, we can move it to the custom attributes container.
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
```diff
@@ -0,0 +1,45 @@
...
+    created_at: String
+    updated_at: String
+    custom_attributes: [CustomAttribute]
+    ...
+}
+```
+
+## System EAV attributes that should be deprecated at the storefront
+
+* `special_price: Float`  These fields should not be used a the source of the discount since the discount also could be caused by group price or rule price, which are not reflected in this field.
```
melnikovi, line 40: I think you can use it with out group prices or rule prices. Do you propose to use catalog price rules for simple cases? Seems like a large change, since we will have to deprecate and remove this feature completely from Magento.

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
```diff
@@ -0,0 +1,45 @@
...
+    meta_keyword: String
+    meta_description: String
+    created_at: String
+    updated_at: String
+    custom_attributes: [CustomAttribute]
+    ...
+}
+```
+
+## System EAV attributes that should be deprecated at the storefront
```
melnikovi, line 38: How do you decide which attributes to move to a new container? Why not move for example `new_from_date`, `new_to_date`?

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
```diff
@@ -0,0 +1,45 @@
+# Remove product dynamic attributes from ProductInterface
+
+## Current state
+
+Currently, Magento exposes EAV attributes as part of the product schema as top-level fields of ProductInterface. As consequences:
+* The merchant can change the GraphQL from the admin, which makes it hard to cache GraphQL schema.
+* Regular product interface overloaded with fields and almost unreadable.
+* It is quite often practice to create product programmatically, so the number of attributes could be really excessive with tens thousands of attributes.
+* Two Magento instances do not have the same GraphQL schema. As a result, it is near impossible to build a simple client/SDK, which could be easily transferred from one instance to another.
+* It is impossible to aggregate the several Magento instances behind the common [BFF](https://docs.microsoft.com/en-us/azure/architecture/patterns/backends-for-frontends).
```
melnikovi, line 10: Extensions can add new attributes on the top level too. Why do we need to hide several Magento instances behind common BFF?

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
```diff
@@ -0,0 +1,45 @@
...
+    ...
+}
+```
+
+## System EAV attributes that should be deprecated at the storefront
+
+* `special_price: Float`  These fields should not be used a the source of the discount since the discount also could be caused by group price or rule price, which are not reflected in this field.
+The same is true for `special_from_date: String` and `special_to_date`.
+
+* `options_container: String` - this field just an exposure of `catalog_product_entity` table field, information about product options could be explicitly retrieved from the corresponding field.
```
melnikovi, line 43: Agree, makes sense to deprecate it.

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
akaplya: `special price` that we have here is the admin business logic field and does not represent the special price that we need at the storefront and it can not be used for the scenarios at the storefront.
Also it could be very misleading; the only and the right way to see the discounted price is to check the final and regular prices.

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
akaplya: Each eav attribute has a property is_system. If the attribute is system this is no way to remove it and very limited capabilities to manage it. that why is safe to assume that this attribute is not dynamic and can be represented as top-level fied.

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
akaplya: for instance, if we need to add a common gateway with query cache and want to make this gateway saas solution.

design-documents/graph-ql/coverage/catalog/remove-product-attributes.md
akaplya: catalog rule price, group price, special price all of them admin fields with the only purpose select the least of them and compare to the base price if the resulting price is lower than the product has a special price that makes sense for the storefront, raw EAV attribute itself does not have any sense.

melnikovi (approved):

PR #439, opened 2020-09-10
Rename all id fields to uid in cart.graphqls

nrkapoor:
## Problem

Schema has a lot of fields named id which doesn't align with the new naming convention.

## Solution

Updating all id fields to uid to be consistent with the new naming convention.

## Requested Reviewers

@melnikovi @prabhuram93 @DrewML

### DIFF
--- a/design-documents/graph-ql/coverage/cart/Cart.graphqls
+++ b/design-documents/graph-ql/coverage/cart/Cart.graphqls
@@ -3,7 +3,7 @@ type Query {
 }
 
 input CartQueryInput {
-    cart_id: String!
+    cart_id: ID!
 }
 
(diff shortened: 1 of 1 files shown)


melnikovi (approved):

design-documents/graph-ql/coverage/cart/Cart.graphqls
```diff
@@ -3,15 +3,16 @@ type Query {
 }
 
 input CartQueryInput {
-    cart_id: String!
+    cart_id: ID!
```
melnikovi, line 6: We actually don't have `CartQueryInput` type, will create a separate PR to update.

PR #455, opened 2020-10-20
Update the `schema.graphql` for additional output attributes

andrewbess:
## Problem

To add a new company team or company user into the existing company team (company user), a customer should transmit the encoded structure ID using attribute `target_id`. However, the user has no way to find out this data.

## Solution

Updating the schema.graphql with adding the output attribute `structure_id` into the types `CompanyStructureItem`, `CompanyTeam`, `Customer`. This solution gives a possibility to get the `structure_id` in responses of the mutations and queries.

## Requested Reviewers

@mauragcyrus @cpartica @prabhuram93

### DIFF
--- a/design-documents/graph-ql/coverage/b2b/company.md
+++ b/design-documents/graph-ql/coverage/b2b/company.md
@@ -120,6 +120,7 @@ union CompanyStructureEntity = CompanyTeam | Customer
 type CompanyStructureItem @doc(description: "Company Team and Customer structure") {
     uid: ID! @doc(description: "ID of the item in the hierarchy")
     parent_id: ID @doc(description: "ID of the parent item in the hierarchy")
+    structure_id: ID! @doc(description: "ID of the company structure")
     entity: CompanyStructureEntity
 }
 
@@ -131,6 +132,7 @@ type CompanyTeam @doc(description: "Company Team entity output data schema.") {
     uid: ID! @doc(description: "Team id.")
     name: String @doc(description: "Team name.")
     description: String @doc(description: "Team description.")
+    structure_id: ID! @doc(description: "ID of the company structure")
 }
 
 input CompanyUsersFilterInput @doc(description: "Defines the input filters for a Company Users query") {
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/b2b/company.md
```diff
@@ -315,5 +317,6 @@ type Customer {
     team: CompanyTeam @doc(description: "Company User team data.")
     telephone: String @doc(description: "Company User phone number.")
     status: CompanyUserStatusEnum @doc(description: "Company User status.")
+    structure_id: ID! @resolver(class: "Magento\\CompanyGraphQl\\Model\\Resolver\\Company\\Customer\\CompanyStructure") @doc(description: "ID of the company structure")
```
cpartica, line 320: this is good, just remove the @resolver here. We don't need it in this repo

cpartica (changes requested):

design-documents/graph-ql/coverage/b2b/company.md
andrewbess: Hello @cpartica
Thank you for your note 
I fixed it.
Please recheck.

Thank you in advance.

design-documents/graph-ql/coverage/b2b/company.md
```diff
@@ -120,6 +120,7 @@ union CompanyStructureEntity = CompanyTeam | Customer
 type CompanyStructureItem @doc(description: "Company Team and Customer structure") {
     uid: ID! @doc(description: "ID of the item in the hierarchy")
     parent_id: ID @doc(description: "ID of the parent item in the hierarchy")
+    structure_id: ID! @doc(description: "ID of the company structure")
```
cpartica, line 123: can you replace structure id and parent id with parent_uid and company_structure_uid? 

Also Ideally instead of parent_uid we would have the originating object name there somewhere  contained in this 'parent_uid' field
this schema needs to represent the future version

cpartica (changes requested):

PR #456, opened 2020-10-20
Update proposal for custom attributes schema

melnikovi:
## Problem

Current schema for custom attributes doesn't support attributes that can have multiple values (checkbox, multiselect).

### DIFF
--- a/design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
+++ b/design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
@@ -13,13 +13,16 @@ To account for dynamic nature of EAV attributes and the need of "getting all fie
 ```graphql
 type CustomAttribute {
     code: String!
-    value: [String]! # We want to account fo attributes that have single (text, dropdown) and multiple values (checkbox, multiselect)
+    values: [String]! # We want to account fo attributes that have single (text, dropdown) and multiple values (checkbox, multiselect)
 }
+```
 
-# We could also make value complex type to be able add more fields in the future
+We could also make value complex type to be able add more fields in the future, but this doesn't seem necessary at this point.
+
+```graphql
 type CustomAttribute {
     code: String!
-    value: [CustomAttributeValue]!
+    values: [CustomAttributeValue]!
 }
 
 type CustomAttributeValue {
@@ -34,16 +37,12 @@ type CustomAttributeInterface {
     code: String!
 }
 
-type CustomAttribute extends CustomAttributeInterface {
-    value: CustomAttributeValue!
-}
-
-type CustomAttributeMulti extends CustomAttributeInterface {
-    values: [CustomAttributeValue]!
-}
-
-type CustomAttributeValue {
+type SingleValueCustomAttribute extends CustomAttributeInterface {
     value: String!
+}
+
+type MultipleValuesCustomAttribute extends CustomAttributeInterface {
+    values: [String]!
 }
 ```
 
--- a/design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
+++ b/design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
@@ -8,11 +8,41 @@ One workaround for "getting all fields" is based on schema introspection, it all
 
 # Proposed solution
 
-To account for dynamic nature of EAV attributes and the need of "getting all fields" in product search queries, `custom_attributes: [CustomAttribute]!` container will be introduced. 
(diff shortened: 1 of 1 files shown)


nrkapoor:
@melnikovi Does this align with the custom attribute proposal from @akaplya for Product? https://github.com/magento/architecture/pull/429

design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
```diff
@@ -8,13 +8,42 @@ One workaround for "getting all fields" is based on schema introspection, it all
 
 # Proposed solution
 
-To account for dynamic nature of EAV attributes and the need of "getting all fields" in product search queries, `custom_attributes: [CustomAttribute]!` container will be introduced. 
+To account for dynamic nature of EAV attributes and the need of "getting all fields" in product search queries,we can introduce `custom_attributes: [CustomAttribute]!` container. 
 
 ```graphql
 type CustomAttribute {
```
mslabko, line 14: I thinks we have to add "type" opinion here to be able to render the appropriate template (datetime, text, multi-select, checkbox ...)

E.g.
```
"attributes":[
    {
      "attributeCode":"multiselect_attribute",
      "type":"multiselect",
      "value":[
        {
          "id":"21",
          "value":"Option 1"
        },
        {
          "id":"22",
          "value":"Option 2"
        }
      ]
    }
  ]
```

melnikovi: Attributes are going to be displayed as information and not as fields you can control, so I don't think it's necessary.

mslabko: it will be useful to render field properly on U without complex logic. E.g. for "Date" field - "10-11-2020"

design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
```diff
@@ -8,13 +8,42 @@ One workaround for "getting all fields" is based on schema introspection, it all
...
+```
+
+Alternative approach would be is to introduce an interface `custom_attributes: [CustomAttributeInterface]!`, but it seems more complicated.
+
+```graphql
+type CustomAttributeInterface {
+    code: String!
+}
+
+type SingleValueCustomAttribute extends CustomAttributeInterface {
```
cpartica, line 40: even if this adds more complexity, it's the graphql way to do it

design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
```diff
@@ -8,13 +8,42 @@ One workaround for "getting all fields" is based on schema introspection, it all
 
 # Proposed solution
 
-To account for dynamic nature of EAV attributes and the need of "getting all fields" in product search queries, `custom_attributes: [CustomAttribute]!` container will be introduced. 
+To account for dynamic nature of EAV attributes and the need of "getting all fields" in product search queries, we can introduce `custom_attributes: [CustomAttribute]!` container (recommended approach).
```
cpartica, line 11: do fragments bring any value for PWA since we still prefer these flat single structure arrays vs infterfaces?

design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
cpartica: yes, they are needed for the forms

design-documents/graph-ql/coverage/custom-attributes/attributes-metadata.graphqls
```diff
@@ -0,0 +1,222 @@
+type Query {
+    customAttributeMetadataV2(attributes: [AttributeMetadataInput!]!): CustomAttributeMetadata
```
cpartica, line 2: A question here if we should require at least one attribute or no attributes will return all schema

design-documents/graph-ql/coverage/custom-attributes/custom-attributes-container.md
paliarush: We have this data already available in the scope of `customAttributeMetadata` query. Since update frequency for attribute metadata and product attribute values is different, it makes sense to separate them to avoid excessive payload size.

paliarush:
1. We need to make sure that data and metadata is not mixed in the same queries because they have different life cycles. Metadata can be cached on the client and reused between queries for data.
2. Each schema modification must be covered with the example of usage. Suggestion is to split the original proposal into smaller ones, which address one problem at a time.

PR #458, opened 2020-10-22
Support of custom attributes values for entities like customer/product/address

melnikovi:
## Problem

Support of custom attributes for customer and customer address.

### DIFF
--- a/design-documents/graph-ql/coverage/custom-attributes/attributes-metadata.md
+++ b/design-documents/graph-ql/coverage/custom-attributes/attributes-metadata.md
@@ -40,9 +40,8 @@ type Attribute {
 
 type AttributeOption {
     uid: ID!
+    value: String @deprecated(reason: "use `uid` instead")
     label: String
-    value: String @deprecated(reason: "use `values` instead")
-    values: [ID]! @doc(description: "Array as container of values of the attribute")
 }
 ```
 
--- a/design-documents/graph-ql/coverage/custom-attributes/attributes-metadata.md
+++ b/design-documents/graph-ql/coverage/custom-attributes/attributes-metadata.md
@@ -34,6 +34,7 @@ type Attribute {
     attribute_type: String
     entity_type: String
     input_type: String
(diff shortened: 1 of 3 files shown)


design-documents/graph-ql/coverage/custom-attributes/attributes-metadata.md
```diff
@@ -29,16 +29,20 @@ type CustomAttributeMetadata {
...
     input_type: String
+    is_required: Bolean!
 }
 
 type AttributeOption {
+    uid: ID!
     label: String
-    value: String
+    value: String @deprecated(reason: "use `values` instead")
+    values: [ID]! @doc(description: "Array as container of values of the attribute")
```
cpartica, line 45: this will be sufficient for single and multiple values

design-documents/graph-ql/coverage/customer/customer.graphqls
```diff
@@ -128,10 +168,23 @@ type CustomerAddress @doc(description: "CustomerAddress contains detailed inform
     vat_id: String @doc(description: "The customer's Value-added tax (VAT) number (for corporate customers)")
     default_shipping: Boolean @doc(description: "Indicates whether the address is the default shipping address")
     default_billing: Boolean @doc(description: "Indicates whether the address is the default billing address")
-    custom_attributes: [CustomerAddressAttribute] @deprecated(reason: "Custom attributes should not be put into container")
+    custom_attributes: [CustomerAddressAttribute] @deprecated(reason: "Use `stored_custom_attributes` instead.")
```
cpartica, line 171: I tried to reuse Attribute from the Metadata query.
Using the already popular and known term  "custom_attributes", It’s not very clear that this will return all information about the storefront dedicated attributes and their options and values (not sure about the options part)
 but also this type has to contain the stored( saved )by the user values. Same are used on the input as well. The difference is the output contains full meta of the attribute + picked values and input only contains refernces of the attribute and picked values

PR #462, opened 2020-11-06
Improvements to directory.graphqls to include more configation settings

paales:
## Problem

To build proper customer address forms, we need to have some additional information.

## Solution

I've split this PR in two commits, the first is a copy of the current schema in the magento repository, the second are my modifications on it: https://github.com/magento/architecture/commit/5c8ab137d02c8b36a0d063d56883bdf743e5f3c2

Introduce additional country fields:

- add field to `Country` type: `region_required`
- add field to `Country` type: `postcode_required`
- add field to `Country` type: `name`
- add field to `Country` type: `eu`
- deprecate `full_name_locale` in favor of `name`
- deprecate `full_name_english` in favor of `name`
- add field to `StoreConfig` type: `default_country`
- add field to `StoreConfig` type: `allow_optional_state`
- adds field docs with sorting information: This should be a nice quality of life improvement for the frontend implementor so they don't have to do this on the frontend.

## Requested Reviewers

@buskamuza Taggin you because you seem to be responsible for this component
@cpartica Tagging you of the discussion on Slack

### DIFF
--- a/design-documents/graph-ql/coverage/directory.graphqls
+++ b/design-documents/graph-ql/coverage/directory.graphqls
@@ -1,0 +1,39 @@
+# Copyright © Magento, Inc. All rights reserved.
+# See COPYING.txt for license details.
+
+type Query {
+    currency: Currency @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Currency") @doc(description: "The currency query returns information about store currency.") @cache(cacheable: false)
+    countries: [Country] @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Countries") @doc(description: "The countries query provides information for all countries.") @cache(cacheable: false)
+    country (id: String): Country @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Country") @doc(description: "The countries query provides information for a single country.") @cache(cacheable: false)
+}
+
+type Currency {
+    base_currency_code: String
+    base_currency_symbol: String
+    default_display_currecy_code: String @deprecated(reason: "Symbol was missed. Use `default_display_currency_code`.")
+    default_display_currency_code: String
+    default_display_currecy_symbol: String @deprecated(reason: "Symbol was missed. Use `default_display_currency_symbol`.")
+    default_display_currency_symbol: String
+    available_currency_codes: [String]
+    exchange_rates: [ExchangeRate]
+}
+
+type ExchangeRate {
+    currency_to: String
+    rate: Float
+}
+
+type Country {
+    id: String
+    two_letter_abbreviation: String
+    three_letter_abbreviation: String
+    full_name_locale: String
+    full_name_english: String
+    available_regions: [Region]
+}
+
+type Region {
+    id: Int
+    code: String
+    name: String
+}
--- a/design-documents/graph-ql/coverage/directory.graphqls
+++ b/design-documents/graph-ql/coverage/directory.graphqls
@@ -3,7 +3,7 @@
 
 type Query {
     currency: Currency @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Currency") @doc(description: "The currency query returns information about store currency.") @cache(cacheable: false)
-    countries: [Country] @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Countries") @doc(description: "The countries query provides information for all countries.") @cache(cacheable: false)
+    countries: [Country] @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Countries") @doc(description: "The countries query provides information for all countries. Sorted by Top Destination and locale specific country name.") @cache(cacheable: false)
     country (id: String): Country @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Country") @doc(description: "The countries query provides information for a single country.") @cache(cacheable: false)
 }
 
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/directory.graphqls
```diff
@@ -0,0 +1,48 @@
...
+    exchange_rates: [ExchangeRate]
+}
+
+type ExchangeRate {
+    currency_to: String
+    rate: Float
+}
+
+type Country {
+    id: String
```
paales, line 27: Should we also change this to a `CountryEnum` OR should `CountryEnum` be deprecated and all instances where `CountryEnum` is currently used be replaced by this `Country` type?

design-documents/graph-ql/coverage/directory.graphqls
```diff
@@ -0,0 +1,48 @@
...
+# See COPYING.txt for license details.
+
+type Query {
+    currency: Currency @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Currency") @doc(description: "The currency query returns information about store currency.") @cache(cacheable: false)
+    countries: [Country] @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Countries") @doc(description: "The countries query provides information for all countries. Sorted by Top Destination and locale specific country name.") @cache(cacheable: false)
+    country (id: String): Country @resolver(class: "Magento\\DirectoryGraphQl\\Model\\Resolver\\Country") @doc(description: "The countries query provides information for a single country.") @cache(cacheable: false)
+}
+
+type Currency {
+    base_currency_code: String
```
cpartica, line 11: there is a different issue with Currency
https://github.com/magento/architecture/pull/474/files
see what I proposed here
https://github.com/magento/architecture/pull/474/files#r556999960
I am looking at other APIs and Currency object is simple usually name, code and symbol

design-documents/graph-ql/coverage/directory.graphqls
```diff
@@ -0,0 +1,48 @@
...
+    currency_to: String
+    rate: Float
+}
+
+type Country {
+    id: String
+    two_letter_abbreviation: String @doc(description: "Country code according to ISO 3166-1 alpha-2")
+    three_letter_abbreviation: String @doc(description: "Country code according to ISO 3166-1 alpha-3") 
+    name: String @doc(description: "Name of the country in the locale of the current store") 
+    full_name_locale: String @deprecated(reason: "Use `name`")
```
cpartica, line 31: everywhere else we have `fullname` used in customers. Of course that's not quite right but would make this an outlier

design-documents/graph-ql/coverage/directory.graphqls
```diff
@@ -0,0 +1,48 @@
...
+type Country {
+    id: String
+    two_letter_abbreviation: String @doc(description: "Country code according to ISO 3166-1 alpha-2")
+    three_letter_abbreviation: String @doc(description: "Country code according to ISO 3166-1 alpha-3") 
+    name: String @doc(description: "Name of the country in the locale of the current store") 
+    full_name_locale: String @deprecated(reason: "Use `name`") 
+    full_name_english: String @deprecated(reason: "Use `name`")
+    region_required: Boolean @doc(description: "Is a the region for this country") 
+    postcode_required: Boolean @doc(description: "Is a postcode required for this country") 
+    eu: Boolean @doc(description: "Country is a member of the European Union")
```
cpartica, line 35: instead of having a specific region for EU, wouldn't it be better to have a "Zone" with value as EU or APAC EMEA, etc
There are other zones besides EU

design-documents/graph-ql/coverage/directory.graphqls
```diff
@@ -0,0 +1,48 @@
...
+
+type Region {
+    id: Int
+    code: String @doc(description: "Retuns the region code if a region has a code, else returns the non localized name") 
+    name: String @doc(description: "Locale specific region name.") 
+}
+
+type StoreConfig {
+    default_country: String @doc(description: "Default country used for country fields") 
+    allow_optional_state: String @doc(description: "Allow a customer to enter a state when it isn't required for a country")
```
cpartica, line 47: what is the path in system config for this?

cpartica (changes requested):

design-documents/graph-ql/coverage/directory.graphqls
```diff
@@ -0,0 +1,48 @@
...
+    rate: Float
+}
+
+type Country {
+    id: String
+    two_letter_abbreviation: String @doc(description: "Country code according to ISO 3166-1 alpha-2")
+    three_letter_abbreviation: String @doc(description: "Country code according to ISO 3166-1 alpha-3") 
+    name: String @doc(description: "Name of the country in the locale of the current store") 
+    full_name_locale: String @deprecated(reason: "Use `name`") 
+    full_name_english: String @deprecated(reason: "Use `name`")
```
paales, line 32: @cpartica Something like this?

```suggestion
    fullname: String @doc(description: "Name of the country in the locale of the current store") 
    full_name_locale: String @deprecated(reason: "Use `fullname`") 
    full_name_english: String @deprecated(reason: "Use `fullname`")
```

design-documents/graph-ql/coverage/directory.graphqls
paales: In the backend of Magento there is a specific configuration field that allows you to check EU countries. But I agree that zone would be better, but that would require modifications on the other layers as well.

design-documents/graph-ql/coverage/directory.graphqls
paales: Uhh, don't know the exact path, but it's this field:

Schermafbeelding 2021-01-14 om 12 43 37

design-documents/graph-ql/coverage/directory.graphqls
paales: Agreed that should be fixed, if there is a conclusion in the other PR I can update this one.

PR #466, opened 2020-11-12
Giftregistry changes uid

cpartica:
## Problem

Gift registry has to be changed to support uid.

## Solution

The schema doesn't reflect new uids on selected options, plus we need to update the proposal for the Gift registry to support same types of uids added to catalog and cart.


design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -240,11 +240,11 @@ type GiftRegistry {
 }
 
 interface GiftRegistryItemInterface {
-    id: String!
+    uid: String!
```
dthampy, line 243: Please fix the type to "ID" instead of "String"

dthampy (changes requested):

PR #468, opened 2020-11-12
Reward points Uid changes

RakeshJesadiya:
## Problem

Customer Reward Points UID Changes in the schema file.

## Solution

To keep consistency throughout the GraphQL schema, we have replaced the Id field with Uid.


RakeshJesadiya:
@cpartica @melnikovi Please check the schema changes. once approve I will do it changes on the code level.

cpartica (approved):

PR #469, opened 2020-11-12
Customer Newsletter subscription label changes.

RakeshJesadiya:
## Problem

Customer Newsletter Subscription label changes.

## Solution

Just add the Newsletter keyword before subscription. The label will be clear for the subscription to the newsletter.

### DIFF
--- a/design-documents/graph-ql/coverage/customer/SubscribeEmailToNewsletter.graphqls
+++ b/design-documents/graph-ql/coverage/customer/SubscribeEmailToNewsletter.graphqls
@@ -3,7 +3,7 @@ type Mutation {
 }
 
 type SubscribeEmailToNewsletterOutput {
-    status: SubscriptionStatusesEnum @doc(description: "Returns a status of subscription")
+    status: SubscriptionStatusesEnum @doc(description: "Returns a status of newsletter subscription")
 }
 
 enum SubscriptionStatusesEnum {


design-documents/graph-ql/coverage/customer/SubscribeEmailToNewsletter.graphqls
```diff
@@ -3,7 +3,7 @@ type Mutation {
 }
 
 type SubscribeEmailToNewsletterOutput {
-    status: SubscriptionStatusesEnum @doc(description: "Returns a status of subscription")
+    status: SubscriptionStatusesEnum @doc(description: "Returns a status of newsletter subscription")
 }
 
 enum SubscriptionStatusesEnum {
```
dthampy, line 9: To better align with the existing enum values,   change "NOT_ACTIVE"  status to INACTIVE

RakeshJesadiya:
@dthampy I have made changes based on your comment.

design-documents/graph-ql/coverage/customer/SubscribeEmailToNewsletter.graphqls
cpartica: yes but we can't remove that value
 we have to keep it, add a comment that's not used and add "INACTIVE" as a new value

cpartica (changes requested):

RakeshJesadiya:
@cpartica I have updated the value of ENUM type with deprecated note.

PR #474, opened 2021-01-04
currency symbol graphql coverage

dani97:
## Problem

Custom Currency Symbol graphql coverage needs to be added. 

## Solution

Created new types and fields in currency query

## Requested Reviewers

@cpartica

## Related Links
[Magento 2 repo feature request link](https://github.com/magento/magento2/issues/31494)
[Pwa studio original issue](https://github.com/magento/pwa-studio/issues/2491)

### DIFF
--- a/design-documents/graph-ql/coverage/store-currency-symbol.md
+++ b/design-documents/graph-ql/coverage/store-currency-symbol.md
@@ -5,19 +5,32 @@ Magento allows to use custom currency symbol.The current graphql implementation 
  
  ## Schema
 ```graphql
-type CurrencyData {
+type CurrencyObject {
     currency_code: String
     currency_symbol: String
 }
 
-type Currency {
-    available_currency_codes: [String] @deprecated
+type CurrencySettings {
     base_currency_code: String
     base_currency_symbol: String
     default_display_currency_code: String
     default_display_currency_symbol: String
     exchange_rates: [ExchangeRate]
-    available_currency: [CurrencyData!]!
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/store-currency-symbol.md
```diff
@@ -0,0 +1,24 @@
+# Coverage for currency symbol
+
+Magento allows to use custom currency symbol.The current graphql implementation does not fully support custom currency symbol feature. The currency query only returns base currency symbol and current store default currency symbol.
+ `available_currency_codes` should be deprecated and instead `available_currency` field should be added, or both should be kept
+ 
+ ## Schema
+```graphql
+type CurrencyData {
```
cpartica, line 8: this is ok but instead of using `CurrencyData` - the word `data` is ambiguous
I can't think of a better name than Currency but it's taken.
Maybe we want to rename existing type `Currency` to CurrencySettings  
currency: Currency deprecating this - 
add currencySettings: CurrencySettings

we can name CurrencyData to CurrencyObject then in the future rename it back to Currency {
    currency_code: String
    currency_symbol: String

}

cpartica (changes requested):

design-documents/graph-ql/coverage/store-currency-symbol.md
dani97: Thanks @cpartica . I have updated the schema

PR #480, opened 2021-03-22
[GQL gift registry] cart management mutations

cpartica:
## Problem
Luma has this operation to add all items from cart to gift registry. We don't have this in graphql.
Not only that we don't have it yet, but we have opened a lot of use cases by adding individual items to gift-registry, which we don't want to support just yet.

## Solution
Add specialized query to move items to cart to gift registry. Alternatively we could use this query to empty the cart while doing so or not.

### DIFF
--- a/design-documents/graph-ql/coverage/customer/gift-registry.graphqls
+++ b/design-documents/graph-ql/coverage/customer/gift-registry.graphqls
@@ -130,12 +130,12 @@ input GiftRegistryDynamicAttributeInput {
     value: String!
 }
 
-interface GiftRegistryUserErrors {
+interface GiftRegistryUserErrorsInterface {
     status: Boolean!
     user_errors: [GiftRegistryItemsUserError]!
 }
 
-interface GiftRegistryItemUserError {
+interface GiftRegistryItemUserErrorInterface {
     status: Boolean!
     user_errors: GiftRegistryItemsUserError
 }
@@ -154,10 +154,10 @@ type RemoveGiftRegistryOutput {
     is_removed: Boolean!
 }
 
-type AddGiftRegistryItemsOutput implements GiftRegistryOutputInterface, GiftRegistryItemUserError {
+type AddGiftRegistryItemsOutput implements GiftRegistryOutputInterface, GiftRegistryItemUserErrorInterface {
 }
 
-type MoveCartItemsToGiftRegistryOutput implements GiftRegistryOutputInterface, GiftRegistryUserErrors {
+type MoveCartItemsToGiftRegistryOutput implements GiftRegistryOutputInterface, GiftRegistryUserErrorsInterface {
 }
 
 type GiftRegistryItemsUserError {
@@ -173,10 +173,10 @@ enum GiftRegistryItemsUserErrorType {
     UNDEFINED @doc(description: "Used for other exceptions like db connection failures or others")
 }
 
-type RemoveGiftRegistryItemsOutput implements GiftRegistryOutputInterface, GiftRegistryItemUserError {
+type RemoveGiftRegistryItemsOutput implements GiftRegistryOutputInterface, GiftRegistryItemUserErrorInterface {
 }
 
-type UpdateGiftRegistryItemsOutput implements GiftRegistryOutputInterface, GiftRegistryItemUserError {
+type UpdateGiftRegistryItemsOutput implements GiftRegistryOutputInterface, GiftRegistryItemUserErrorInterface {
 }
 
 type AddGiftRegistryRegistrantsOutput implements GiftRegistryOutputInterface {
--- a/design-documents/graph-ql/coverage/customer/gift-registry.graphqls
+++ b/design-documents/graph-ql/coverage/customer/gift-registry.graphqls
@@ -26,7 +26,7 @@ type Mutation {
     updateGiftRegistry(giftRegistryUid: ID!, giftRegistry: UpdateGiftRegistryInput!): UpdateGiftRegistryOutput
     removeGiftRegistry(giftRegistryUid: ID!): RemoveGiftRegistryOutput
 
-    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput @deprecated(reason: "Use `moveCartItemsToGiftRegistry` instead. Not supported for gift registry") @doc(description: "Adds individual items to the gift registry")
+    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput @doc(description: "Adds individual items to the gift registry")
     moveCartItemsToGiftRegistry(cartUid: ID!, giftRegistryUid: ID!): MoveCartItemsToGiftRegistryOutput @doc(description: "Moves all items items from cart to the gift registry")
 
     removeGiftRegistryItems(giftRegistryUid: ID!, itemUids: [ID!]!): RemoveGiftRegistryItemsOutput
@@ -79,7 +79,7 @@ input EnteredOptionInput {
 }
 
 input UpdateGiftRegistryItemInput {
-    giftRegistryItemUid: ID!
+    gift_registry_item_uid: ID!
     quantity: Float!
     note: String
 }
@@ -94,10 +94,9 @@ input UpdateGiftRegistryInput {
 }
 
 input CreateGiftRegistryInput {
-    giftRegistryUid: ID @deprecated(reason: "Use `gift_registry_uid` instead") @doc(description: "Optional uid, can be generated on the client and used for sending multiple gift-registry related mutations in a single request. For example, create registry and immediatly add items or registrants.")
-    gift_registry_uid: ID
+    gift_registry_uid: ID @doc(description: "Optional uid, can be generated on the client and used for sending multiple gift-registry related mutations in a single request. For example, create registry and immediatly add items or registrants.")
     event_name: String!
-    giftRegistryTypeUid: ID!
+    gift_registry_type_uid: ID!
     message: String!
     privacy_settings: GiftRegistryPrivacySettings!
     status: GiftRegistryStatus!
@@ -112,7 +111,6 @@ input GiftRegistryShippingAddressInput @doc(description: "Either address data or
 }
 
 input UpdateGiftRegistryRegistrantInput {
-    giftRegistryRegistrantUid: ID! @deprecated(reason: "Use `gift_registry_registrant_uid` instead")
     gift_registry_registrant_uid: ID
     first_name: String
     last_name: String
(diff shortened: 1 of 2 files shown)


design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -9,7 +9,7 @@ type Query {
         email: String! @doc(description: "The registrant's email")
     ): [GiftRegistrySearchResult] @doc(description: "Search for gift registries by specifying a registrant email address")
     giftRegistryIdSearch(
-        giftRegistryUid: String! @doc(description: "The ID of the gift registry")
+        giftRegistryUid: ID! @doc(description: "The ID of the gift registry")
```
cpartica, line 12: this was delivered in 2.4.2

larsroettig:
Looks nice will check soon is okay if I reviewed it?

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -129,26 +134,55 @@ input GiftRegistryDynamicAttributeInput {
 }
 
 type CreateGiftRegistryOutput {
+    status: Boolean!
+    create_gift_registry_user_errors: [GiftRegistryItemsUserError]!
```
eduard13, line 138: What about having some general UserError type, and move the `status` and `user_errors` into a new OutputInterface, so each Output type is going to implement this interface?

```suggestion
    user_errors: [GiftRegistryItemsUserError]!
```

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -26,7 +26,9 @@ type Mutation {
     updateGiftRegistry(giftRegistryUid: ID!, giftRegistry: UpdateGiftRegistryInput!): UpdateGiftRegistryOutput
     removeGiftRegistry(giftRegistryUid: ID!): RemoveGiftRegistryOutput
 
-    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput
+    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput @deprecated(reason: "Use `moveCartItemsToGiftRegistry` instead. Not supported for gift registry") @doc(description: "Adds individual items to the gift registry")
```
eduard13, line 29: @cpartica could you please clarify why this mutation has been deprecated, and suggests to use the `moveCartItemsToGiftRegistry`, since the purpose of these 2 mutations are a bit different?

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
cpartica: > What about having some general UserError type, and move the `status` and `user_errors` into a new OutputInterface, so each Output type is going to implement this interface?
So I did, except there's a difference between gift_registry_user_errors and gift_registry_item_user_errors
One is an array and the the one is not. Check it out on the latest version

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,68 @@ input GiftRegistryDynamicAttributeInput {
     value: String!
 }
 
-type CreateGiftRegistryOutput {
-    gift_registry: GiftRegistry
+interface GiftRegistryUserErrors {
+    status: Boolean!
+    gift_registry_user_errors: [GiftRegistryItemsUserError]!
```
cpartica, line 135: @eduard13 check this out https://github.com/magento/architecture/pull/480#discussion_r618590876

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,62 @@ input GiftRegistryDynamicAttributeInput {
     value: String!
 }
 
-type CreateGiftRegistryOutput {
-    gift_registry: GiftRegistry
+interface GiftRegistryUserErrors {
+    status: Boolean!
+    user_errors: [GiftRegistryItemsUserError]!
```
cpartica, line 135: @eduard13 added change on your request https://github.com/magento/architecture/pull/480#discussion_r618590876

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -26,7 +26,9 @@ type Mutation {
     updateGiftRegistry(giftRegistryUid: ID!, giftRegistry: UpdateGiftRegistryInput!): UpdateGiftRegistryOutput
     removeGiftRegistry(giftRegistryUid: ID!): RemoveGiftRegistryOutput
 
-    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput
+    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput @doc(description: "Adds individual items to the gift registry")
```
cpartica, line 29: @eduard13 please refer to https://github.com/magento/architecture/pull/480#discussion_r618556304

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,63 @@ input GiftRegistryDynamicAttributeInput {
...
-    gift_registry: GiftRegistry
+type MoveCartItemsToGiftRegistryOutput implements GiftRegistryOutputInterface, GiftRegistryUserErrorsInterface {
 }
 
-type UpdateGiftRegistryItemsOutput {
-    gift_registry: GiftRegistry
+type GiftRegistryItemsUserError {
+    message: String!
+    product_uid: ID
+    gift_registry_item_uid: ID
```
cpartica, line 166: gift_registry_item_uid and gift_registry_uid are optional - in case of MoveCartItemsToGiftRegistry we might not have either
product_uid is also optional because we might refer to a product that doesn't exist anymore

In case we have a valid reference these should be populated

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
eduard13: @cpartica could you please remove the mutation from GQL schema? I've opened a PR that removes its implementation, and would be good to have the same changes here.
Thank you.

design-documents/graph-ql/directives.graphqls
```diff
@@ -0,0 +1,39 @@
+directive @doc(description: String="") on QUERY
+    | MUTATION
```
cpartica, line 2: added  this because it was missing from this repo - not related to GR

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -26,7 +26,9 @@ type Mutation {
     updateGiftRegistry(giftRegistryUid: ID!, giftRegistry: UpdateGiftRegistryInput!): UpdateGiftRegistryOutput
     removeGiftRegistry(giftRegistryUid: ID!): RemoveGiftRegistryOutput
 
-    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput
+    addGiftRegistryItems(giftRegistryUid: ID!, items: [AddGiftRegistryItemInput!]!): AddGiftRegistryItemsOutput @doc(description: "Adds individual items to the gift registry")
+    moveCartItemsToGiftRegistry(cartUid: ID!, giftRegistryUid: ID!): MoveCartItemsToGiftRegistryOutput @doc(description: "Moves all items items from cart to the gift registry")
```
pdohogne-magento, line 30: Extra "items" in doc comment.

```suggestion
    moveCartItemsToGiftRegistry(cartUid: ID!, giftRegistryUid: ID!): MoveCartItemsToGiftRegistryOutput @doc(description: "Moves all items from cart to the gift registry")
```

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,63 @@ input GiftRegistryDynamicAttributeInput {
...
-type CreateGiftRegistryOutput {
-    gift_registry: GiftRegistry
+interface GiftRegistryUserErrorsInterface {
+    status: Boolean!
+    user_errors: [GiftRegistryItemsUserError]!
+}
+
+interface GiftRegistryItemUserErrorInterface {
+    status: Boolean!
+    user_errors: GiftRegistryItemsUserError
```
pdohogne-magento, line 140: I don't like having a `GiftRegistryItemsUserError` type that doesn't implement `GiftRegistryItemUserErrorInterface`, that's a bad naming convention.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
pdohogne-magento: Will they never be populated on `MoveCartItemsToGiftRegistry` at all?

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
pdohogne-magento: I would rename `GiftRegistryItemsUserError` to `GiftRegistryItemError`; not all error types are due to user input (such as DB connection failure).

pdohogne-magento: Similarly I would rename the `user_errors` field to just `errors`.  Also it should be singular in this interface as we only return one.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,63 @@ input GiftRegistryDynamicAttributeInput {
     value: String!
 }
 
-type CreateGiftRegistryOutput {
-    gift_registry: GiftRegistry
+interface GiftRegistryUserErrorsInterface {
+    status: Boolean!
```
pdohogne-magento, line 134: Needs to be the three-state enum we discussed elsewhere to allow for partial success.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,63 @@ input GiftRegistryDynamicAttributeInput {
...
+    user_errors: [GiftRegistryItemsUserError]!
+}
+
+interface GiftRegistryItemUserErrorInterface {
```
pdohogne-magento, line 138: Same comment about removing "User" from the name.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,63 @@ input GiftRegistryDynamicAttributeInput {
     value: String!
 }
 
-type CreateGiftRegistryOutput {
-    gift_registry: GiftRegistry
+interface GiftRegistryUserErrorsInterface {
```
pdohogne-magento, line 133: Same comment about removing "User" from the name.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,63 @@ input GiftRegistryDynamicAttributeInput {
...
+    message: String!
+    product_uid: ID
+    gift_registry_item_uid: ID
+    gift_registry_uid: ID
+    code: GiftRegistryItemsUserErrorType!
 }
 
-type AddGiftRegistryRegistrantsOutput {
-    gift_registry: GiftRegistry
+enum GiftRegistryItemsUserErrorType {
```
pdohogne-magento, line 171: Same comment about removing "User" from the name.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
cpartica: in this case no, because if one fails, no GR will be created.
there is no partial error.
I will remove status though.

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
cpartica: it's actually aligned with the rest of the schema:
https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer/Wishlist.graphqls#L84

I don't disagree that the "User" naming is redundant

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
cpartica: I added a comment about, we would have to change it everywhere else but we also have to start somewhere

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
cpartica: > I don't like having a `GiftRegistryItemsUserError` type that doesn't implement the `GiftRegistryItemUserErrorInterface` interface that we also have, that's a bad naming convention.

Yes that's a good point, it's a lower hierarchy. Would only apply to the same level or hierarchy if GiftRegistryItemsUserError type that doesn't implement the GiftRegistryItemUserErrorInterface

The term User doesn't come because the user caused an error. It comes from the fact that we have some error to output to the user. However the user here is PWA
And if we look what's inside, information is mixed, some of the info like code:OUT_OF_STOCK has to be processed to be output.

```
type GiftRegistryItemsUserError {
    message: String! - output for the user
    product_uid: ID - not output for the user
    gift_registry_item_uid: ID not output for the user
    gift_registry_uid: ID not output for the user
    code: GiftRegistryItemsUserErrorType! - output for the user but after translated by PWA
  }

```

design-documents/graph-ql/coverage/customer/gift-registry.graphqls
```diff
@@ -128,40 +130,61 @@ input GiftRegistryDynamicAttributeInput {
...
 }
 
-type CreateGiftRegistryOutput {
-    gift_registry: GiftRegistry
+interface GiftRegistryErrorsInterface {
+    errors: [GiftRegistryItemError]!
+}
+
+interface GiftRegistryItemErrorInterface {
+    error: GiftRegistryItemError
```
cpartica, line 138: @pdohogne-magento I addressed your concerns, but this one
I do like grouping fields contained in GiftRegistryItemError under a type on a different node 'error'
in that case GiftRegistryItemError implements GiftRegistryItemErrorInterface could happen but doesn't have a lot of sense unless we use that on the entire schema in a more general way:

`GiftRegistryItemError implements ItemErrorInterface`

as for the GiftRegistryItemErrorInterface I suggest we find another name for it?

PR #481, opened 2021-03-25
X vary cache proposal

cpartica:
## Problem
Vary header exposes the wrong data in GraphQL and poses a problem with dish cache in browser.
We also need to add private data part of the cache key like Customer-Group 

## Solution
Fix the Vary header output in GraphQL
Add `X-Magento-Request-Id` on output as a replacement for X-Magento-Vary cookie and fix VCL

### DIFF
--- a/design-documents/graph-ql/improved-caching.md
+++ b/design-documents/graph-ql/improved-caching.md
@@ -40,7 +40,7 @@ Vary: Accept-Encoding, Cookie
 
 All components/headers that compose cache key should be included in the Vary.
 
-## Solution
+## The Solution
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/improved-caching.md
```diff
@@ -0,0 +1,67 @@
...
+````
+Vary: Accept-Encoding, Cookie
+````
+
+All components/headers that compose cache key should be included in the Vary.
+
+## The Solution
+Graphql would compute the cache key and return a header with the same function as the Luma cookie (`X-Magento-Vary`), except we won't name it that way. The `Vary` header is not meant to have a hash.
+````
+X-Magento-Request-Id: 85a0b196524c60eaeb7c87d1aa4708a3fb20c6a1
```
ihor-sviziev, line 46: Please don't name it as "request-id" as it's not an identifier of the request.

PR #482, opened 2021-03-25
Add wishlists in products query as a graph

cpartica:
## Problem

Products query doesn't contain reference to wishlist, it takes a round trip to render a PDP page dropdown to what wishlists are assgined to the product

## Solution
add the reference in schema

### DIFF
--- a/design-documents/graph-ql/coverage/customer/Wishlist.graphqls
+++ b/design-documents/graph-ql/coverage/customer/Wishlist.graphqls
@@ -159,7 +159,7 @@ type CreateWishlistOutput {
 
 type DeleteWishlistOutput {
     status: Boolean!
-    wishlists: [Wishlist!]!
+    wishlists: [Wishlist!]! @doc(description: "A customer will always have at least one wishlist")
 }
 
 input WishlistItemCopyInput {
@@ -189,4 +189,9 @@ type MoveProductsBetweenWishlistsOutput {
 type StoreConfig {
     maximum_number_of_wishlists: Int @doc(description: "If multiple wish lists are enabled, the maximum number of wish lists the customer can have")
     enable_multiple_wishlists: Boolean @doc(description: "Indicates whether customers can have multiple wish lists.")
+}
+
+#Allow Products to show assigned wishlists
+type ProductInterface {
+    wishlists: [Wishlist]! @doc(description: "A product can be assigned to multiple wishlist of none")
 }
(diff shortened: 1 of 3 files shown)


supernova-at:
Thanks for this!

Yes, https://jira.corp.magento.com/browse/PWA-1268 calls for the following UI:

Screen Shot 2021-03-25 at 11 02 53 AM

Notice the heart indicators for whether or not a product has already been favorited.

This PR will help the storefront know immediately how to display that heart / favorite indicator instead of having to do a bunch of custom client-side logic and additional wishlist lookups.

design-documents/graph-ql/coverage/customer/Wishlist.graphqls
```diff
@@ -190,3 +190,8 @@ type StoreConfig {
     maximum_number_of_wishlists: Int @doc(description: "If multiple wish lists are enabled, the maximum number of wish lists the customer can have")
     enable_multiple_wishlists: Boolean @doc(description: "Indicates whether customers can have multiple wish lists.")
 }
+
+#Allow Products to show assigned wishlists
+type ProductInterface {
+    wishlists: [Wishlist]! @doc(description: "A product can be assigned to multiple wishlist of none")
```
supernova-at, line 196:
```suggestion
    wishlists: [Wishlist]! @doc(description: "A product can be assigned to multiple wishlist or none")
```

design-documents/graph-ql/coverage/customer/wishlist.md
```diff
@@ -111,3 +111,35 @@ type WishlistItem {
...
+``` graphql
+type ProductInterface {
+    wishlists: [Wishlist]! @doc(description: "A product can be assigned to multiple wishlist of none")
+}
+```
+
+By default a product is not assigned to any wishlist
+
+### Considerations of performance versus graphql specs
+The UI really needs this to render a PDP page or Category with products page and list a dropdown of wishlists of which the product is assigned to.
```
cpartica, line 128: @supernova-at how do you feel about this? we would lean to not expose the whole wishlist type and create a loop

supernova-at: Yes, I'd lean that way too. I suppose I could envision a scenario where the UI would want to show the names of the wishlists that the product belonged to, but I think it's perfectly reasonable to have the front end make an additional query for the full wishlist data.

As long as the minimal type would give us enough information to look up the full wishlist, I think that's fine 👍 .

For example, type gives `wishlistID` and then there's a way to get wishlists by `wishlistID`.

PR #483, opened 2021-03-29
Adjust neg. quote operations to contain sort and status

cpartica:
## Problem

Mutations in schema is not align with statuses on operations that might have partial or total success, which is an overall effort. 
We also can't sort the quotes returned. 

## Solution
To align mutations schema with operations that might have partial or total success, we need to add statuses.
Also we can add sorting on quotes solving the sorting problems

### DIFF
--- a/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
@@ -14,7 +14,7 @@ type Query {
 }
 
 type NegotiableQuotesOutput {
-    items: [NegotiableQuote]
+    items: [NegotiableQuote!]!
     page_info: SearchResultPageInfo!
     total_count: Int!
 }
@@ -97,7 +97,6 @@ input DeleteNegotiableQuotesInput {
 }
 
 type DeleteNegotiableQuotesOutput {
-    status: Boolean!
     # Implementation Note: We don't make the deleted quotes accessible
     # because "deleted" means they're hidden from the storefront UI. They will
     # still be visible (and labeled as deleted) for a Seller in the admin
@@ -114,7 +113,6 @@ input CloseNegotiableQuotesInput {
 }
 
 type CloseNegotiableQuotesOutput {
-    status: Boolean!
     closed_quotes(
         filter: NegotiableQuoteFilterInput,
         pageSize: Int = 20,
--- a/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
@@ -9,6 +9,7 @@ type Query {
         filter: NegotiableQuoteFilterInput,
         pageSize: Int = 20,
         currentPage: Int = 1
+        sort: NegotiableQuoteSortInput
     ): NegotiableQuotesOutput @doc(description: "A (optionally filtered) list of negotiable quotes viewable by the logged-in customer")
 }
 
@@ -96,6 +97,7 @@ input DeleteNegotiableQuotesInput {
 }
 
 type DeleteNegotiableQuotesOutput {
+    status: Boolean!
     # Implementation Note: We don't make the deleted quotes accessible
     # because "deleted" means they're hidden from the storefront UI. They will
     # still be visible (and labeled as deleted) for a Seller in the admin
@@ -103,6 +105,7 @@ type DeleteNegotiableQuotesOutput {
         filter: NegotiableQuoteFilterInput,
         pageSize: Int = 20,
         currentPage: Int = 1
+        sort: NegotiableQuoteSortInput
     ): NegotiableQuotesOutput @doc(description: "List of negotiable quotes available to customer")
 }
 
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -111,12 +114,19 @@ input CloseNegotiableQuotesInput {
 }
 
 type CloseNegotiableQuotesOutput {
-    closed_quotes: [NegotiableQuote!] @doc(description: "Quotes that were just closed")
+    status: Boolean!
+    closed_quotes(
```
cpartica, line 118: [NegotiableQuote!] - would need NegotiableQuotesOutput { items: [NegotiableQuote!]}

while NegotiableQuotesOutput currently has  items: [NegotiableQuote]!

Should we add a specific type for both? or modify NegotiableQuotesOutput to have
 items: [NegotiableQuote] - basically satisfying both cases but leaving schema less specific

amolina-magento: For reference regarding nullability of array members: https://www.apollographql.com/blog/using-nullability-in-graphql-2254f84c4ed7

```
drinkSizes: [String!]

This means that the list itself can be null, but it can’t have any null members. For example, in JSON:

drinkSizes: null // valid
drinkSizes: [] // valid
drinkSizes: ["small", "large"] // valid
drinkSizes: ["small", null, "large"] // error
```

cpartica: so it seems that we should go for `items: [NegotiableQuote!]!`

prabhuram93: Adding more specificity is risky as it tends to be the least BIC. I would suggest items: [NegotiableQuote] for now so it can be consistent with the rest of the schema. We can always add more specificity later.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -96,13 +97,15 @@ input DeleteNegotiableQuotesInput {
 }
 
 type DeleteNegotiableQuotesOutput {
+    status: Boolean!
```
pdohogne-magento, line 100: According to discussion with the PWA team, this field should be changed to an enum with (at least) three states: success, failure, and partial success.  Also rename to something more descriptive, possibly `resultState`?

pdohogne-magento:
The result status field should be added to the output for all mutations.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
pdohogne-magento: Related: https://jira.corp.magento.com/browse/PWA-1523 will add expanded error information to the output and allow callers to easily grab the list of failing ids when there's a partial success.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -199,6 +209,17 @@ input NegotiableQuoteFilterInput {
     name: FilterMatchTypeInput @doc(description: "Filter by negotiable quote name")
 }
 
+input NegotiableQuoteSortInput {
+    name: SortEnum @doc(description: "Sort by negotiable quote name")
```
prabhuram93, line 213: This looks good. Additiona note, Luma doesn't support sorting at this point. 
Screen Shot 2021-04-09 at 1 45 31 PM

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
prabhuram93: @cpartica since we are not tackling status now, can you remove the status fields from this PR. Rest looks good.

prabhuram93 (approved):

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -9,11 +9,12 @@ type Query {
         filter: NegotiableQuoteFilterInput,
         pageSize: Int = 20,
         currentPage: Int = 1
+        sort: NegotiableQuoteSortInput
     ): NegotiableQuotesOutput @doc(description: "A (optionally filtered) list of negotiable quotes viewable by the logged-in customer")
 }
 
 type NegotiableQuotesOutput {
-    items: [NegotiableQuote]!
+    items: [NegotiableQuote!]!
```
cpartica, line 17: @amolina-magento added this non nullable like you suggested

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
cpartica: done

prabhuram93 (approved):

amolina-magento (approved):

prabhuram93 (approved):

PR #495, opened 2021-09-15
Added shipping address property to NegotiableQuote object

amolina-magento:
## Problem

**Problem 1:** In GraphQL, the shipping addresses are not returned as part of the Negotiable Quote object. Thus, the current state of a Negotiable Quote's shipping addresses cannot be determined. As a result, errors will be unknown to the user until they attempt to place an order, leading to a poor user experience.

**Problem 2:** In GraphQL, when setting a shipping address on a Negotiable Quote, we currently support only existing customer addresses from an address book. In order to maintain parity with Luma, we should allow the user to supply a new address for the Negotiable Quote.

## Solution

Update the current schema so that the NegotiableQuote object includes its shipping addresses
Update the current schema so that the SetNegotiableQuoteShippingAddressInput object supports a new shipping address object


design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -76,9 +76,24 @@ type AddNegotiableQuoteItemsOutput {
...
 input SetNegotiableQuoteShippingAddressInput {
     quote_uid: ID! @doc(description: "ID obtained from NegotiableQuote type")
-    customer_address_id: Int! @doc(
+    customer_address_id: Int @doc(
         description: "ID obtained from the CustomerAddress type. A new address can be added using Mutation.createCustomerAddress"
     )
+    address: NegotiableQuoteAddressInput
+}
+
+input NegotiableQuoteAddressInput {
```
pdohogne-magento, line 85: I would suggest making a `QuoteAddressInputInterface` in QuoteGraphQL with these fields and have both `CartAddressInput` and `NegotiableQuoteAddressInput` implement it.

amolina-magento: I thought we were holding off on all unification via interfaces for now? Would it not be inconsistent to do that for the CartAddressInput and not the Cart itself? 

Also, in all likelihood, this wouldn't cause any issues with any existing CartAddress functionality. However, isn't there an added element of risk?

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -190,6 +205,7 @@ type NegotiableQuote {
     created_at: String @doc(description: "Timestamp indicating when the negotiable quote was created.")
     updated_at: String @doc(description: "Timestamp indicating when the negotiable quote was updated.")
     status: NegotiableQuoteStatus!
+    shipping_addresses: [ShippingCartAddress!]!
```
pdohogne-magento, line 208: I'd do the same thing as above with an interface and reorder the naming to `QuoteShippingAddress` on the new types (rather than `ShippingQuoteAddress`).

amolina-magento: But leave the name `ShippingCartAddress` as is, right?

pdohogne-magento: Also, why is `ShippingCartAddress` required here?

amolina-magento: The outer exclamation point indicates that the value cannot be null. In this case we expect, at the very least, an empty array. 

The inner exclamation point indicates that if the array is not empty, every element in the array must be of type `ShippingCartAddress`.

pdohogne-magento: We don't have the inner exclamation in the QuoteGraphQl schema's field, though.

pdohogne-magento: > I'd do the same thing as above with an interface and reorder the naming to `QuoteShippingAddress` on the new types (rather than `ShippingQuoteAddress`).

We should actually think on this some more.  Having a different name for this one would mean anything that extends `ShippingCartAddress` would not be applied to `QuoteShippingAddress` even though they implement the same interface.



> But leave the name `ShippingCartAddress` as is, right?

Yes, we have to leave the existing names in place for BiC.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
pdohogne-magento: It's not inconsistent.  It's not changing anything about what shows up on `CartAddressInput`, but it lets us use the same backing type for Negotiable Quotes.  Everything in the schema shows up and is treated exactly the same, just like moving a function to a parent class in code.  There's no risk here, it's not renaming anything that's already there or changing where it's already accessible or how it's extended.

We're holding off on unifying the existing schema because we need to investigate the level of effort involved, not because it isn't the right thing to do.  This is a new addition to the schema; not doing it with a unified interface now just means that much more needs to happen in the future.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
amolina-magento: > We don't have the inner exclamation in the QuoteGraphQl schema's field, though.

I'm aware of this. My logic was that the NegotiableQuote schema already deviated from the Cart (specifically the items). It seemed appropriate to structure the NegotiableQuote shipping_addresses property as it _should_ be. It's also easy to go from more restrictive to less restrictive.

However, I'm not steadfast on this. For the purpose of unifying these objects under a common interface, it might be better to define them as closely as possible.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
pdohogne-magento: It also doesn't currently make a lot of sense to do a unified interface between `Cart` and `NegotiableQuote`.  The only field they share with an identical definition is `prices`.

amolina-magento: > The only field they share with an identical definition is prices.

There's also `items`. In general though, I think the unified interface was in anticipation for checkout where there will be several more common fields. Still, it's better to hold off.

amolina-magento: > There's no risk here, it's not renaming anything that's already there or changing where it's already accessible or how it's extended.
As long as there's no risk, then I'm in favor of just a `QuoteAddressInputInterface`. What are your thoughts on a `NegotiableQuoteAddressInput` though?

pdohogne-magento: I prefer `QuoteAddressInput` myself as ideally that's what it should have been in both places, but since we have to have them separate I'm fine with `NegotiableQuoteAddressInput` to match the naming in the rest of this module.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -190,6 +208,7 @@ type NegotiableQuote {
     created_at: String @doc(description: "Timestamp indicating when the negotiable quote was created.")
     updated_at: String @doc(description: "Timestamp indicating when the negotiable quote was updated.")
     status: NegotiableQuoteStatus!
+    shipping_addresses: [ShippingCartAddress]!
```
amolina-magento, line 211: Need help deciding what to do here. The structure for ShippingCartAddress in the QuoteGraphQL module is as follows. Do we need to duplicate _each_ of them for NegotiableQuote?

` CartAddressInterface` -> `ShippingCartAddress` -> `CartAddressRegion`/`CartAddressCountry`

pdohogne-magento: Yes

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -223,6 +229,9 @@ type NegotiableQuote {
     status: NegotiableQuoteStatus!
 }
 
+type NegotiableQuoteShippingAddress implements NegotiableQuoteAddressInterface {
```
amolina-magento, line 232: Shipping method properties to be added as part of [PWA-2115](https://jira.corp.magento.com/browse/PWA-2115)

cspruiell (approved):

PR #496, opened 2021-09-17
PWA-2114: Create mutation to set payment method on negotiable quote

cspruiell:
## Problem

We cannot currently support end-to-end checkout of negotiable quotes in GraphQL because several mutations are missing, including one to set the payment method on the negotiable quote. (Other missing mutations will be addressed separately.)

## Solution

Add the setNegotiableQuotePaymentMethod mutation to the NegotiableQuoteGraphQl schema and return the full negotiable quote as the output.


dthampy:
The PR looks good @cspruiell

amolina-magento (approved):

amolina-magento (approved):

PR #502, opened 2021-10-27
Updated placeOrder mutation for negotiable quote usage

amolina-magento:
## Problem

The existing placeOrder mutation works for Cart, but does not account for Negotiable Quote usage.

## Solution

Override the mutation to support Negotiable Quotes.

## Requested Reviewers

@cpartica

### DIFF
--- a/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
@@ -73,7 +73,7 @@ type Mutation {
 
     placeNegotiableQuoteOrder(
         input: PlaceNegotiableQuoteOrderInput
-    ): PlaceOrderOutput @doc(description: "Place an order using the negotiable quote")
+    ): PlaceNegotiableQuoteOrderOutput @doc(description: "Place an order using the negotiable quote")
 
     # Pending decision on design of file upload (design doc still pending decisions)
     # addNegotiableQuoteFiles(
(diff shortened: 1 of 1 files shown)


nmarquez2:
@amolina-magento I don't think we should override the placeOrder mutation itself, we could extend the existing PlaceOrderInput type to make cart_id optional and add a quote_uid optional parameter so the mutation parameters can be consistent with other Negotiable Quote Mutations. 

After that we can do a simple plugin on the placeOrder resolver to perform Negotiable Quote validations and ensure either cart_id or quote_uid is provided and modify the input accordingly to allow the original method to run normally.

nmarquez2:
After internal discussions we concluded that we need to have a precise separate mutation to place an order from a negotiable quote to retain the clear distinction between quote types available actions. The PR was adjusted to reflect that by adding the placeNegotiableQuoteOrder mutation with a PlaceNegotiableQuoteOrderInput type that keeps the input consistent with other NQ mutations.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -71,6 +71,10 @@ type Mutation {
         input: SetNegotiableQuotePaymentMethodInput!
     ): SetNegotiableQuotePaymentMethodOutput @doc(description: "Set the payment method on the negotiable quote")
 
+    placeNegotiableQuoteOrder(
+        input: PlaceNegotiableQuoteOrderInput
+    ): PlaceOrderOutput @doc(description: "Place an order using the negotiable quote")
```
pdohogne-magento, line 76: This should have a NQ-specific output class as well.

nmarquez2: Schema has been updated

keharper (approved):

prabhuram93 (approved):

pdohogne-magento (approved):

PR #503, opened 2021-11-02
[PWA-2062] reCAPTCHA for PWA

tjwiebell:
## Problem

Would like to support reCAPTCHA v3 Invisible on PWA storefronts.

## Solution

Add reCAPTCHA configuration to `storeConfig` response in our core extension, magento-commerce/magento2-pwa.

### DIFF
--- a/design-documents/graph-ql/coverage/re-captcha.graphqls
+++ b/design-documents/graph-ql/coverage/re-captcha.graphqls
@@ -1,0 +1,31 @@
+enum ReCaptchaVersion {
+  V3
+}
+
+enum ReCaptchaForms {
+  APPLY_GIFT_CARD
+  APPLY_COUPON
+  CHANGE_PASSWORD
+  CHECKOUT
+  LOGIN
+  CREATE_CUSTOMER
+  EDIT_CUSTOMER
+  NEWSLETTER
+}
+
+type ReCaptchaConfiguration {
+  type: ReCaptchaVersion!
+  website_key: String!
+  minimum_score: Float!
+  badge_position: String!
+  language_code: String
+  failure_message: String!
+  forms: [ReCaptchaForms!]!
+}
+
+# Google reCAPTCHA config - will return null if v3 invisible is not configured
+# for at least one Storefront form.
+type StoreConfig {
+  recaptcha: ReCaptchaConfiguration
+    @doc(description: "Google reCAPTCHA Configuration")
+}
--- a/design-documents/graph-ql/coverage/re-captcha.graphqls
+++ b/design-documents/graph-ql/coverage/re-captcha.graphqls
@@ -1,31 +1,44 @@
-enum ReCaptchaVersion {
-  V3
-}
-
-enum ReCaptchaForms {
-  APPLY_GIFT_CARD
-  APPLY_COUPON
-  CHANGE_PASSWORD
-  CHECKOUT
-  LOGIN
-  CREATE_CUSTOMER
-  EDIT_CUSTOMER
+enum ReCaptchaFormEnum {
+  PLACE_ORDER
+  CONTACT
+  CUSTOMER_LOGIN
+  CUSTOMER_FORGOT_PASSWORD
+  CUSTOMER_CREATE
+  CUSTOMER_EDIT
   NEWSLETTER
+  PRODUCT_REVIEW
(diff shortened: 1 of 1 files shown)


design-documents/graph-ql/coverage/re-captcha.graphqls
```diff
@@ -0,0 +1,31 @@
...
+  NEWSLETTER
+}
+
+type ReCaptchaConfiguration {
+  type: ReCaptchaVersion!
+  website_key: String!
+  minimum_score: Float!
+  badge_position: String!
+  language_code: String
+  failure_message: String!
```
pdohogne-magento, line 22: Why is this required?  What if there's no failure?

design-documents/graph-ql/coverage/re-captcha.graphqls
```diff
@@ -0,0 +1,31 @@
+enum ReCaptchaVersion {
+  V3
```
pdohogne-magento, line 2: Why have an enum with only one value?

design-documents/graph-ql/coverage/re-captcha.graphqls
pdohogne-magento: Or is this the failure message that's configured to display if a captcha is failed, not if something goes wrong with the config retrieval?

tjwiebell: Following [user guide docs](https://docs.magento.com/user-guide/configuration/security/google-recaptcha-storefront.html?itm_source=merchdocs&itm_medium=quick_search&itm_campaign=federated_search&itm_term=recaptch#recaptcha-v3-invisible) this appeared non-nullable to me.

pdohogne-magento: Shouldn't language code be required as well then?

design-documents/graph-ql/coverage/re-captcha.graphqls
tjwiebell: Attempt to be forward thinking, if at some point we add support for V2 Invisible or if a V4 ever comes out. Not needed at all, but I didn't think a generic looking `recaptcha` field made sense just for V3.

pdohogne-magento: If 3rd-party versioning is a concern, I would consider putting a version label in the type name instead of an interior enum, as values that are currently required may not be available in future versions.  But that might be another discussion I'd want someone else's input on as well

design-documents/graph-ql/coverage/re-captcha.graphqls
tjwiebell: >Or is this the failure message that's configured to display if a captcha is failed, not if something goes wrong with the config retrieval?

Yep, exactly. If their score is under the threshold, we will prevent further submissions of the form and display this message.

>Shouldn't language code be required as well then?

That one should be safe as nullable; default behavior is to display the badge using the browser detected language, but you can optionally pass this override if you want.

design-documents/graph-ql/coverage/re-captcha.graphqls
tjwiebell: Sure, how do I go about facilitating that conversation? I don't care too much about this enum, so if there's an edit you can suggest to get this approved, I'm fine proceeding that way.

pdohogne-magento (approved):

keharper (approved):

eug123 (approved):

abrhim (approved):

design-documents/graph-ql/coverage/re-captcha.graphqls
```diff
@@ -0,0 +1,44 @@
...
+      description: "The message that appears to the user if validation fails"
+    )
+  forms: [ReCaptchaFormEnum!]!
+    @doc(description: "A list of forms that have reCAPTCHA V3 enabled")
+}
+
+# Google reCAPTCHA config - will return null if v3 invisible is not configured
+# for at least one Storefront form.
+type Query {
+  recaptchaV3Config: ReCaptchaConfigurationV3
```
cpartica, line 42: the only red flag that I see here is that this is a config query with no parms that could be a candidate for storeconfig

PR #505, opened 2021-11-09
PWA-1611: Adding the ability to report partial successes

pdohogne-magento:
PWA-1611: Adding the ability to report partial successes on batch operations in Negotiable Quote

## Problem
Some batch mutations in the Negotiable Quote module can have partial successes where some of the items succeed and some fail.  This needs to be reported in a usable way.

## Solution
Add a tri-state result status field to these operations with a `SUCCESS`, `FAILURE`, and `PARTIAL_SUCCESS` enum, and include an array result field that reports either a success or a typed failure for each item specified in the batch.

### DIFF
--- a/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
@@ -77,6 +77,16 @@ type Mutation {
     # ): AddNegotiableQuoteFilesInput
 }
 
+interface NegotiableQuoteUidNonFatalResultInterface {
+    quote_uid: ID!
+}
+
+type NegotiableQuoteUidOperationSuccess implements NegotiableQuoteUidNonFatalResultInterface {
+}
+
+type NegotiableQuoteInvalidStateError implements ErrorInterface {
+}
+
 type AddNegotiableQuoteItemsOutput {
     quote: NegotiableQuote
     # TODO: We'll probably want to add the same
@@ -157,29 +167,18 @@ input DeleteNegotiableQuotesInput {
     quote_uids: [ID!]! @doc(description: "A List of UIDs obtained from negotiable quote types")
 }
 
-type NegotiableQuoteOperationSuccess {
+union DeleteNegotiableQuoteError = NegotiableQuoteInvalidStateError | NoSuchEntityUidError | InternalError
+
+type DeleteNegotiableQuoteOperationFailure {
     quote_uid: ID!
+    errors: [DeleteNegotiableQuoteError!]!
 }
 
-interface NegotiableQuoteOperationErrorInterface {
-    quote_uid: ID!
-    message: String
-}
-
-type NegotiableQuoteInvalidStateError implements NegotiableQuoteOperationErrorInterface {
-}
-
-type NegotiableQuoteNotFoundError implements NegotiableQuoteOperationErrorInterface {
-}
-
-type NegotiableQuoteUndefinedError implements NegotiableQuoteOperationErrorInterface {
-}
-
-union NegotiableQuoteOperationResult = NegotiableQuoteOperationSuccess | NegotiableQuoteOperationErrorInterface
+union DeleteNegotiableQuoteOperationResult = NegotiableQuoteUidOperationSuccess | DeleteNegotiableQuoteOperationFailure
 
 type DeleteNegotiableQuotesOutput {
-    result_status: MutationResultStatus!
-    operation_results: [NegotiableQuoteOperationResult!]!
+    result_status: BatchMutationStatus!
+    operation_results: [DeleteNegotiableQuoteOperationResult!]!
     # Implementation Note: We don't make the deleted quotes accessible
     # because "deleted" means they're hidden from the storefront UI. They will
     # still be visible (and labeled as deleted) for a Seller in the admin
@@ -195,9 +194,18 @@ input CloseNegotiableQuotesInput {
     quote_uids: [ID!]! @doc(description: "A List of IDs from negotiable quote objects")
 }
 
+union CloseNegotiableQuoteError = NegotiableQuoteInvalidStateError | NoSuchEntityUidError | InternalError
+
+type CloseNegotiableQuoteOperationFailure {
+    quote_uid: ID!
+    errors: [CloseNegotiableQuoteError!]!
+}
+
+union CloseNegotiableQuoteOperationResult = NegotiableQuoteUidOperationSuccess | CloseNegotiableQuoteOperationFailure
+
 type CloseNegotiableQuotesOutput {
-    result_status: MutationResultStatus!
-    operation_results: [NegotiableQuoteOperationResult!]!
+    result_status: BatchMutationStatus!
+    operation_results: [CloseNegotiableQuoteOperationResult!]!
     closed_quotes: [NegotiableQuote!] @doc(description: "An array containing the negotiable quotes that were just closed") @deprecated(reason: "Replaced with operation_results")
     #optionally display all negotiable quotes
     negotiable_quotes(
@@ -450,26 +458,7 @@ type RequestNegotiableQuoteOutput {
     quote: NegotiableQuote
 }
 
-type NegotiableQuoteItemOperationSuccess {
-    quote_item_uid: ID!
-}
-
-interface NegotiableQuoteItemOperationErrorInterface {
-    quote_item_uid: ID!
-    message: String
-}
-
-type NegotiableQuoteItemNotFoundError implements NegotiableQuoteItemOperationErrorInterface {
-}
-
-type NegotiableQuoteItemUndefinedError implements NegotiableQuoteItemOperationErrorInterface {
-}
-
-union NegotiableQuoteItemOperationResult = NegotiableQuoteItemOperationSuccess | NegotiableQuoteItemOperationErrorInterface
-
 type RemoveNegotiableQuoteItemsOutput {
-    result_status: MutationResultStatus!
-    operation_results: [NegotiableQuoteItemOperationResult!]!
     quote: NegotiableQuote
 }
 
--- a/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
+++ b/design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
@@ -178,7 +178,7 @@ type NegotiableQuoteUndefinedError implements NegotiableQuoteOperationErrorInter
 union NegotiableQuoteOperationResult = NegotiableQuoteOperationSuccess | NegotiableQuoteOperationErrorInterface
 
 type DeleteNegotiableQuotesOutput {
-    result_status: TriStateMutationResultStatus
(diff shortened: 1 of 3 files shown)


design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -174,12 +196,9 @@ input CloseNegotiableQuotesInput {
...
 type CloseNegotiableQuotesOutput {
-    closed_quotes(
-        filter: NegotiableQuoteFilterInput,
-        pageSize: Int = 20,
-        currentPage: Int = 1
-        sort: NegotiableQuoteSortInput
-    ): NegotiableQuotesOutput @doc(description: "Quotes that were just closed")
+    result_status: TriStateMutationResultStatus
+    operation_results: [NegotiableQuoteOperationResult!]!
+    closed_quotes: [NegotiableQuote!] @doc(description: "An array containing the negotiable quotes that were just closed") @deprecated(reason: "Replaced with operation_results")
```
pdohogne-magento, line 201: `closed_quotes: [NegotiableQuote!]` is the actual schema in source (which didn't get updated here), and it should now be deprecated in favor of the new `operation_results` field.

pdohogne-magento:
Here's the article where this strategy for handling error results came from explaining why it's a helpful approach: https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -174,12 +196,9 @@ input CloseNegotiableQuotesInput {
 }
 
 type CloseNegotiableQuotesOutput {
-    closed_quotes(
-        filter: NegotiableQuoteFilterInput,
-        pageSize: Int = 20,
-        currentPage: Int = 1
-        sort: NegotiableQuoteSortInput
-    ): NegotiableQuotesOutput @doc(description: "Quotes that were just closed")
+    result_status: TriStateMutationResultStatus!
```
keharper, line 199: Is TriState futureproof? Could an UNKNOWN or other status be added down the road? Maybe rename to something like NegotiableQuoteMutationResultStatus?

amolina-magento: I second this.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -157,7 +157,29 @@ input DeleteNegotiableQuotesInput {
...
+}
+
+type NegotiableQuoteUndefinedError implements NegotiableQuoteOperationErrorInterface {
+}
+
+union NegotiableQuoteOperationResult = NegotiableQuoteOperationSuccess | NegotiableQuoteOperationErrorInterface
+
 type DeleteNegotiableQuotesOutput {
+    result_status: TriStateMutationResultStatus!
+    operation_results: [NegotiableQuoteOperationResult!]!
```
amolina-magento, line 182: Just to be clear, if the result_status is a SUCCESS, then we don't really need to concern ourselves with the individual operation_results. Is that right?

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
pdohogne-magento: The idea was to have a shared enum that could be reused elsewhere outside this module for results that could have partial successes, rather than trying to shoehorn everything into a boolean success status.  I could see maybe adding an "unknown" status, but I can't think of any cases where it would be used; if something happens that a result isn't known, isn't that a failure?

This is meant to be used as a top-level flag for whether or not results need to be investigated.  I don't see a case where we'd want more specific details to be included in this enum; anything that needs more precise information should use a module-specific field.  For NegotiableQuote, that means the `operation_results` field.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
pdohogne-magento: Correct

pdohogne-magento: Thoughts on renaming this field to `quote_operation_results`?

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
amolina-magento: I do like the idea of a shared enum that can be used outside of this module. However, I'm still not a fan of `TriState`. Why not just `MutationResultStatus`?

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
amolina-magento: Hmm, I think I prefer leaving it as `operation_results`.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
pdohogne-magento: Yeah I'm not a huge fan of `TriState` either, I was just trying to come up with something that was an accurate descriptor of its intended use case.  I wish we could replace all usages of Boolean status results with a shared enum like this (which would make me more comfortable with the `MutationResultStatus` name), but alas we cannot.  I do think that's at least better than TriState.

amolina-magento: Even if we can't replace all usages, I still think this is an appropriate name for it. Curious what others, namely @cpartica thinks.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
jimbo: Looks like you're proposing a single array containing a mixture of successes and errors. We had discussed having two lists, separated into successes and errors, but one list will work as well. It would be nice if `NegotiableQuoteOperationResult` included a status as well. As proposed, we'd have to use `message` as a heuristic, right?

```jsx
const elements = []

for (const result of data.operation_results) {
  const { quote_uid: id, message } = result
  const hasError = message != null // heuristic

  // ...
}
```

pdohogne-magento: The idea behind the Union type is that you can filter on the type of the result; it will either be an instance of `NegotiableQuoteOperationSuccess` or `NegotiableQuoteOperationFailureInterface`.  See the article I posted previously talking about this strategy: https://github.com/magento/architecture/pull/505#issuecomment-964570720

pdohogne-magento: Yeah we talked about having the separate lists, but after you dropped off we continued and decided the unified list made more sense and should be just as easy to work with.

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
cspruiell: I prefer `MutationResultStatus` also

aniham (commented):
can't approve or request changes because i don't know enough :) - but makes sense and naming/hierarchy looks reasonable to me

keharper (approved):

aniham (commented):
viewed ✅

eug123 (approved):

abrhim (dismissed):

design-documents/graph-ql/directives.graphqls
```diff
@@ -37,3 +37,9 @@ directive @typeResolver(class: String="") on UNION
     | OBJECT
 
 directive @cache(cacheIdentity: String="" cacheable: Boolean=true) on QUERY
+
+enum MutationResultStatus {
+    SUCCESS
+    FAILURE
+    PARTIAL_SUCCESS
```
cpartica, line 44: other words we can use: recoverable failure/error, 
if PARTIAL_SUCCESS then can we have a total success, total failure?
there's a REST & HTTP counterpart for this: 202: Accepted, 207 Multi-Status

there are IEEE papers on this : https://ieeexplore.ieee.org/document/5216347

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
cpartica: ok, it seems that it's the desired approach as verbally discussed to go further, so consider this comment as a sign off

cpartica (approved):

design-documents/graph-ql/directives.graphqls
pdohogne-magento: Multi-Status instead of Partial success?

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -157,7 +157,29 @@ input DeleteNegotiableQuotesInput {
...
+type NegotiableQuoteInvalidStateError implements NegotiableQuoteOperationErrorInterface {
+}
+
+type NegotiableQuoteNotFoundError implements NegotiableQuoteOperationErrorInterface {
+}
+
+type NegotiableQuoteUndefinedError implements NegotiableQuoteOperationErrorInterface {
+}
+
+union NegotiableQuoteOperationResult = NegotiableQuoteOperationSuccess | NegotiableQuoteOperationErrorInterface
```
pdohogne-magento, line 178: This will not work, you cannot have a union type include an interface.  You can still union on all the concrete types and filter on the shared interface in the query though, so the effect is the same.

```suggestion
union NegotiableQuoteOperationResult = NegotiableQuoteOperationSuccess | NegotiableQuoteInvalidStateError | NegotiableQuoteNotFoundError | NegotiableQuoteUndefinedError
```

design-documents/graph-ql/coverage/b2b/negotiableQuotes.graphqls
```diff
@@ -431,7 +450,26 @@ type RequestNegotiableQuoteOutput {
...
+    message: String
+}
+
+type NegotiableQuoteItemNotFoundError implements NegotiableQuoteItemOperationErrorInterface {
+}
+
+type NegotiableQuoteItemUndefinedError implements NegotiableQuoteItemOperationErrorInterface {
+}
+
+union NegotiableQuoteItemOperationResult = NegotiableQuoteItemOperationSuccess | NegotiableQuoteItemOperationErrorInterface
```
pdohogne-magento, line 468:
```suggestion
union NegotiableQuoteItemOperationResult = NegotiableQuoteItemOperationSuccess | NegotiableQuoteItemNotFoundError | NegotiableQuoteItemUndefinedError
```

aniham (commented):
viewed

eug123 (approved):

cspruiell (approved):

abrhim (commented):
makes relative sense to me. marking this as viewed.

keharper:
Re-approved

PR #510, opened 2022-07-25
Updated purchase order graphql schema

sivaschenko:
Updated purchase order graphql schema 

- Split into separate purchase-order and purchase-order-rule schemas per module
 - Added created_at and updated_at fileds to PurchaseOrder
 - Added addPurchaseOrderItemsToCart mutation:q

### DIFF
--- a/design-documents/graph-ql/coverage/b2b/purchase-order-rule.graphqls
+++ b/design-documents/graph-ql/coverage/b2b/purchase-order-rule.graphqls
@@ -7,15 +7,21 @@ type Mutation {
 type CreatePurchaseOrderApprovalRuleOutput {
     approval_rule: PurchaseOrderApprovalRule @doc(description: "Created purchase order approval rule")
     approval_rules(currentPage: Int = 1, pageSize: Int = 20): PurchaseOrderApprovalRules @doc(description: "A list of purchase order approval rules visible to the customer")
+    status: Boolean! @doc(description: "Indicates whether the operation was successful.")
+    errors: [String] @doc(description: "An array of error messages encountered while performing the operation.")
 }
 
 type UpdatePurchaseOrderApprovalRuleOutput {
     approval_rule: PurchaseOrderApprovalRule @doc(description: "Updated purchase order approval rule")
     approval_rules(currentPage: Int = 1, pageSize: Int = 20): PurchaseOrderApprovalRules @doc(description: "A list of purchase order approval rules visible to the customer")
+    status: Boolean! @doc(description: "Indicates whether the operation was successful.")
+    errors: [String] @doc(description: "An array of error messages encountered while performing the operation.")
 }
 
 type DeletePurchaseOrderApprovalRuleOutput {
     approval_rules(currentPage: Int = 1, pageSize: Int = 20): PurchaseOrderApprovalRules @doc(description: "A list of purchase order approval rules visible to the customer")
+    status: Boolean! @doc(description: "Indicates whether the operation was successful.")
+    errors: [String] @doc(description: "An array of error messages encountered while performing the operation.")
 }
 
 input CreatePurchaseOrderApprovalRuleInput {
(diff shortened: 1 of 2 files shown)


cpartica (changes requested):
- we need to extract approval_flow from Purchase order object and add it to `purchase-order-rule.graphqls` (line 115)
- do we need a `StoreConfig` Setting as well? How does the PWA know how to render "Place purchase Order" button in cart? I even wonder if we should keep PO enabled in the customer - because even if it's customer specific - PWA will still read settings from StoreConfig - and the value in the Storeconfig can be customer specific - but we don't have to give PWA a custom place to look for settings - it just adds additional complexity to the frontend logic
- when placing the purchase order, we might need a purchase order id or a full object as a result of the placeOrder mutation or a mutation similar to `placeNegotiableQuoteOrder` that has a  `PlaceNegotiableQuoteOrderOutput` - note that `Order` ids are different than `purchaseOrder` Ids
- we also need to think about available payment methods for purchase orders - I don't think we can grab those (both online and offline) - note that we do have to know which ones are online or offline - because there's a message to be displayed for online ones in the cart that payment will be collected later when PO is approved
- can we show/hide the tabs on the approval page using the proposed schema? how?
- we should find a way for PurchaseOrder to be concrete impl. of an OrderInterface just like Negotiable Quote is a concrete impl. of a Quote (a cart object) - that way by placing an order - a client can request concrete info of a Purchase Order by knowing that the result is a purchase order with results from the server and just using inline fragments -

sivaschenko:
Thanks for review @cpartica !

Updates and comments:

- approval_flow was extracted from Purchase order object and added to purchase-order-rule.graphqls
- Added purchase_orders_enabled field is present as part of the Customer type

Questions:

- Do we need to add an is_purchase_order field to the quote to identify which title should be displayed?
- Do we need to add is_online field to the quote?
- Do we need to add placePurchaseOrder mutation (and corresponding setPurchaseOrderPaymentMethod, setPurchaseOrderShippingMethod, etc)?
- Are payment method for purchase orders and regular orders different?
- I would consider adding my_purchase_orders, company_purchase_orders and to_approve_purchase_order filed to the Customer type (corresponding tabs should not be displayed in case of empty return for this fields)

sivaschenko:
- Sort Order field determine the position of this item in the list of payment methods that is displayed during checkout.
This is relative to the other payment methods. (0 = first, 1 = second, 2 = third, and so on.) Is the current schema able to afford for this?
 
- If an item in a purchase order is out of stock, or has insufficient quantity available, when the purchase order is converted into an actual order, an error occurs. If backorders are enabled, the order is processed normally.
 
- If they have items in their shopping cart when they click the Add Items to Shopping Cart link, the system displays a dialog. This dialog allows them to choose between merging the cart items with the new items or replacing the items in the shopping cart with the items in the PO.
 

- Tabs 
Company Purchase Orders: POs made by subordinate users within the company (depends on company structure and roles).
Requires My Approval tab: (Visible for designated approvers BASED on company structure (is manager of PO creator) or role (isBuyerAdmin)) POs that are waiting for the customer’s approval. The counter shows how many orders are awaiting approval.

- Bulk actions on POs: Approve Selected POs, Approve Rejected POs (no bulk actions for PO Approval Rules)

sivaschenko:
- The sequence of payment methods should be handled by Quote module: sorted array of available_payment_methods field of Cart type.
- PlaceOrder mutation defined in the Quote module does not have the errors field. However, changing the behaviour of placeOrder mutation is a risky backward incompatible change. I would avoid modifying it. Backorders will be handled by implementation.
- The dialog should be handled by retrieving shopping cart items and items from the purchase order. Depending on user selection the required items will be added to the shopping card using the QuoteGraphQL API.
- I would consider adding my_purchase_orders, company_purchase_orders and to_approve_purchase_order filed to the Customer type @cpartica what do you think about it?
- Bulk actions for approve and reject operation have been added to the schema

sivaschenko:
- type PurchaseOrder implements OrderInterface
- type PlaceOrderOutput {orderV2: OrderInteface!     errors: [PlaceOrderOutputError!]!}
- Quote -> Purchase Order - existing placeOrder mutation from Quote schema, Purchase Order -> Order - add new placePurchaseOrder mutation with a new input, reuse the output
- input AddPurchaseOrderItemsToCartInput {
    purchase_order_uid: ID!
    cart_id: String! @doc(description: "The ID to assign to the cart.")
    replace_existing_cart_items: Boolean!
}
- Add company_purchase_orders, require_my_approval fields to PurchaseOrdersFilterInput
- type AvailablePaymentMethod {is_online }
- type StoreConfig { purchaseorder_enabled: Boolean! } - possibly add to Comany type as well - check how it is used on frontend

